PR #31, opened 2022-05-19
[task] update docker-compose setup,

ZenoLC:
refs #78015

### DIFF
--- a/docker/sample-damap-realm-export.json
+++ b/docker/sample-damap-realm-export.json
@@ -534,7 +534,7 @@
         "roles",
         "profile",
         "email",
-        "tissID"
+        "personID"
       ],
       "optionalClientScopes": [
         "address",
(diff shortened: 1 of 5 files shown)


docker/sample-damap-realm-export.json
```diff
@@ -701,8 +701,12 @@
       "enabled": true,
       "alwaysDisplayInConsole": false,
       "clientAuthenticatorType": "client-secret",
-      "redirectUris": [],
-      "webOrigins": [],
+      "redirectUris": [
+        "http://localhost:8085/*"
+      ],
+      "webOrigins": [
+        "*"
```
flozzone, line 708: I think webOrigins could be "+" here, but since its a example it doesn't count that much, but others could take this example realm to bootstrap other environments.

flozzone (commented):
Looks functional complete now. We could comment the service parts in the compose file a bit, especially why the proxy container is needed.

PR #43, opened 2022-09-16
add project lead on project change

### LINKED ISSUE #41: The Role "Project Leader" should be assigned automatically to the default contact.
When a new project is selected, then the project leader of said project is gathered from the project database and automatically assigned the role/flag of contact. Additionally they should be assigned the "Project Leader" contributor-role.
This helps to differentiate between the two main roles/responsibilities tied to a DMP: that of contact person for the DMP, and that of the Princiupal Investigator of the project. These might be vested onto two different people or be held by the same person. We assume by default the same person, and the user will manually change this if incorrect.

rekt-hard:
closes #41 

On project change, the project leader will be:
- fetched from the connected systems
- added as contributor (if not already added)
- assigned the role of project leader (if no other role has been defined before)
- set as contact (if no other contributor has been set as contact)


rekt-hard:
Sonar failure will probably be fixed by merging it with the repo

ZenoLC (changes requested):
very nicely coded! 
the only issue encountered, is when changing project, the new method does not account for the fact that only one contact should be allowed, and adds new ones.
simple solution: the new project leader should not replace the previous one as contact, nor be assigned the project leader role if someone already has it.
we will check this solution with RDM in case the requirements should differ.

ZenoLC (approved):
issue was with our customised backend, pre-selecting contact persons. this PR is without fault

Issue #63, opened 2022-11-30
Provide Base Services and Resources


rekt-hard:
For project and person adaption possibilities, the provided resources and services should be streamlined. Generic interfaces should be provided, such that institutions have as much freedom as possible when adapting DAMAP.

Additionally, a pagination mechanism or definition should be implemented

At least `read` and `search` interfaces should be provided for resources and services.

PR #64, opened 2022-11-30
Reusable base service

rekt-hard:
closes GH-63 

provide basic interfaces for resources and services.


rekt-hard:
Currently, only person specific adaptions have taken place. If the changes are accepted, project specific adaptions will be performed.

rekt-hard:
@ZenoLC : Should the search method on the service level only have one parameter (query parameters) and then let the implementation decide what to do? Creating a `Search` object seems like putting logic in the resource layer, when it should really be in the service layer (i.e. defining endpoints and parameters vs implementing logic).

ZenoLC:
Yes, you are right, since we are not sending the additional information to the frontend we can just defer to the implementations discretion.

rekt-hard:
Rebased the branch on `next`

src/main/java/at/ac/tuwien/damap/rest/persons/orcid/ContributorORCIDExpandedSearchMapper.java
```diff
@@ -11,6 +11,7 @@ public class ContributorORCIDExpandedSearchMapper {
     public ContributorDO mapEntityToDO(ContributorORCIDExpandedSearch contributor, ContributorDO contributorDO) {
 
         contributorDO.setId(null);
+        contributorDO.setUniversityId(contributor.orcidId);
```
ZenoLC, line 14: please remove this before merge, as discussed previously

ZenoLC (changes requested):
Looks good and works fine. 
Only thing missing is to remove the assigning of ORCID-ID to uni-ID
And wait for frontend changes approval.

This changes the current behavior, in which a default list of projects is loaded initially. I personally would bring that back, as an additional check, but it is not within the scope of this PR, and will be made its own task.

ZenoLC (approved):
nice

PR #79, opened 2023-02-07
refactor ORCID related files

rekt-hard:
refactor/rename ORCID related files

## Description

Bugfix, Refactoring

#### What does this PR do?

Set university ID of ORCID person, as this is also used for front-end logic (e.g. selecting contact person).
Refactor/Rename ORCID related files to match ORCID API. Also eases reading and understanding the code.
Implement fetching a single person from ORCID by its orcid.

#### Breaking changes

none

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-

### DIFF
--- a/src/main/java/at/ac/tuwien/damap/rest/persons/orcid/ORCIDExpandedSearchResult.java
+++ b/src/main/java/at/ac/tuwien/damap/rest/persons/orcid/ORCIDExpandedSearchResult.java
@@ -4,6 +4,7 @@ import java.util.List;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+
 import lombok.Data;
 
 @Data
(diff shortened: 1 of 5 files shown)


ZenoLC:
The universityID is often used in back- and frontend to limit the use of certain contributor objects to functions accessible only by members of the institution running an implementation of damap. Filling it with a non-functional ID, that does not correspond with an institutional ID will cause several issues when trying to gather information from other services with it. E.g. the EDITOR assigning feature will display ORCID users, even though they can not log into the application.

If there are issues that rise from not assigning this ID, then they should be solved differently, as the overhead from having to manage different sets of IDs in the same ID-variable will prove more effort.

rekt-hard:
Thank you for the clarification :)
A possible solution would then be to change the logic in the front-end for selecting the contact person. What would you propose?

ZenoLC:
changing frontend logic sounds like the way to go. i do not have an issue assigning an ORCID contributor as contact person. what is the issue you are experiencing on your end?

rekt-hard:
Discussed via meeting.
Outcome: do not set UniversityID for ORCID contributor, because of reasons mentioned above. Instead, change front-end logic when selecting a contact person (and other places if needed)

ZenoLC (approved):
good job

Issue #77, opened 2023-01-26
GDPR data


clarascu:
**Description**
Users must be able to see what kind of personal data Damap has stored about them.

**Proposed solution**
Add an API, that allows these data to be queried by users (via the frontend), and possibly by other (connected) tools.
Use annotations to mark GDPR related fields, so that it can be easily adapted in the future.

- [ ] Include audited data

PR #80, opened 2023-02-14
Add GDPR utility

clarascu:
## Description

Feature

#### What does this PR do?
Adds annotations that mark GDPR related fields, and endpoint to retrieve data.

#### Breaking changes
None

#### Code review focus
If there's an edge case, that is not covered properly.

### Checks

- [x] Tested with Oracle/PostgreSQL
- [x] Documentation added
- [x] Tests added
- [x] E2e tests created

closes GH-77

### DIFF
--- a/src/main/java/at/ac/tuwien/damap/repo/GdprRepo.java
+++ b/src/main/java/at/ac/tuwien/damap/repo/GdprRepo.java
@@ -1,0 +1,40 @@
+package at.ac.tuwien.damap.repo;
+
+import at.ac.tuwien.damap.rest.gdpr.domain.GdprResult;
+import at.ac.tuwien.damap.rest.gdpr.domain.HqlQuery;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import javax.persistence.EntityManager;
+import java.util.*;
+
+@ApplicationScoped
+public class GdprRepo {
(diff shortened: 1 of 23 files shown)


src/main/java/at/ac/tuwien/damap/repo/GdprRepo.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private GdprResult getGdprEntityDataByUniversityId(HqlQuery query, String universityId) {
+
+        GdprResult gdprResult = new GdprResult();
+        gdprResult.setEntity(query.getEntityName());
+
+        @SuppressWarnings("unchecked")
+        List<Map<String, Object>> result = entityManager.createQuery(query.getHql(), (Class<Map<String, Object>>) (Class<?>) Map.class)
+                .setParameter("id", universityId).getResultList();
+        result.forEach(map -> map.values().removeIf(Objects::isNull));
```
ZenoLC, line 34: please add a comment detailing the behavior of this removal, as on first glance i was worried we might be removing key information here (like personID, which is relevant result itself)
```suggestion
        //check if any of the principal output/result entities (e.g. Access, Consent, Contributor) are empty, and remove them if so.
```

ZenoLC (approved):
looks nicely generalized

PR #82, opened 2023-02-23
Fix HibernateProxy error

clarascu:
Apparently reflection retrieves the HibernateProxies for annotated classes as well, which is not what we want as we only need (and can work with) the actual classes.

## Description

Bugfix

#### What does this PR do?
Fixes a bug introduced in ec5ccaade4acacaacaea926acca3e221e489299d.
HibernateProxy classes are now ignored, when creating the GDPR queries.

### Checks

refs GH-77


ZenoLC (approved):
good job

ZenoLC (approved):
good job

PR #83, opened 2023-02-27
Invenio-damap: Support connection from InvenioRDM

rekt-hard:
## Description

Feature

#### What does this PR do?

- Supports incoming connections from InvenioRDM-based repositories.
- Adds 2 new endpoints for:
  -  fetching DMPs and,
  -  setting DMP information based on the InvenioRDM record which is being connected.
- Creates a new version after adding a dataset remotely.

#### Code review focus

Mapping common maDMP standard dataset to DmpDO.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [x] Tests added
- [ ] E2e tests created

closes https://github.com/fair-data-austria/invenio-damap/issues/13, https://github.com/fair-data-austria/invenio-damap/issues/7

### DIFF
--- a/src/main/java/org/damap/base/enums/EDataAccessType.java
+++ b/src/main/java/org/damap/base/enums/EDataAccessType.java
@@ -46,4 +46,36 @@ public enum EDataAccessType {
       MAP.put(type.getValue(), type);
     }
   }
+
+  /**
+   * Compares this {@code EDataAccessType} instance with another {@code EDataAccessType} to
+   * determine which is more restrictive. The order of restriction is defined as:
+   *
+   * <ul>
+   *   <li>{@code CLOSED} is the most restrictive.
+   *   <li>{@code RESTRICTED} is more restrictive than {@code OPEN} but less restrictive than {@code
+   *       CLOSED}.
+   *   <li>{@code OPEN} is the least restrictive.
+   * </ul>
+   *
+   * @param other the other {@code EDataAccessType} to compare to; can be {@code null}. If {@code
+   *     null}, this instance is considered more restrictive.
+   * @return {@code 1} if this instance is more restrictive than {@code other}, {@code -1} if this
+   *     instance is less restrictive than {@code other}, and {@code 0} if both instances are equal.
+   */
+  public int compare(EDataAccessType other) {
+    if (other == null) {
+      return 1;
+    }
+
+    if (this == other) {
+      return 0;
+    }
+
+    return switch (this) {
+      case CLOSED -> 1;
+      case RESTRICTED -> (other == CLOSED) ? -1 : 1;
+      case OPEN -> -1;
+    };
+  }
 }
--- a/src/main/java/org/damap/base/rest/invenio_damap/DMPPayload.java
+++ b/src/main/java/org/damap/base/rest/invenio_damap/DMPPayload.java
@@ -1,0 +1,13 @@
+package org.damap.base.rest.invenio_damap;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Data;
+import org.damap.base.rest.madmp.dto.Dataset;
+
+@Data
+public class DMPPayload {
+  @JsonProperty("dmp_id")
+  private long dmpId;
+
+  private Dataset dataset;
+}
--- a/src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
+++ b/src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
@@ -1,0 +1,73 @@
+package org.damap.base.rest.invenio_damap;
+
+import io.quarkus.resteasy.reactive.server.EndpointDisabled;
+import io.quarkus.security.ForbiddenException;
+import jakarta.enterprise.context.RequestScoped;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.NotFoundException;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.core.HttpHeaders;
+import jakarta.ws.rs.core.MediaType;
+import java.util.AbstractMap.SimpleEntry;
+import java.util.List;
+import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.rest.dmp.domain.DmpDO;
+import org.damap.base.rest.madmp.dto.Dataset;
+import org.damap.base.security.SecurityService;
+import org.damap.base.validation.AccessValidator;
+
+@Path("/api/madmps")
+@RequestScoped
+@Produces(MediaType.APPLICATION_JSON)
+@JBossLog
+@EndpointDisabled(name = "invenio.disabled", stringValue = "true")
+public class InvenioDAMAPResource {
+
+  @Inject
+  public InvenioDAMAPResource(
+      AccessValidator accessValidator,
+      InvenioDAMAPService invenioDAMAPService,
+      SecurityService securityService) {
+    this.accessValidator = accessValidator;
+    this.invenioDAMAPService = invenioDAMAPService;
+    this.securityService = securityService;
+  }
+
+  AccessValidator accessValidator;
+  InvenioDAMAPService invenioDAMAPService;
+  SecurityService securityService;
+
+  @GET
+  public List<DmpDO> getDmpListByPerson(@Context HttpHeaders headers) {
+    SimpleEntry<List<DmpDO>, String> result =
+        invenioDAMAPService.resolveDmpsAndIds(securityService.checkIfUserIsAuthorized(headers));
+    return result.getKey();
+  }
+
+  @POST
+  @Consumes(MediaType.APPLICATION_JSON)
+  public DmpDO addDataSetToDmp(@Context HttpHeaders headers, DMPPayload payload) {
+    SimpleEntry<List<DmpDO>, String> result =
+        invenioDAMAPService.resolveDmpsAndIds(securityService.checkIfUserIsAuthorized(headers));
+
+    long dmpId = payload.getDmpId();
+    log.info("Add dataset to dmp with id: " + dmpId);
+    Dataset dataset = payload.getDataset();
+
+    if (result.getKey().stream().noneMatch(dmpDO -> dmpDO.getId().equals(dmpId))) {
+      throw new NotFoundException("DMP with id " + dmpId + " could not be found.");
+    }
+
+    String personId = result.getValue();
+    // Assuming all IDs provided fetch the same DMPs, check permissions.
+    if (!accessValidator.canEditDmp(dmpId, personId))
+      throw new ForbiddenException(
+          "Person " + personId + "Not authorized to access dmp with id " + dmpId);
+    return invenioDAMAPService.addDataSetToDMP(dmpId, dataset);
+  }
+}
--- a/src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPService.java
+++ b/src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPService.java
@@ -1,0 +1,116 @@
+package org.damap.base.rest.invenio_damap;
+
+import io.quarkus.security.UnauthorizedException;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import jakarta.json.JsonObject;
+import jakarta.transaction.Transactional;
+import java.text.MessageFormat;
+import java.util.AbstractMap.SimpleEntry;
+import java.util.Date;
+import java.util.List;
+import org.damap.base.rest.dmp.domain.DatasetDO;
+import org.damap.base.rest.dmp.domain.DmpDO;
+import org.damap.base.rest.dmp.service.DmpService;
+import org.damap.base.rest.madmp.dto.Dataset;
+import org.damap.base.rest.version.VersionDO;
+import org.damap.base.rest.version.VersionService;
+import org.eclipse.microprofile.jwt.JsonWebToken;
+
+@ApplicationScoped
+public class InvenioDAMAPService {
+
+  protected DmpService dmpService;
+  protected VersionService versionService;
+
+  @Inject
+  InvenioDAMAPService(DmpService dmpService, VersionService versionService) {
+    this.dmpService = dmpService;
+    this.versionService = versionService;
+  }
+
+  // Resolves user identity based on the DMPs and returns dmp list and list of
+  // identifiers if successful.
+  public SimpleEntry<List<DmpDO>, String> resolveDmpsAndIds(JsonWebToken jwt) {
+    JsonObject invenioDamapClaim = jwt.getClaim("invenio-damap");
+    if (invenioDamapClaim == null) {
+      throw new UnauthorizedException("Missing invenio-damap claim in jwt.");
+    }
+
+    JsonObject identifiers = invenioDamapClaim.getJsonObject("identifiers");
+    if (identifiers == null || identifiers.isEmpty()) {
+      throw new UnauthorizedException("No valid authentication schema was provided.");
+    }
+
+    List<DmpDO> personDmpList = null;
+    String matchingIdentifier = null;
+
+    for (String key : identifiers.keySet()) {
+      String identifier = identifiers.getString(key);
+      List<DmpDO> dmps = dmpService.getDmpDOListByPersonId(identifier);
+
+      // Check if returned DMPs list is empty.
+      // If yes, no resolving can take place, move to the next identifier.
+      if (dmps.isEmpty()) continue;
+
+      // If a second identifier matches, something went wrong
+      if (matchingIdentifier == null) {
+        matchingIdentifier = identifier;
+        personDmpList = dmps;
+      } else {
+        throw new UnauthorizedException("Mismatch in resolved identities.");
+      }
+    }
+
+    // Check if identifier is null. This means that no resolving was possible.
+    if (matchingIdentifier == null)
+      throw new UnauthorizedException("Identities could not be resolved.");
+
+    return new SimpleEntry<>(personDmpList, matchingIdentifier);
+  }
+
+  @Transactional
+  public DmpDO addDataSetToDMP(long dmpId, Dataset dataset) {
+
+    DmpDO dmpDO = dmpService.getDmpById(dmpId);
+    var datasetDO =
+        dmpDO.getDatasets().stream()
+            .filter(
+                ds -> {
+                  var localIdentifier = ds.getDatasetId();
+                  var externalIdentifier = dataset.getDatasetId();
+
+                  if (localIdentifier == null || externalIdentifier == null) {
+                    return false;
+                  }
+
+                  return localIdentifier.getIdentifier() != null
+                      && externalIdentifier.getIdentifier() != null
+                      && localIdentifier.getType() != null
+                      && externalIdentifier.getType() != null
+                      && localIdentifier.getIdentifier().equals(externalIdentifier.getIdentifier())
+                      && localIdentifier
+                          .getType()
+                          .toString()
+                          .equalsIgnoreCase(externalIdentifier.getType().name());
+                })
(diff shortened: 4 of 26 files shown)


ZenoLC (approved):

rekt-hard:
Please make sure to squash the commits :)

SotosTsepe (approved):

src/main/java/org/damap/base/rest/invenio_damap/DMPPayload.java
```diff
@@ -0,0 +1,27 @@
+package org.damap.base.rest.invenio_damap;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.damap.base.rest.madmp.dto.Dataset;
+
+public class DMPPayload {
```
rekt-hard, line 8: Add the `@Data` annotation from lombok for easy (de)serialization

src/main/java/org/damap/base/rest/invenio_damap/DMPPayload.java
```diff
@@ -0,0 +1,27 @@
...
+  @JsonProperty("dmp_id")
+  private long dmp_id;
```
rekt-hard:
```suggestion
  private long dmpId;
```
The json property annotation takes care of (de)serialization to/from `dmp_id`

src/main/java/org/damap/base/rest/invenio_damap/DMPPayload.java
```diff
@@ -0,0 +1,27 @@
...
+    this.dmp_id = dmp_id;
+  }
+
+  public Dataset getDataset() {
+    return dataset;
+  }
+
+  public void setDataset(Dataset dataset) {
+    this.dataset = dataset;
+  }
```
rekt-hard: The `@Data` annotation takes care of generating casual getters and setters. These can be removed.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
```diff
@@ -0,0 +1,171 @@
...
+
+  /*
+   * Maybe make it configurable?
+   * Could be more elaborate but should suffice for PoC
+   * Could also go into the {@link org.damap.base.security.SecurityService},
+   * but keeping everything together for PoC
+   *
+   */
+  private JsonWebToken validateAuthHeader(HttpHeaders headers) {
+    String jwt_token = headers.getHeaderString("X-Auth");
```
rekt-hard: https://jwt.io/introduction/ states that a jwt is usually sent as `Authorization: Bearer <token>`. Why are we not using this here?

rekt-hard:
```suggestion
    String jwtToken = headers.getHeaderString("X-Auth");
```

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
```diff
@@ -0,0 +1,171 @@
...
+
+    if (jwt_token != null && !jwt_token.isEmpty()) {
+      try {
+        JsonWebToken jwt = parser.verify(jwt_token, sharedSecret);
+
+        long exp = jwt.getExpirationTime();
+        long currentTime = System.currentTimeMillis() / 1000;
+
+        if (currentTime >= exp) throw new UnauthorizedException("Token expired.");
```
rekt-hard: Do we need to do this manually or will this be taken care of by verify?

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    // Check if DMP list is null. This means that no resolving was possible.
+    if (personDmpList == null) throw new NotFoundException("No valid identities were found.");
+
+    return new SimpleEntry<>(personDmpList, matchingIdentifiers);
+  }
+
+  @GET
+  public List<DmpListItemDO> getDmpListByPerson(@Context HttpHeaders headers) {
```
rekt-hard: We should return a list of DMPs that conforms to the maDMP standard of the RDA. So we have to map the results from here to maDMP (there should be a mapper available).

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
```diff
@@ -0,0 +1,171 @@
...
+    JsonWebToken jwt = checkIfUserIsAuthorized(headers);
+    JsonObject invenioDamap = jwt.getClaim("invenio-damap");
+    JsonObject identifiers = invenioDamap.getJsonObject("identifiers");
```
rekt-hard:
```suggestion
    JsonObject identifiers = jwt.getClaim("invenio-damap").getJsonObject("identifiers");
```

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
```diff
@@ -0,0 +1,171 @@
...
+    return dmps;
+  }
+
+  @POST
+  @Consumes(MediaType.APPLICATION_JSON)
+  public DmpDO addDataSetToDmp(@Context HttpHeaders headers, DMPPayload payload) {
+
+    JsonWebToken jwt = checkIfUserIsAuthorized(headers);
+    JsonObject invenioDamap = jwt.getClaim("invenio-damap");
+    JsonObject identifiers = invenioDamap.getJsonObject("identifiers");
```
rekt-hard:
```suggestion
    JsonObject identifiers = jwt.getClaim("invenio-damap").getJsonObject("identifiers");
```

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
```diff
@@ -0,0 +1,171 @@
...
+
+    SimpleEntry<List<DmpListItemDO>, List<String>> result = resolveDmpsAndIds(identifiers);
+    List<String> personIds = result.getValue();
+
+    long dmp_id = payload.getDmpId();
```
rekt-hard:
```suggestion
    long dmpId = payload.getDmpId();
```

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
```diff
@@ -0,0 +1,171 @@
...
+
+  // Resolves user identity based on the DMPs and returns dmp list and list of
+  // identifiers if successful.
+  private SimpleEntry<List<DmpListItemDO>, List<String>> resolveDmpsAndIds(JsonObject identifiers) {
+    List<DmpListItemDO> personDmpList = null;
+    List<String> matchingIdentifiers = new ArrayList<>();
+
+    for (String key : identifiers.keySet()) {
+      String identifier = identifiers.getString(key);
+      List<DmpListItemDO> dmps = dmpService.getDmpListByPersonId(identifier);
```
rekt-hard: general note: We should __really__ try to get some sort of user management into DAMAP. Otherwise, we always have to fetch DMPs :)

rekt-hard (commented):
nice work implementing the jwt authorization :rocket: 

What might be missing is refactoring/moving some of the logic from the resource level to the service level. Also making sure that the service can be overwritten aka an own service can be provided.
Might be appropriate to move authorization logic to the security service (also ensuring an own service can be provided).

I still have to test this locally so some other issues might come up.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
SotosTsepe: Here I didn't keep you in the loop with some limitations I faced. If we pass the exact `Authorization Bearer <token>` header, then some functionality from the quarkus-oidc library triggers, which forwards the request to keycloak. This code is not executed at all, hence this is the decision I took.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
SotosTsepe: This was discovered while running the automated tests, it is not taken care of automatically.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
SotosTsepe: I need a little more context here, a list of DMPDOs or the actual maDMPs? In any case, something is indeed missing. Having finalized the InvenioRDM frontend changes, the "already linked" label is not displayed and the cause seems to be the absence of the dataset in the response.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
SotosTsepe: Agree with that, this was rather painful :smiley: This should come as a future work though :slightly_smiling_face:

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
rekt-hard: A list of maDMPs. Some sort of pagination would also be great but we did not settle on this.
I think we had maDMPs before as well which is what the InvenioRDM implementation is based on

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
```diff
@@ -0,0 +1,178 @@
...
+      // iterated identifier.
+      matchingIdentifiers.add(identifier);
+
+      // Keep track of the 1st list that matches at least one DMP.
+      // Then compare it with other non-null or non-empty lists.
+      if (personDmpList == null) personDmpList = dmps;
+
+      // If a DMP list is different than the previous one, a different identity was
+      // resolved.
+      else if (!personDmpList.equals(dmps))
```
ValentinFutterer: It is (theoretically) possible that two identifiers return the same dmps - but the ordering is not guaranteed by the db. Therefore, this compare should be made with containsAll(), just to be sure.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
rekt-hard: Sorry about the confusion and extra work. maDMPs do not have a local identifier (i.e. database id) which is why we can not simply use a list of maDMPs. The Invenio frontend would not know which id to send in order to add a dataset to a DMP. Using a List of `DmpDO` solves this issue, and also includes the datasets (needed for some UX on Invenio side) as compared to the `DmpListItemDO`.

SotosTsepe: Agreed, thanks! :rocket:

src/main/java/org/damap/base/rest/invenio_damap/InvenioDamapResourceMapper.java
```diff
@@ -0,0 +1,172 @@
...
+    datasetDO.setSize(0L);
+
+    mapDistribution(madmpDataset, datasetDO, dmpDO);
+
+    datasetDO.setOtherProjectMembersAccess(EAccessRight.READ);
+
+    Boolean personalData = true;
+    madmpDataset.setPersonalData(
+        Objects.requireNonNullElse(madmpDataset.getPersonalData(), Dataset.PersonalData.UNKNOWN));
+    switch (madmpDataset.getPersonalData()) {
```
ValentinFutterer: It is not necessary to initialize the personalData variable.
It is also unnecessary to set personalData in madmpDataset - `Objects.requireNonNullElse(madmpDataset.getPersonalData(), Dataset.PersonalData.UNKNOWN))` can be used in the switch statement instead of `madmpDataset.getPersonalData()`.

the same is also true for the sensitiveData part.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPService.java
```diff
@@ -0,0 +1,110 @@
...
+                          .toString()
+                          .equalsIgnoreCase(externalIdentifier.getType().name());
+                })
+            .findFirst()
+            .orElse(new DatasetDO());
+
+    datasetDO = InvenioDamapResourceMapper.mapMaDMPDatasetToDatasetDO(dataset, datasetDO, dmpDO);
+
+    dmpDO.getDatasets().add(datasetDO);
+    dmpDO = dmpService.update(dmpDO);
```
ValentinFutterer: If `.findFirst()` returns a datasetDO, it is readded to the dmpDO datsasets. This leads to a duplicated dataset in the list.

rekt-hard (commented):
DAMAP should return a:
- 401 if the no valid authentication schema is provided, identity could not be resolved or multiple identities resolve to different DMPs.
- 403 if the identity can be resolved, but this identity is not allowed to perform a certain action
- 404 if an endpoint is not found, a resource is not available (i.e. wrong DMP id when adding a dataset), or to mask a 403 if the existence of a resource should not be revealed)

src/main/java/org/damap/base/enums/EDataAccessType.java
```diff
@@ -46,4 +46,20 @@ public static EDataAccessType getByValue(String value) {
       MAP.put(type.getValue(), type);
     }
   }
+
+  public int compare(EDataAccessType other) {
```
rekt-hard, line 66: We can generate the doc string for missing methods via a maven command.
Either way, a description of what this compare exactly does would be great.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
```diff
@@ -0,0 +1,69 @@
...
+      throw new NotFoundException("DMP with id " + dmpId + " could not be found.");
+    }
+
+    // Assuming all IDs provided fetch the same DMPs, check permissions.
+    for (String personId : personIds) {
+      if (!accessValidator.canEditDmp(dmpId, personId))
+        throw new ForbiddenException(
+            "Person " + personId + "Not authorized to access dmp with id " + dmpId);
+    }
+    log.info("Add dataset to dmp with id: " + dmpId);
```
rekt-hard: We usually have these log statements at the beginning of the method

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPService.java
```diff
@@ -0,0 +1,120 @@
...
+    List<DmpDO> personDmpList = null;
+    List<String> matchingIdentifiers = new ArrayList<>();
+
+    for (String key : identifiers.keySet()) {
+      String identifier = identifiers.getString(key);
+      List<DmpDO> dmps = dmpService.getDmpDOListByPersonId(identifier);
+
+      // Check if returned DMPs is null or the list is empty.
+      // If yes, no resolving can take place, move to the next identifier.
+      if (dmps == null || dmps.isEmpty()) continue;
```
rekt-hard: From the service level, we do/should not return null for a search/get method. Should be fine to omit the null check.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPService.java
```diff
@@ -0,0 +1,120 @@
...
+
+      // Here dmps would contain at least one DMP so keep track of the current
+      // iterated identifier.
+      matchingIdentifiers.add(identifier);
```
rekt-hard: From DAMAP side at this level, if we get results for different identifiers, can we assume different identities? Do we even have to continue after getting results for two different identifiers? It is not like we will store all the DMPs once with an email and once with the university ID for the same person. It is more like: the external application does not know which identifier is the correct one (email or university ID), so both are sent and one should be resolvable.

It will not hurt to continue here. We could omit the equality check of the DMP list though.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPService.java
rekt-hard: So we should check the `datasetDO` id before mapping. Only add the dataset, if the id was not set before?

rekt-hard (commented):
Looking very good. Just some small comments.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDamapResourceMapper.java
```diff
@@ -0,0 +1,169 @@
...
+
+      if (d.getByteSize() != null
+          && d.getByteSize() > Objects.requireNonNullElse(datasetDO.getSize(), 0L)) {
+        datasetDO.setSize(d.getByteSize().longValue());
+      }
+
+      if (d.getHost() == null) {
+        // Nothing more to do
+        return;
+      }
```
rekt-hard: Should we just continue the loop here or really return from the method?

src/main/java/org/damap/base/rest/invenio_damap/InvenioDamapResourceMapper.java
```diff
@@ -0,0 +1,169 @@
...
+      if (datasetHashes.contains(datasetDO.getReferenceHash())) {
+        datasetHashes.add(datasetDO.getReferenceHash());
+      }
+    }
+
+    // TODO: Support multiple licenses
+    Arrays.stream(ELicense.values())
+        .filter(eLicense -> licenses.contains(eLicense.getUrl()))
+        .findFirst()
+        .ifPresent(datasetDO::setLicense);
```
rekt-hard: For now, we should set the first license we can resolve from a distribution and ignore further licenses. The maDMP datamodel does not support multiple licenses so we should at least try to set one.

src/main/java/org/damap/base/enums/EDataAccessType.java
ValentinFutterer: Yes true, i forgot, thanks for reminding me.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPResource.java
ValentinFutterer: Easy fix 👍

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPService.java
ValentinFutterer: True that will never happen, an empty list is possible though. So i will just emit the null check.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPService.java
ValentinFutterer: Yes, i was so preoccupied with the mapper i forgot to address my own comment 😅 
Thanks for reminding me.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDAMAPService.java
ValentinFutterer: > From DAMAP side at this level, if we get results for different identifiers, can we assume different identities?

I would say yes, since DAMAP will always use only one type of identifier and invenio will send only different types. So the only way this error could even be triggered, is by some admin making a mistake. So i would propose to throw the unauthorized exception when two identifiers match as a failsafe) and to remove the list equality check, like you suggested.

src/main/java/org/damap/base/rest/invenio_damap/InvenioDamapResourceMapper.java
ValentinFutterer: `.findFirst()` ignores further licenses after the first match - or am i missing something?

src/main/java/org/damap/base/rest/invenio_damap/InvenioDamapResourceMapper.java
ValentinFutterer: Right, this should be a continue.

SotosTsepe:
LGTM :rocket:

src/main/java/org/damap/base/rest/invenio_damap/InvenioDamapResourceMapper.java
rekt-hard: discussed offline. we should iterate the received licenses and match them with the enum license

PR #99, opened 2023-04-20
update customisation instructions

ZenoLC:
## Description

Update customisation instruction to contain the new person service integration.


rekt-hard (approved):
lgtm

PR #102, opened 2023-07-11
[Bugfix] change scope of export classes to not retain previous data.

ZenoLC:
## Description

Currently the export classes are set to ApplicationScope, which retains the information of the previous export if not overwritten during the new export.

#### What does this PR do?

This PR changes the scope to RequestScope, so that information is retained only per request.

closes GH-


rekt-hard (approved):
good catch. LGTM

PR #116, opened 2023-09-27
exports: add qute template for horizon europe

rekt-hard:
First try of resolving https://github.com/tuwien-csd/damap-backend/issues/114

This shows a basic implementation of

- having html template
- render with qute
- convert to to .docx for the export

As can be seen, the code required for a new template would be more concise than the current implementation. It also allows to track the changes made to the template.

The template is not complete. Converting html to .docx comes with some issues:

- Styling: Not everything supported in html will also be supported in .docx.
- Enumerations: Padding and styling is not converted correctly from html. Probably needs additional XML.
- Header/Footer per page: Requires additional XML in the template.
- Table of contents: Easy to link to certain parts/sections but getting the page number could be difficult.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<project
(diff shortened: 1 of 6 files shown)


rekt-hard:
Might have tackled this the wrong way.
Another possibility is to create .xml templates for each section. Then render these section in the export and put them in the document with docx4j. This way, we retain the file changes via git and get all the functionality of the layout possibilities - no need to convert from .html to .docx.

Writing the correct xml is not straight forward. A possible way to achieve this is to create a certain part of the finished document (i.e. section or table) in a document processing tool (i.e. microsoft word, libreoffice writer) and then inspect the corresponding xml output in the .docx archive. The xml will then have to be adapted in the template.

ghost:
Closing as discussed.

Issue #100, opened 2023-07-06
Change liquibase changesets managements to use several files


ZenoLC:
**Changes**
Currently we have a single file managing all liquibase statements (db/changeLog.yaml).
Instead we want to be able to create new files for each (set of) changeset.
So we should adopt a coordinator/manager file, which includes all newly created files.
The coordinator file will then replace the changeLog.yaml in the configurations.

**Additional notes**
This will enable implementing institutions to be more flexible with which changesets they want to include.
They can use the default one, or have their own manager which allows them to skip certain changesets.
Also it provides an example of how implementing institutions can add scripts of their own in between the open source changesets.

**Assumptions**
Ideally a partner will never have to skip any changeset on the official project.
Just idea of keeping the changesets split up into their respective files will be a huge advantage for readability.

As far as adding custom changesets in between:
Liquibase does not care about the sequence of the changesets. It will only check if the hashes of each changeset are correct/unmodified after execution. Therefore addine some in between (e.g. adding data) will not interfere with the project.

**Acceptance Criteria**
Please do thoroughly check the correct functionality, as this will affect productive systems. 
All releases containing changes to how changesets are managed or new changesets should be very explicitly labeled.

PR #121, opened 2023-11-16
#100 - liquibase changesets are noe managed using multiple files (bre…

ValentinFutterer:
…aking changes)

## Description

Refactoring/Documentation

#### What does this PR do?

Liquibase files are now supposed to be split. Added new filestructure and updated config files. Added documentation on how to use the new system.
Edit-----------------
Before version 3, all changeset where recorded in changelog.yaml, this is now kept as a legacy changelog.
changeLog.yaml is being kept outside the damap folder for compatibility with older institutional backends.
New files should now be created for each new changeset. These new files are all included manually.
Using includeAll with this naming scheme was not possible, since Liquibase would execute version 10 before version 2. A custom comparator cannot be used, since liquibase is loaded via the quarkus class loader (Base Runtime ClassLoader),
which cannot load the application root or any hot deployed code.

#### Breaking changes

This is a breaking change - institutional backends need to revise their liquibase file structure and adapt to the new system.

#### Code review focus

Is the new file structure and naming scheme sensible? Is the added documentation enough / is it missing important parts / is it understandable?

### Checks

- [x] Tested with Oracle/PostgreSQL
- [x] Documentation added

closes GH-


ZenoLC (changes requested):
Hi, I have some notes (mostly on the documentation structure):

* the example changelog file is not necessary, i would remove it altogether, as it is well explained in the root file.
* The text is a bit too long, i would limit it to strictly the instructions on how to add new items and naming convention. (rows 2-9) and the immediate instructions 30-32.
* I would also suggest that we use the full version, including the hotfix, to avoid confusion (so "changeLog-3.0.0_1.yaml")
* explanations on the solution decisions belong into the PR description directly for posterity (row 1, and 11-13). In here it is not immediately useful to the future developers.
* I would move the customization instructions (15-28) into the readme file CUSTOMISING.md, where it will be more at home alongside the other customization instructions for institutions implementing their own thing. 
* additionally in the customizing text I would add the information on how to override the root liquibase file.
* and also add the information, that the changeset ID can be anything, and does not need to be sequential (liquibase will not care what the name is, therefore no conflict will arise from having two different parallel ID systems). but reinforce (as you did) that they should avoid using integer sequence to avoid conflicts.

CUSTOMISING.md
```diff
@@ -111,6 +111,26 @@ damap:
       db-kind: postgresql # your database type
 ```
```
ZenoLC, line 113: placement is good after the database, I would put it in its own title/section
```suggestion

### Custom Database Changes

```

CUSTOMISING.md
```diff
@@ -111,6 +111,26 @@ damap:
...
 
+If you want to customize liquibase for your own institutional backend:
+- Create a new folder in db with your institutions name - like e.g. tuwien
+- Create your own root changelog and change the liquibase path in application.yaml to point to your new root changelog
+- Your new root changelog should look something like this:
```
ZenoLC, line 117: i would also show an exmple of what to insert into the application yaml:
```suggestion
Here an example of how to override the root changelog in the application.yaml file of your application:

quarkus:
  liquibase:
    migrate-at-start: true
    change-log: db/tuwien/mainChangeLog.yaml

- Your new root changelog should look something like this:
```
also not forget to make the sample code formatted to yaml

ZenoLC (changes requested):
2 more minor notes, then it is perfect :)

src/main/resources/db/damap/changeLog-root.yaml
```diff
@@ -0,0 +1,21 @@
...
+# minor version, use "changeLog-3.0.0_2.yaml" to distinguish between them.
+# Please stick to these naming guidelines, to avoid conflicts.
+#
+# To include a new changeset, generate a new version folder (if not already there) and then create your changeset - follow
+# the naming guidelines. Then append an include statement below, including your new changeset.
+# Important - keep in mind the id - check which one your predecessor used.
+#
+# The commented out include statement below includes a sample changeset, to show how it should be done. If you are
+# the first one to add a changeset using the new system - uncomment it and adapt it to your needs.
+# Afterward please delete this comment-block.
```
rekt-hard, line 15: nitpick: The procedure is explained in the paragraphs above. I would remove this and the example include.

src/main/resources/application.yaml
```diff
@@ -69,7 +69,7 @@ quarkus:
 
   liquibase:
     migrate-at-start: true
-    change-log: db/changeLog.yaml
+    change-log: db/damap/changeLog-root.yaml
```
rekt-hard, line 72: Shall the namespacing be more explicit to be safe? `at/ac/tuwien/damap` (as is done with java source files)

CUSTOMISING.md
```diff
@@ -111,6 +111,26 @@ damap:
...
+- After that, you can create your own changelogs in your custom institutional folder and include them in your root
+- Damap uses sequential integers as ids - to avoid conflicts, you should adpopt your own id system, that does not clash
+  with damap. Liquibase allows any type of id, as long as it is unique - e.g. tuwien_1, tuwien_2...
+
+If you want to completely do your own thing, keep everything above the same, but instead of including the changeLog-root.yaml
+file, include every changeset you want to keep separately. Beware - this approach requires a lot of maintenance, since
+every new damap version has to be checked for new changesets you might want to include.
+Also before version 3, all changesets where kept in db/changeLog.yaml. If you might want to remove certain changesets in
+this file, you would have to create your own version in your institutional folder. 
+
```
rekt-hard, line 133: minor: including a link to the liquibase documentation might be nice :)

rekt-hard (commented):
Good job! Just minor annotations and making sure we settle on a fixed name spacing

src/main/resources/application.yaml
ValentinFutterer: Would it be necessary in that case? I am not familiar with the reason why it is done to be honest. Is it to prevent conflicts with dependencies? Maybe you could give me an example?

rekt-hard: The namespacing with the URL makes it very explicit where something originates from in order to prevent conflicts or misunderstandings.
Imagine TU Graz, TU Wien and the open source DAMAP project creating source files under the `damap` namespace only. This would result in conflicts, if two files are named the same.

DAMAP: `damap/ProjectService.java`
TU Wien: `damap/ProjectService.java`
TU Graz: `damap/ProjectService.java`

Now, which one should be used when `damap/ProjectService.java` is included?
Additionally, other dependencies might provide a similar file structure, which would also lead to the same issue.

Maybe we could also namespace everything under the resources folder?
```
src
  - main
    - resources
      - at
        - ac
          - tuwien
            - damap
              - db
              - schema
              - spec
              - templates
        - tugraz
          - damap
              - db
              - schema
              - spec
              - templates
```

Up for discussion though :)

ValentinFutterer: I see where you are coming from. But I am not really familiar with how the institutional backends look. Do they have the same file structure as the damap project? If yes, then it should not be possible to create a file/folder with a duplicate path. Or is it just about being clear where the files originated? Maybe you should talk with @ZenoLC  about this, to find the best possible solution.

rekt-hard: It would be great to be able to create whatever folder structure a custom project wants to setup, without having to care about another one. If a custom project adds a `damap/projects.yaml`, which does not exist in the open source project at the time, everything works fine. Down the line, the open source project also adds a `damap/projects.yaml` file. Now the custom project does not build anymore and has to be adapted. This can be mitigated in the first place via namespacing.

ValentinFutterer: Which solution would you suggest? Should the namespacing start at db or at resources?
Also, what namespace should we use for the open source? db/at/ac/tuwien/damap would clash with the tuwien custom project.

rekt-hard: Putting it at resources ensures that it only has to be setup once per project and not repeated per folder. Not sure about it though, since I do not have that much experience with java projects.
Maybe `org.damap`? As said, not sure about the decision at all. Up for discussion with @ZenoLC as well (especially for the custom project)

ValentinFutterer: As discussed, I have namespaced the resources folder. The many changes are due to changing file locations / filepaths. Notice that I added a logicalFilePath attribute to changeLog.yaml. 
`- logicalFilePath:  db/changeLog.yaml`
This was necessary due to the way liquibase distinguishes between changesets, namely with id/author/filepath. The hash is used to detect if a changeset changed, but id/author/filepath determines the changeset identity. logicalFilePath makes it so that liquibase uses the path specified in the logicalFilePath attribute to ascertain identity - the physical file path stays as is (even in log files).

docker/docker-compose.yaml
rekt-hard, line 1: major: no need to update the docker-compose file in this PR

rekt-hard (commented):
Good job! Just make sure to reset the `docker-compose` file.

Tested migration with three local setups:
- no previous state, building completely new
- previous open source state
- custom project with additional changes
  - previous change set
  - custom change sets
  - this change set

Worked all times without any issues. Also tried exporting which worked flawlessly.

docker/docker-compose.yaml
ValentinFutterer: Right thanks for catching that, this went under with the many files that were changed.

rekt-hard (approved):

CUSTOMISING.md
ValentinFutterer: Done :)

CUSTOMISING.md
ValentinFutterer: Good idea :)

CUSTOMISING.md
```diff
@@ -110,6 +110,32 @@ damap:
...
+```yaml
+liquibase:
+  migrate-at-start: true
+  change-log: db/tuwien/newChangeLog-root.yaml
+```
+- Your new root changelog should look something like this:
+```yaml
+databaseChangeLog:
+  - include:
+      file: db/damap/changeLog-root.yaml
```
ZenoLC, line 127: this should be updated to the new address of the file
at/ac/tuwien/damap/db/changeLog-root.yaml

ZenoLC, line 127:
```suggestion
      file: at/ac/tuwien/damap/db/changeLog-root.yaml
```

ZenoLC (approved):
looks good, please fix just the minor address update in the CUSTOMISING instructions

CUSTOMISING.md
ValentinFutterer: Good catch, I have also updated the text above.

Issue #122, opened 2023-11-27
Update config for dev mode


ValentinFutterer:
**Description**
Currently, the dev flag does not do anything - to start the backend without docker, the config files have to be changed manually every time.

**Proposed solution**
Update dev mode in aplplication.yaml, so that everything is taken care of, when the project ist started via command line with `mvn compile quarkus:dev`.

PR #123, opened 2023-11-28
#122 - Updated the config for dev-mode

ValentinFutterer:
## Description
Config

#### What does this PR do?
Prev dev-mode config did not override enough. Configs like `fits-url` had to be overwritten manually. Now dev-mode overrides everything it needs to. The changes assume, that the user wants to start the backend and frontend together, outside of docker. To test this, one has to uncomment this part in the docker-compose.yaml file.
```
  # Uncomment the following to directly access the database
  # at localhost:8088
    ports:
     - "8088:5432"
```

#### Code review focus
I tested this with the docker-compose file in different states.
- Nothing uncommented
- The above lines uncommented
- Everything uncommented that can be

For every state, I started the application with docker and outside of docker.
It behaved like expected at every step.
Please make sure, that it does not interfere with custom projects.

closes GH-122

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -85,10 +85,16 @@ rest:
 "%dev":
   damap:
     origins: http://localhost:4200
+    auth:
+      backend:
+        url: http://localhost:8087/auth/realms/damap # https://your.authentication.server
     datasource:
-      url: jdbc:postgresql://localhost:5432/damap
+      url: jdbc:postgresql://localhost:8088/damap
       username: damap
       password: pw4damap
+    projects-url: http://localhost:8091
+    persons-url: http://localhost:8091
+    fits-url: http://localhost:8089/fits
 
 "%oracle":
   quarkus:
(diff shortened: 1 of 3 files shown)


src/main/resources/application.yaml
```diff
@@ -85,10 +85,16 @@ rest:
 "%dev":
   damap:
     origins: http://localhost:4200
+    auth:
+      backend:
+        url: http://localhost:8087/auth/realms/damap # https://your.authentication.server
     datasource:
-      url: jdbc:postgresql://localhost:5432/damap
+      url: jdbc:postgresql://localhost:8088/damap
```
rekt-hard, line 92: With all these changes, it also makes sense to remove the comment on the port and map the port 8088 open by default for the `damap-db` service in the `docker-compose` file.
https://github.com/tuwien-csd/damap-backend/blob/c9e43f8cd4c742652fa7162148bea986af119aa1/docker/docker-compose.yaml#L21-L24

rekt-hard (commented):
Works well locally - good job!

src/main/resources/application.yaml
ValentinFutterer: Yes that is a good idea. Out of curiosity, what was the reason for leaving it commented out?

rekt-hard (approved):

PR #128, opened 2023-12-13
103 orcid contributor affiliation selected randomly

rekt-hard:
#### What does this PR do?
Information for new contributors with ORCID identifier set will be fetched on DMP create/update. Affiliation will be selected to be current, with the earliest start date.

Since the information for email and the affiliations is scattered around the person attribute and the summary attribute, the whole ORCID record will be fetched. Required classes have been specified.

The ORCID test record resource is from the public ORCID API and the name has been modified.

#### Breaking changes

ORCID service will retrieve information from the `/<orcid>/record` endpoint.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [x] Tests added
- [ ] E2e tests created

closes https://github.com/tuwien-csd/damap-backend/issues/103


src/main/java/at/ac/tuwien/damap/rest/persons/orcid/ORCIDMapper.java
```diff
@@ -0,0 +1,121 @@
...
+
+        IdentifierDO identifierContributorDO = new IdentifierDO();
+        identifierContributorDO.setIdentifier(orcidRecord.getPerson().getName().getPath());
+        identifierContributorDO.setType(EIdentifierType.ORCID);
+
+        contributorDO.setPersonId(identifierContributorDO);
+        return contributorDO;
+    }
+
+    private final Comparator<? super ORCIDAffiliation> sortByCurrentStartDate = ((ORCIDAffiliation a,
```
ValentinFutterer, line 81: The ORCIDRecord does not contain this information, but maybe there is an easy way to fetch the source of an affiliation and give preferential treatment to those that were not created by the orcid person. See the picture, right now DAMAP chooses "Interdisziplinäre Werteforschung"  instead of "Universität Wien" as the affiliation. Just a suggestion, if it is easily doable.
image

src/test/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponentsTest.java
```diff
@@ -25,13 +27,17 @@ class AbstractTemplateExportScienceEuropeComponentsTest extends AbstractTemplate
...
+
     @BeforeEach
     public void setup() {
         Mockito.when(mockProjectService.read(anyString())).thenReturn(testDOFactory.getTestProjectDO());
+        Mockito.when(orcidPersonServiceImpl.read(any(String.class))).thenReturn(testDOFactory.getTestContributorDO());
     }
 
     @Test
-    void determinteDatasetIDsTest(){
+    void determinteDatasetIDsTest() {
```
ValentinFutterer, line 40: Small typo, should be `void determineDatasetIDsTest() {`.

src/main/java/at/ac/tuwien/damap/rest/persons/orcid/ORCIDMapper.java
```diff
@@ -0,0 +1,136 @@
...
+    public ContributorDO mapRecordEntityToPersonDO(ORCIDRecord orcidRecord,
+            ContributorDO contributorDO) {
+
+        contributorDO.setId(null);
+        contributorDO.setFirstName(orcidRecord.getPerson().getName().getGivenNames().getValue());
+        contributorDO.setLastName(orcidRecord.getPerson().getName().getFamilyName().getValue());
+
+        var primaryMail = orcidRecord.getPerson().getEmails().getEmail().stream().filter(ORCIDEmail::isPrimary)
+                .findFirst();
+        contributorDO.setMbox(primaryMail.isPresent() ? primaryMail.get().getEmail() : null);
```
ValentinFutterer, line 53: If a user only has a secondary email (e.g. the main one is private), then the second email shows up in the frontend. 
Before saving
But after saving the dmp, it is gone.
After saving
This should probably be changed to accept the secondary email, if the primary is not available.

src/main/java/at/ac/tuwien/damap/rest/dmp/service/DmpService.java
```diff
@@ -234,6 +245,21 @@ private void updateProjectLead(Dmp dmp) {
...
+        dmp.getContributorList().forEach(contributor -> {
+            // Existing contributor? Do not fetch data.
+            if (contributor.id != null) {
+                return;
+            }
+
+            var identifier = contributor.getPersonIdentifier();
+            if (identifier != null
+                    && identifier.getIdentifierType().equals(EIdentifierType.ORCID)) {
+                ContributorDOMapper.mapDOtoEntity(orcidPersonService.read(identifier.getIdentifier()), contributor);
```
ValentinFutterer, line 258: If you create a contributor manually, like so:
image
And try to save the dmp, it fails.
The problem is, that the PersonIdentifier != null, and the IdentifierType is ORCID. This leads to a get request in the OrcidPersonService with the orcid query parameter being "". This returns an uncaught 404 error.

ValentinFutterer (changes requested):

src/main/java/at/ac/tuwien/damap/rest/persons/orcid/ORCIDMapper.java
rekt-hard: Good catch, thank you. Changed it to take the first email, if the primary one is not available.

src/main/java/at/ac/tuwien/damap/rest/dmp/service/DmpService.java
rekt-hard: Thank you. Wrapped it in a general try/catch block.

src/main/java/at/ac/tuwien/damap/rest/persons/orcid/ORCIDMapper.java
ValentinFutterer: Perfect, now it takes the secondary email, if the primary is not available.
Small thing, but before saving, no email is shown in the frontend:
Before saving 2 0
After saving 2 0

src/test/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponentsTest.java
ValentinFutterer, line 1: Contributors

There appear two null values when exporting - I think the first one is part of the contributorId and the second one is due to contributorAffiliationId. Please take a look at it. These are probably appearing, bcause the ORCID API returns "null" as String, but I did not look into it in detail.

rekt-hard: This was related to the export and was fixed in the meantime

src/main/java/at/ac/tuwien/damap/rest/persons/orcid/ORCIDMapper.java
ValentinFutterer: Yes you are right, thanks for looking into it.

ValentinFutterer (approved):

PR #137, opened 2023-12-19
DmpVersion: Add foreign key revision entity

ValentinFutterer:
Relates to issue 39 in the frontend: https://github.com/tuwien-csd/damap-frontend/issues/39

## Description
Feature

#### What does this PR do?
Introduces a revision entity foreign key contraint to Dmpversion. Before just the revision entity id was stored in the DmpVersion. Sorting the list is not necessary, since it already is sorted correctly.

#### Breaking changes
Adapted DmpVersion entity + added liquibase changeset.

#### Code review focus
Please check if the liquibase changeset wont break anything in production.
In VersionDOMapper - getRevisionByRevisionNumber(), should we catch RevisionDoesNotExistException or do we assume that the frontend sends correct information always

#### Dependencies
An older frontend still works with these changes, so no.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-50

### DIFF
--- a/src/main/java/at/ac/tuwien/damap/domain/DmpVersion.java
+++ b/src/main/java/at/ac/tuwien/damap/domain/DmpVersion.java
@@ -25,12 +25,17 @@ public class DmpVersion extends PanacheEntity {
     @EqualsAndHashCode.Exclude
     private Dmp dmp;
 
-    @Column(name = "revision_number")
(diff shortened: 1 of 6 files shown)


rekt-hard (approved):
looks good!

Issue #132, opened 2023-12-15
Export: Empty table(s)


rekt-hard:
In area 4 there is an empty table if no data is reused. This should simply not be present in this case.
There might also be other tables which are empty, if no data is selected.
![image](https://github.com/tuwien-csd/damap-backend/assets/72449192/1e9a6c16-dfb1-414a-a295-19aaf0a581cb)

PR #138, opened 2024-01-08
Export: Delete delete/reused datasets tables when empty

ValentinFutterer:
## Description
Bugfix

#### What does this PR do?
The reused dataset and dataset delete tables are now removed when there is no data for them.

### Checks

- [x] Export updated
- [x] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-132

### DIFF
--- a/src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportFunctions.java
+++ b/src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportFunctions.java
@@ -2,6 +2,7 @@ package at.ac.tuwien.damap.conversion;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.reflect.Field;
 import java.util.*;
 import java.text.SimpleDateFormat;
 
@@ -35,7 +36,7 @@ public abstract class AbstractTemplateExportFunctions {
      * @return
      * @throws Exception
      */
-    public XWPFDocument loadTemplate (InputStream template, String startChar, String endChar) throws Exception{
+    public XWPFDocument loadTemplate(InputStream template, String startChar, String endChar) throws Exception {
         //Extract document using Apache POI https://poi.apache.org/
         XWPFDocument document = new XWPFDocument(template);
 
@@ -119,7 +120,7 @@ public abstract class AbstractTemplateExportFunctions {
      * @throws XmlException
      * @throws Exception
      */
-    public XWPFTableRow insertNewTableRow(XWPFTableRow sourceTableRow, int pos) throws XmlException, IOException  {
+    public XWPFTableRow insertNewTableRow(XWPFTableRow sourceTableRow, int pos) throws XmlException, IOException {
         XWPFTable table = sourceTableRow.getTable();
         CTRow newCTRrow = CTRow.Factory.parse(sourceTableRow.getCtRow().newInputStream());
         XWPFTableRow tableRow = new XWPFTableRow(newCTRrow, table);
@@ -152,7 +153,7 @@ public abstract class AbstractTemplateExportFunctions {
      * @param table
      * @param replacements
      */
-    static void replaceTableVariables(XWPFTable table, Map<String, String> replacements){
+    static void replaceTableVariables(XWPFTable table, Map<String, String> replacements) {
         //this replaces variables in tables (e.g. costcurrency)
         List<XWPFTableRow> tableRows = table.getRows();
         for (XWPFTableRow xwpfTableRow : tableRows) {
@@ -260,7 +261,7 @@ public abstract class AbstractTemplateExportFunctions {
      * @param startChar
      * @param endChar
      */
-    public void formattingTable(List<XWPFTable> xwpfTables, String startChar, String endChar){
+    public void formattingTable(List<XWPFTable> xwpfTables, String startChar, String endChar) {
         if (xwpfTables != null) {
             for (XWPFTable xwpfTable : xwpfTables) {
                 for (XWPFTableRow row : xwpfTable.getRows()) {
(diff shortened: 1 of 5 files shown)


src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -876,13 +882,14 @@ public void composeTableReusedDatasets(XWPFTable xwpfTable){
                 insertTableCells(xwpfTable, newRow, docVar);
             }
             xwpfTable.removeRow(xwpfTable.getRows().size() - 1);
+            xwpfTable.removeRow(1);
+            replaceTableVariables(xwpfTable, replacements);
         } else {
-            //clean row
             ArrayList<String> emptyContent = new ArrayList<String>(Arrays.asList("", "", "", "", "", ""));
             insertTableCells(xwpfTable, xwpfTable.getRows().get(xwpfTable.getRows().size() - 1), emptyContent);
```
rekt-hard, line 889: Can these be removed as the table and the paragraph will be removed anyway a line after?

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -813,6 +818,7 @@ public void composeTableNewDatasets(XWPFTable xwpfTable){
         }
         //end of dynamic table rows code
         xwpfTable.removeRow(1);
+        replaceTableVariables(xwpfTable, replacements);
```
rekt-hard, line 821: Tables are composed and filled in a general function. So we should either call `replaceTableVariables` in each specific compose method (as is done here) or just after calling the compose method (as is done here https://github.com/tuwien-csd/damap-backend/blob/ecead07507490d78a3c50e72030c1085c4784ce1/src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java#L749)

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -1140,12 +1152,16 @@ public void composeTableDatasetDeletion(XWPFTable xwpfTable){
                 insertTableCells(xwpfTable, newRow, docVar);
             }
             xwpfTable.removeRow(xwpfTable.getRows().size() - 1);
+            xwpfTable.removeRow(1);
+            replaceTableVariables(xwpfTable, replacements);
         } else {
             //clean row
             ArrayList<String> emptyContent = new ArrayList<String>(Arrays.asList("", "", "", "", ""));
             insertTableCells(xwpfTable, xwpfTable.getRows().get(xwpfTable.getRows().size() - 1), emptyContent);
```
rekt-hard, line 1160: Same as above: Can it be removed?

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -559,7 +560,7 @@ public void repoinfoAndToolsInformation() {
         }
     }
 
-    public void legalEthicalInformation() {
+    public void  legalEthicalInformation() {
```
rekt-hard, line 563: typo

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
ValentinFutterer: image

Unfortunately, this instance of `replaceTableVariables` cannot be called inside the specific compose method, since it is meant to fill tables that do not have a specific compose method. 

An alternative would be to have the compose methods return a boolean, to indicate if the table was deleted or not. Then we can call `replaceTableVariables` after the switch statement.

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
ValentinFutterer: Good catch

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
rekt-hard: Ah right, we remove tables now. Will it throw, if we remove a table and then call `replaceTableVariables` on it?

ValentinFutterer: Yes, the problem here is, that removing a table from a document removes the underlying xml, but the java obejct stays. replaceTableVariables tries to access the text (xml), which causes an exception. I have now changed it so that replaceTableVariables is called after all tables are composed and only for tables that still exist.

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportFunctions.java
```diff
@@ -326,4 +325,143 @@ public void formattingParagraph(List<XWPFParagraph> xwpfParagraphs, String start
...
+
+    /**
+     * Removes a paragraph inside a document or inside a table.
+     * This will not work for paragraphs inside nested tables.
+     * If the paragraph is not to be found inside the doc or inside non nested tables, nothing happens.
+     *
+     * @param doc
+     * @param paragraph
+     */
+    public void removeParagraph(XWPFDocument doc, XWPFParagraph paragraph) {
```
rekt-hard, line 380: Is this method used anywhere?

ValentinFutterer: No I wrote it in an earlier attempt on solving this and kept it in since it could be useful in the future.

rekt-hard: If it is not used, it should be removed

ValentinFutterer: Done

rekt-hard (approved):
good job!

Issue #139, opened 2024-01-08
Export: Science Europe placeholders are not removed


ValentinFutterer:
**Describe the bug**
When exporting to the Science Europe template and no start / enddate is given, the placeholders remain in the table.

**Is this a regression?**
No.

**Steps to Reproduce**
1. Export a DMP to Scienece Europe with just the project name filled out.

**Expected behavior**
The table cells should be empty.

**Screenshots**
<img width="539" alt="image" src="https://github.com/tuwien-csd/damap-backend/assets/80054001/b6009c3e-adab-44fa-a080-b7d9509d9991">

**Environment**
 - Browser chrome Version 120.0.6099.130
 - Damap version 2.6.0

PR #140, opened 2024-01-08
Export: Correctly replace date placeholders

ValentinFutterer:
## Description
Bugfix

#### What does this PR do?
In the Science Europe template, the [startdate] and [enddate] placeholders are now correctly replaced with "", if no dateswhere given. Before. the placeholders just remained in the word document.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-139

### DIFF
--- a/src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -103,10 +103,8 @@ public abstract class AbstractTemplateExportScienceEuropeComponents extends Abst
         }
 
         // variable project start date and end date
-        if (project.getStart() != null)
-            addReplacement(replacements, "[startdate]", formatter.format(project.getStart()));
-        if (project.getEnd() != null)
-            addReplacement(replacements, "[enddate]", formatter.format(project.getEnd()));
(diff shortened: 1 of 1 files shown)


src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -103,11 +103,8 @@ public void titlePage() {
         }
 
         // variable project start date and end date
-        if (project.getStart() != null)
-            addReplacement(replacements, "[startdate]", formatter.format(project.getStart()));
-        if (project.getEnd() != null)
-            addReplacement(replacements, "[enddate]", formatter.format(project.getEnd()));
-
+        addReplacement(replacements, "[startdate]", project.getStart() == null ? null : formatter.format(project.getStart()));
+        addReplacement(replacements, "[enddate]", project.getStart() == null ? null : formatter.format(project.getEnd()));
```
rekt-hard, line 107: Would it make sense to be more specific about the value and set it to "" explicitly? `addReplacement` implementation might change in the future, so we are sure what will be inserted.

ValentinFutterer: Yes very true, good suggestion.

rekt-hard (approved):

PR #153, opened 2024-01-31
Deployment enhancements

SotosTsepe:
#### What does this PR do?
* Adds start dependencies between the services (see below).
* Adds healthcheck for damap-backend (fixed previous one) and Postgres.
* Sets auto restart policy unless explicitly stopped (this would help specifically when deploying to a live server so that we don't have to restart the services ourselves on every potential reboot).

#### Problems addressed
* Previously, the damap-backend was crashing because it was starting earlier than the Postgres database. To make the whole stack work properly, one had to first start manually Postgres and then the other services.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,7 +65,4 @@ EXPOSE 8080
 # for Openshift based unprivilegued Kubernetes environments, we will set the user to 1001
 USER 1001
 
-HEALTHCHECK --interval=5m --timeout=3s \
-  CMD curl -f http://localhost/q/health/live || exit 1
-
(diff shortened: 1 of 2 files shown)


ValentinFutterer:
I like the changes.
Quick question: is it necessary that the frontend depends on the backend? One can just start the frontend container without it crashing.

SotosTsepe:
Yes, because the frontend does HTTP calls to the backend. If the backend takes too long to start or exits with an error, the frontend should not start as well.

rekt-hard (approved):
Great, thank you!
This should ensure, that starting all default services is more consistent across devices

PR #166, opened 2024-02-29
tests: Add create DmpDO method and refactor

rekt-hard:
## Description

Bugfix

#### What does this PR do?

Add and use method to create a new DMP for tests.

In a future refactor, all tests should be updated to create a new DMP specifically for the test case. Relying on a certain DMP to be present and modifying the same DMP in several test cases might lead to unforeseen issues.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-


ValentinFutterer (approved):
Looks good!

Issue #164, opened 2024-02-28
PersonService: Trying to edit access to a DMP leads to a 500er


ValentinFutterer:
**Describe the bug**
Trying to edit the access to a DMP leads to a 500 error and no options are shown. 

**Is this a regression?**
Yes.

**Steps to Reproduce**
Try to edit the access to a DMP.

**Expected behavior**
Should show a list of demo persons.

**Environment**
 - Chrome 122.0.6261.58
 - Damap version 3.0.0

PR #167, opened 2024-03-04
MockPersonService: Add additional example users

ValentinFutterer:
## Description

Bugfix/Feature

#### What does this PR do?
Added additional testusers and updated the universityId of the original testuser to match the keycloak personId, which fixed the bug.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-164


rekt-hard (commented):
Great, thank you!

Can you please also add an entry for the admin user? Since we can sign in to the application and create a DMP with the admin account, it should also be able to manage access with it.

PR #177, opened 2024-04-02
config: indicate availability of FITS service

### LINKED ISSUE #59: disable FITS component/tab if FITS is not available (currently backend crashes)

larnhold:
#### What does this PR do?

Adds a new field to ConfigDO indicating if a non-empty URL "fits-url" has been defined in application.yaml

#### Breaking changes

No breaking changes

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-59

### DIFF
--- a/src/main/java/at/ac/tuwien/damap/rest/ConfigResource.java
+++ b/src/main/java/at/ac/tuwien/damap/rest/ConfigResource.java
@@ -12,6 +12,9 @@ import at.ac.tuwien.damap.rest.config.domain.ConfigDO;
 import at.ac.tuwien.damap.rest.config.domain.PersonServiceConfigurations;
 import lombok.extern.jbosslog.JBossLog;
 
+import java.net.URL;
+import java.util.Optional;
+
 @Path("/api/config")
 @Produces(MediaType.APPLICATION_JSON)
 @PermitAll
@@ -35,6 +38,9 @@ public class ConfigResource {
     @ConfigProperty(name = "damap.person-services")
     PersonServiceConfigurations personServiceConfigurations;
 
+    @ConfigProperty(name = "damap.fits-url")
+    Optional<URL> fitsUrl;
+
     @GET
     public ConfigDO config() {
         ConfigDO configDO = new ConfigDO();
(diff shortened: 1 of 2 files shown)


ValentinFutterer (commented):
I would suggest to check if the fits service is running and available, since currently it is only checked if the service has a url in the application.yaml. If the fits container crashes, the frontend still displays the fits component but trying to use it causes an error.

rekt-hard:
The same can be said about all other services the back end interacts with. It might be that a service is available one second and unavailable the next second. We would have to monitor a services availability, update the configs accordingly and let the front end know (or periodically fetch the current status from the front end). Being this dynamic and reactive requires thorough changes in the front end and back end.

For now, I think it is a fair assumption that if a URL is specified, a service should be available in general.

src/main/java/at/ac/tuwien/damap/rest/ConfigResource.java
```diff
@@ -44,7 +50,12 @@ public ConfigDO config() {
         configDO.setAuthUser(authUser);
         configDO.setEnv(env);
         configDO.setPersonSearchServiceConfigs(personServiceConfigurations.getConfigs());
+        configDO.setFitsServiceAvailable(getFitsServiceAvailability());
 
         return configDO;
     }
+
+    private Boolean getFitsServiceAvailability() {
+        return fitsUrl.isPresent() && !fitsUrl.toString().isEmpty();
```
rekt-hard: URL performs some checks on the provided string (e.g. protocol, address) and will throw, if it is not a valid URL. Therefore, the second part of the condition can be omitted. What do you think?

src/main/java/at/ac/tuwien/damap/rest/config/domain/ConfigDO.java
```diff
@@ -16,4 +16,5 @@ public class ConfigDO {
     private String authUser;
     private String env;
     private List<ServiceConfig> personSearchServiceConfigs;
+    private Boolean fitsServiceAvailable;
```
rekt-hard: I would lean towards using primitive types wherever possible :)

rekt-hard (commented):
Nice work. Just minor comments

src/main/java/at/ac/tuwien/damap/rest/ConfigResource.java
```diff
@@ -44,7 +50,12 @@ public ConfigDO config() {
         configDO.setAuthUser(authUser);
         configDO.setEnv(env);
         configDO.setPersonSearchServiceConfigs(personServiceConfigurations.getConfigs());
+        configDO.setFitsServiceAvailable(getFitsServiceAvailability());
 
         return configDO;
     }
+
+    private Boolean getFitsServiceAvailability() {
```
ValentinFutterer: Can you change this one to boolean too :)

rekt-hard (approved):
Looking good, thank you!

ValentinFutterer (approved):
Nice!

PR #179, opened 2024-04-03
Service: Fix bug where contributor email disappears

### LINKED ISSUE #178: Email disappears for contributor if they have ORCID ID
**This is a copy of https://github.com/tuwien-csd/damap-frontend/issues/190 , since the issue is backend related.**

Describe the bug
When adding a project contributor who has an ORCID ID and an email address, the email address disappears. This happens both in DAMAP and the TU Wien DMP Tool

Is this a regression?
No.

To Reproduce

Go to 'People involved in data management'
Search for a person from the TU Wien address book that has an ORCID ID and email address, e.g. Tomasz or Christiane. OR manually add a person with and email address and ORCID ID
Click "Save changes"
See email disapear
Expected behavior
All info input stays in tool, including email address

Environment

Browser Edge version 122
Damap and TU Wien DMP Tool
Additional context

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?
https://github.com/tuwien-csd/damap-backend/pull/128 introduced the fetching of ORCID-emails on Contributor create or update. The fetched email was alaways used instead of the current one, leading to it being overwritten - most likely with null.
Now the ORCID email is only used when no other email is present (this leads to ORCID email updates not being picked up by the tool, but this should not be a big problem in my opinion).

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-178

### DIFF
--- a/src/main/java/at/ac/tuwien/damap/rest/dmp/mapper/ContributorDOMapper.java
+++ b/src/main/java/at/ac/tuwien/damap/rest/dmp/mapper/ContributorDOMapper.java
@@ -62,6 +62,7 @@ public class ContributorDOMapper {
             contributor.setAffiliationId(affiliationIdentifier);
         } else
             contributor.setAffiliationId(null);
+
         contributor.setContact(contributorDO.isContact());
         contributor.setContributorRole(contributorDO.getRole());
 
--- a/src/main/java/at/ac/tuwien/damap/rest/dmp/service/DmpService.java
+++ b/src/main/java/at/ac/tuwien/damap/rest/dmp/service/DmpService.java
@@ -12,6 +12,8 @@ import javax.inject.Inject;
 import javax.transaction.Transactional;
 import javax.validation.Valid;
 
+import at.ac.tuwien.damap.domain.Identifier;
+import at.ac.tuwien.damap.rest.dmp.mapper.*;
 import org.hibernate.envers.AuditReader;
 import org.hibernate.envers.AuditReaderFactory;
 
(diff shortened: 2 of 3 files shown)


rekt-hard (commented):
Thank you for addressing this issue. Works fine.

Can you also address other fields? The `personIdentifier` and the `affiliation` might be set via the university already and would be overridden by an ORCID search.

Right now, we do not allow to modify already added contributors via the UI (but would be possible through the API). If we allow edits through the UI, we have to revisit this and make sure that new data takes precedence over old data.

ValentinFutterer:
> Can you also address other fields? The `personIdentifier` and the `affiliation` might be set via the university already and would be overridden by an ORCID search.

Good point, I did not think about the other fields :)

ValentinFutterer:
The ContributorDO generated from ORCID has all fields set to null other than name, affiliation, personID and mbox. When the ContributorDOMapper is used, these null values override legit fields. What do you think of the current (naive) solution? The logic should probably be in its own mapper. Should we just add a method to ORCIDMapper like `mapORCIDPersonDOToEntity`? But this would just chain mapper calls. Maybe adjust `mapRecordEntityToPersonDO(ORCIDRecord orcidRecord,
            ContributorDO contributorDO)` to `mapRecordEntityToPersonEntity(ORCIDRecord orcidRecord,
            Contributor contributor)` and move the new logic in this mapper method.

rekt-hard:
> The ContributorDO generated from ORCID has all fields set to null other than name, affiliation, personID and mbox. When the ContributorDOMapper is used, these null values override legit fields. What do you think of the current (naive) solution?

Seems fine to be honest. This is a special case anyway. Adapting a general mapper would most likely lead to future issues.

rekt-hard (approved):

PR #196, opened 2024-06-17
Export: Add PID information to reused datasets table

### LINKED ISSUE #98: Add PID information to the repository table in the export templates
**Changes**
The export templates have a table listing repositories in which datasets are stored.
The PID column is empty/not being filled out. This information should be availyble through the r3data API.
Extract it and add it to the template.
PID example

**Additional notes**
Depending on what is implemented already this task might be easy, by just supplying the information to the template, or more complex, if the PID is not collected already.

**Assumptions**
The templates themselves should not be changed in order to implement this feature.

### LINKED ISSUE #115: Insert PID type in template dataset publication tables
**Description**
Affects science europe table 5a (Data publication and access conditions) and equivalent IV.2 in FWF template.
Insert in the PID column the PID type information  ("PID Systems") extracted from r3data.

**Proposed solution**
Perceive the PID type from r3data.

**Alternative solution**
If not available in r3data then allow implementing institutions to have a list of repositories that will provide a hardcoded information.
e.g. TUWRD would dictate PID type "DOI"

**Additional context**
Currently the field is empty, and researchers do not know what input is required.

_If this issue requests multiple changes consider using checkboxes or lists._

ValentinFutterer:
## Description

Feature

#### What does this PR do?
Takes the supported PID-System from the repository information available through R3Data and fills the info in the PID column in the templates. Since R3Data supported more identifiers than damap had as enums, a database change was necessary to update them. The enums are HDL, PURL and URN.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-98, GH-115

### DIFF
--- a/src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportFunctions.java
+++ b/src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportFunctions.java
@@ -208,24 +208,18 @@ public abstract class AbstractTemplateExportFunctions {
      * @param variableList
      * @return
      */
-    public String multipleVariable(List<String> variableList) {
-        switch (variableList.size()) {
-            case 0:
-                return "";
-            default:
-                return String.join(", ", variableList);
-        }
+    public String joinWithComma (List<String> variableList) {
+        return String.join(", ", variableList);
     }
 
-    public String multipleVariableAnd(List<String> variableList) {
-        switch (variableList.size()) {
-            case 0:
-                return "";
-            case 1:
-                return variableList.get(0);
-            default:
-                return String.join(", ", variableList.subList(0, (variableList.size() - 1))) + ", and " + variableList.get(variableList.size() - 1);
-        }
+    public String joinWithCommaAnd(List<String> variableList) {
+        return switch (variableList.size()) {
+            case 0 -> "";
+            case 1 -> variableList.get(0);
+            default ->
+                    String.join(", ", variableList.subList(0, (variableList.size() - 1))) + " and "
+                            + variableList.get(variableList.size() - 1);
+        };
     }
 
     /**
--- a/src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -154,7 +154,7 @@ public abstract class AbstractTemplateExportScienceEuropeComponents extends Abst
         String contactAffiliationId = "";
 
         if (contact == null) {
-            addReplacement(replacements, "[contact]", multipleVariable(contactItems));
+            addReplacement(replacements, "[contact]", joinWithComma(contactItems));
             return;
         }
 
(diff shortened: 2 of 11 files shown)


src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportFunctions.java
```diff
@@ -208,23 +208,19 @@ public void replaceTextInFooter(XWPFDocument doc, Map<String, String> replacemen
...
-                return variableList.get(0);
-            default:
-                return String.join(", ", variableList.subList(0, (variableList.size() - 1))) + ", and " + variableList.get(variableList.size() - 1);
+    public String joinWithCommaAnd(List<String> variableList) {
+        if (variableList.isEmpty()) {
+            return "";
+        } else if (variableList.size() == 1) {
+            return variableList.get(0);
+        } else if (variableList.size() == 2) {
+            return variableList.get(0) + " and " + variableList.get(1);
```
rekt-hard: Is this already covered by fall back else?

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportFunctions.java
```diff
@@ -208,23 +208,19 @@ public void replaceTextInFooter(XWPFDocument doc, Map<String, String> replacemen
...
-    public String multipleVariableAnd(List<String> variableList) {
-        switch (variableList.size()) {
-            case 0:
-                return "";
-            case 1:
-                return variableList.get(0);
-            default:
-                return String.join(", ", variableList.subList(0, (variableList.size() - 1))) + ", and " + variableList.get(variableList.size() - 1);
+    public String joinWithCommaAnd(List<String> variableList) {
+        if (variableList.isEmpty()) {
```
rekt-hard: minor: I prefer to use a switch on the size just to be clear about which value changes the outcome. Up to you

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -297,7 +297,7 @@ public void contributorInformation(){
     public void costInformation(){
         String costs = "";
         if (dmp.getCostsExist() != null) {
-            if (dmp.getCostsExist().booleanValue()) {
+            if (dmp.getCostsExist()) {
                 costs = loadResourceService.loadVariableFromResource(prop, "costs.avail");
             }
             else {
```
rekt-hard: How about we simplify this to
```java
        if (Boolean.TRUE.equals(dmp.getCostsExist()))
```

Then we can get rid of the nested conditional

rekt-hard (commented):
nice work! just a small annotation

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportFunctions.java
ValentinFutterer: yeah that's fair, i will change it

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportFunctions.java
ValentinFutterer: you are right, whoops :)

src/main/java/at/ac/tuwien/damap/conversion/AbstractTemplateExportScienceEuropeComponents.java
ValentinFutterer: Good catch!

rekt-hard (approved):

PR #234, opened 2024-08-05
export: add description to dataset tables

### LINKED ISSUE #230: Export: Dataset description not exported
**Describe the bug**
Dataset description fields are not found in the exported documents.

**Is this a regression?**
No.

**Steps to Reproduce**
1. Create a DMP
2. Add dataset with description
3. Export document
4. Open document
5. Scroll to dataset table

**Expected behavior**
Entered dataset descriptions should be exported to the documents

**Screenshots**
image

**Additional context**

rekt-hard:
check number of table row cells and provided values on row replacement

## Description

Bugfix

#### What does this PR do?

When the dataset tables were created on export, the description of the datasets was not added to the tables.
Descriptions are now added in a new column at the end of the table.

These are the templates with track changes enabled:
~[FWFTemplate.docx](https://github.com/user-attachments/files/16635109/FWFTemplate.docx)~
~[horizonEuropeTemplate.docx](https://github.com/user-attachments/files/16635112/horizonEuropeTemplate.docx)~
~[scienceEuropeTemplate.docx](https://github.com/user-attachments/files/16635092/scienceEuropeTemplate.docx)~

Updated templates with tracked changes:
[FWFTemplate_changes.docx](https://github.com/user-attachments/files/16675550/FWFTemplate_changes.docx)
[horizonEuropeTemplate_changes.docx](https://github.com/user-attachments/files/16675552/horizonEuropeTemplate_changes.docx)
[scienceEuropeTemplate_changes.docx](https://github.com/user-attachments/files/16675553/scienceEuropeTemplate_changes.docx)

FWF
Screenshot from 2024-08-05 14-14-32

Horizon Europe
Screenshot from 2024-08-05 14-15-13

Science Europe
Screenshot from 2024-08-05 14-15-52

When inserting values into a row of a table, a check will be performed on the number of cells and the number of provided values. If the values mismatch, a warning will be logged. The warning consists of the text content from the first row of the table (for info and easier problem finding) as well as the number of cells and number of provided values.

image

#### Breaking changes

None

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes #230


rekt-hard:
The last commit adds support for multi line comments (introduced by tuwien-csd/damap-frontend/issues/278)
image

ValentinFutterer:
Good work, nice refactoring. Three things to note: 
- Are the changes from https://github.com/tuwien-csd/damap-backend/pull/235 included in the template changes? If not, could you please add them. 
- On the same note, could you supply a Word document with all changes marked in the review mode, so i can compile them for the release?
- The table cells had a little spacing to the left before, now the text starts exactly at the cell wall.

rekt-hard:
> Good work, nice refactoring. Three things to note:
> 
>     * Are the changes from [Export: Fixed alignment of reused data table in SE template #235](https://github.com/tuwien-csd/damap-backend/pull/235) included in the template changes? If not, could you please add them.
Changes are added. I think otherwise there would be a conflict reported on this PR as well (for binary files at least).
> 
>     * On the same note, could you supply a Word document with all changes marked in the review mode, so i can compile them for the release?
How do I do that? I also do not have Microsoft Office here so support may be different.
Would it make sense to just do that at a release? I.e. open the document of the last release, copy the content of the document of this release and then generate the changes document?
> 
>     * The table cells had a little spacing to the left before, now the text starts exactly at the cell wall.
Can you provide screenshots? I just checked an older export and it looks the same.

ValentinFutterer:
> ```
> How do I do that? I also do not have Microsoft Office here so support may be different. Would it make sense to just do that at a release? I.e. open the document of the last release, copy the content of the document of this release and then generate the changes document?
> ```
Are you using Libre Office Writer? I would prefer to collect all changes in a single place during development. Going through all documents after the fact is more work and also error prone - e.g. changing table settings would not be caught this way. In this case I could create the review document myself, since the only changes are one added column. But in the future we need a solution for this.
> ```
> Can you provide screenshots? I just checked an older export and it looks the same.
> ```

image

image

As an example from horizon europe template. It doesnt seem to affect every table the same way though - e.g. the reused dataset table stayed the same. Also notice that the license is not a hyperlink anymore after the changes.

rekt-hard:
> > ```
> > How do I do that? I also do not have Microsoft Office here so support may be different. Would it make sense to just do that at a release? I.e. open the document of the last release, copy the content of the document of this release and then generate the changes document?
> > ```
> 
> Are you using Libre Office Writer? I would prefer to collect all changes in a single place during development. Going through all documents after the fact is more work and also error prone - e.g. changing table settings would not be caught this way. In this case I could create the review document myself, since the only changes are one added column. But in the future we need a solution for this.

I am using Libre Office Writer.
Is it not more work to document every single change instead of documenting all changes at once for a release? I would assume tracking the changes in review mode by copying the content of the final version into the old version should do the trick. Then this would have to be done only once and not for every change to an export document.
> 
> > ```
> > Can you provide screenshots? I just checked an older export and it looks the same.
> > ```
> 
> image image
> 
> As an example from horizon europe template. It doesnt seem to affect every table the same way though - e.g. the reused dataset table stayed the same. Also notice that the license is not a hyperlink anymore after the changes.

Thank you. Have not touched that table so it is strange that it would change. Will have a look

rekt-hard:
I just checked with an older version (4.0.2) and the exported documents look the same when comparing the same template (except for the description column).
Licenses are also a link in all exported documents.

Sorry, seems like I can not reproduce the behaviour. Please feel free to update the templates, if this issue still occurs.

rekt-hard:
@ValentinFutterer I have modified the files again with the `Track changes` mode enabled. They are attached in the first comment.

I also understand now why simply tracking changes and copying the whole content from one version to another does not work. Track changes will only document two changes: Delete everything and insert everything. So not very helpful indeed :smile:

ValentinFutterer (approved):
I have added the new templates + the change files. I also fixed a small bug with fontsizes and changed the table formatting. The formatting breaks from before are gone now.

PR #236, opened 2024-08-06
nginx: add rules for openapi docs

SotosTsepe:
Makes swagger UI and OpenAPI accessible in a dockerized setup.


rekt-hard (approved):
great, thank you!

PR #238, opened 2024-08-12
40 add a format value to dataset info

### LINKED ISSUE #40: Add a format value to dataset info
Datasets should have an additional free text input, where data format can be set (e.g. 'PDF').
Ideally the use of FITS component to automatically determine the data type should fill this field out automatically as well.

rekt-hard:
## Description

Feature

#### What does this PR do?

Add file format attribute to dataset.
Update fits mapper to use mime type for setting this attribute.
Set attribute to empty string for OpenAIRE dataset, as this info is not available via the API.
Update export to set the file format for new datasets table.

Restructure change log imports to import one change log per major version. This keeps the root change log as short as possible.

image

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes https://github.com/tuwien-csd/damap-frontend/issues/40


src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.x.yaml
```diff
@@ -0,0 +1,7 @@
+databaseChangeLog:
```
ValentinFutterer, line 1: I do not think further splitting of the changesets is strictly necessary, since there number will always be relatively small. Though, i do not have anything against it either. If you want to keep it, you should write a sentence or two in our documentation of the liquibase workflow. You could also add a sentence about always writing a rollback statement, when necessary.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.x.yaml
```diff
@@ -0,0 +1,7 @@
...
+  - include:
+      file: org/damap/base/db/changeLog-4.x/changeLog-4.0.0_1.yaml
+  - include:
+      file: org/damap/base/db/changeLog-4.x/changeLog-4.0.0_2.yaml
+  - include:
+      file: org/damap/base/db/changeLog-4.x/changeLog-4.0.0_3.yaml
```
ValentinFutterer: Small mistake, this should be 4.0.3_1.yaml.

src/main/java/org/damap/base/rest/openaire/mapper/OpenAireMapper.java
```diff
@@ -125,6 +127,7 @@ public DatasetDO mapAtoB(String propertyName, QualifierType qualifierType, Datas
    * @return a {@link org.damap.base.rest.dmp.domain.DatasetDO} object
    */
   public DatasetDO mapAtoB(String propertyName, String string, DatasetDO datasetDO) {
+    // TODO: check if file format is provided
```
ValentinFutterer: File format is provided as `format`. See target/generated-sources/jaxb/eu/openaire/oaf/Result.java or https://www.openaire.eu/schema/1.0/doc/oaf-result-1_0_xsd.html#result_format. I tested it and the return value was a mime type - although it is not stated that this is always the case.

ValentinFutterer:
Also general warning: https://github.com/tuwien-csd/damap-backend/pull/227 has liquibase chnages, this will lead to conflicts, so whoever merges it should proceed with caution.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.x.yaml
rekt-hard: Added documentation for it in the change log root file.

src/main/java/org/damap/base/rest/openaire/mapper/OpenAireMapper.java
rekt-hard: Thank you. I could not find a dataset, where this is set. So I just used the same logic of trying to parse a mime type.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.x.yaml
rekt-hard: Updated to take a previous change log into account. Also updated the id to 13

ValentinFutterer: Perfect

src/main/java/org/damap/base/rest/openaire/mapper/OpenAireMapper.java
ValentinFutterer: Yes i think this reasonable

ValentinFutterer:
I searched a little and found a dataset on OpenAire with a format: `10.48436/tkkfs-11b75`
I tested it with this DOI and the dataset had null has fileFormat. The same happened when I added a dataset through the fits service. But inside the OpenAireMapper, the DatasetDO gets assigned the fileFormat correctly - there must be a bug somewhere else.

src/main/java/org/damap/base/rest/fits/mapper/FitsMapper.java
```diff
@@ -26,6 +26,9 @@ public Dataset mapAtoB(Fits fits, Dataset dataset) {
     dataset.setSize(getSize(fits));
     IdentificationType.Identity identity = getMajorityVoteIdentity(fits);
     dataset.setType(mapFileFormat(identity));
+    // Set file format based on mime type
```
ValentinFutterer, line 29: I checked the [RDA common standard](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard?tab=readme-ov-file#distribution_format) and found that fileFormat should be formated like a mime type when possible. So the string splitting should be removed.

ValentinFutterer: Ah and now that i think about it, the frontend should probably give instructions to use a mime type when filling out the file format field.

rekt-hard: set it to the mime type for OpenAIRE mapper and FITS mapper

src/main/java/org/damap/base/rest/dmp/mapper/DatasetDOMapper.java
```diff
@@ -116,6 +117,7 @@ public Dataset mapDOtoEntity(DatasetDO datasetDO, Dataset dataset, MapperService
               }
             });
     dataset.setType(typeList);
+    dataset.setFileFormat(datasetDO.getFileFormat());
```
ValentinFutterer, line 120: https://github.com/tuwien-csd/damap-backend/pull/238#issuecomment-2288044778

ValentinFutterer (approved):
👍

PR #242, opened 2024-08-20
docker: move extra services to own file

rekt-hard:
#### What does this PR do?

Split `docker-compose.yaml` file. One compose file only includes the services. For development, it is usually not relevant to have an extra backend and frontend running, since the local development instances for backend and frontend are used.

To run only the extra services, the following command can be executed:
`docker-compose -f docker-compose-services.yaml up`

#### Breaking changes

None

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -18,7 +18,16 @@ To start up the cluster of components just issue the following command:
 
 ```shell
 cd docker
-docker-compose up -d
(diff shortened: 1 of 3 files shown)


TomMiksa (approved):
Good job! :)

PR #246, opened 2024-08-27
Global: Update internal storage and add CRUD operations

### LINKED ISSUE #95: Allow internal storages to be disabled for the frontend
**Description**
Currently the design of the application does not allow to retire internal storage options from the available list of options.
Simply deleting them will result in a foreign key conflict.
So we must be able to remove them in a way that would allow continued support for old DMPs that have these options selected, while removing it from the list of selectable options available in the frontend.

**Proposed solution**
Possibly add a column (e.g. "active") to the database, to indicate if a storage option is deprecated.
The frontend needs to hide this option if tagged as inactive.
We still need to communicate this option to the frontend, in order for it to display it if it was selected in a past DMP.
The backend validator needs to alert, in case the user attempts to indicate an inactive storage option in their request.

**Alternative solution**
The above solution is from the top of my head, please feel free to use different approaches.

**Additional context**
Internal storage options are added by implementing institutions in their custom implementation. They should have a possibility of changing these options after the fact.

### LINKED ISSUE #198: Expand the InternalStorageService
**Description**
Currently, internal storages can only be added via a sql script. This is annoying and a lot of extra work, if the list has to be changed.

**Proposed solution**
Expand the already existing InternalStorageService to take care of storage related CRUD operations. Every operation aside from reading should be reserved for admins only.

**Additional context**
Keep in mind this related issue https://github.com/tuwien-csd/damap-backend/issues/95
This could be solved together with this one.

https://github.com/tuwien-csd/damap-frontend/issues/236 will provide frontend support so that the new service is used.

GeoffreyKarnbach:
## Description
Feature:

#### What does this PR do?
 Added active boolean to internal Storage Entities. Added CRUD for Internal Storage and Internal Storage Translations.
The validator of the backend for dmp, where the activeness status of the Internal Storages is disabled (bug, needs to be adressed).

#### Breaking changes
Rework of the InternalStorageDO, changes are reflected in the frontend as well.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-198
closes GH-95


docker/docker-compose.yaml
rekt-hard, line 1: Changes should be obsolete via a git rebase with next

src/main/java/org/damap/base/repo/InternalStorageTranslationRepo.java
```diff
@@ -39,4 +39,28 @@ public InternalStorageTranslation getInternalStorageById(Long storageId, String
...
         .firstResult();
   }
+
+  public InternalStorageTranslation getAnyTranslationByInternalStorageById(Long storageId) {
+    return find(
+            "select storage from InternalStorageTranslation storage"
+                + " where storage.internalStorageId.id = :storageId",
+            Parameters.with("storageId", storageId))
+        .firstResult();
+  }
```
rekt-hard: I could not find a usage for this

src/main/java/org/damap/base/repo/InternalStorageRepo.java
```diff
@@ -17,4 +19,42 @@ public class InternalStorageRepo implements PanacheRepository<InternalStorage> {
...
+  /**
+   * Search by parameters.
+   *
+   * @param url a {@link java.lang.String} object
+   * @param storageLocation a {@link java.lang.String} object
+   * @param backupLocation a {@link java.lang.String} object
+   * @param active a {@link java.lang.Boolean} object
+   * @return a {@link java.util.List} object
+   */
+  public List<InternalStorage> searchByParameters(
```
rekt-hard: see review comment about general search possibility

src/main/java/org/damap/base/rest/DataManagementPlanResource.java
```diff
@@ -133,7 +133,12 @@ public DmpDO updateDmp(@PathParam String id, @Valid DmpDO dmpDO) {
     if (!accessValidator.canEditDmp(dmpId, personId)) {
       throw new ForbiddenException(unauthorizedMessage(dmpId));
     }
-    return dmpService.update(dmpDO);
+    try {
+      return dmpService.update(dmpDO);
+    } catch (ClientErrorException e) {
+      throw new ClientErrorException(
+          "Error updating dmp with id: " + id + " " + e.getMessage(), e.getResponse());
+    }
```
rekt-hard: I do not think there is a need to re-throw this. If the service level throws, the resource will propagate it to the client. A client also has enough context to know on which request this happened (i.e. it knows that it tried to update a DMP and something went wrong).

src/main/java/org/damap/base/rest/dmp/mapper/DmpDOMapper.java
```diff
@@ -175,6 +175,11 @@ public DmpDO mapEntityToDO(Dmp dmp, DmpDO dmpDO) {
    * @return a {@link org.damap.base.domain.Dmp} object
    */
   public Dmp mapDOtoEntity(DmpDO dmpDO, Dmp dmp, MapperService mapperService) {
+
+    for (StorageDO storageDO : dmpDO.getStorage()) {
+      log.info(storageDO.getDatasets());
+    }
```
rekt-hard: debug remnant?

src/main/java/org/damap/base/rest/dmp/mapper/MapperService.java
```diff
@@ -67,11 +68,14 @@ public InternalStorage getInternalStorageById(Long id) {
    * @return a {@link org.damap.base.rest.storage.InternalStorageDO} object
    */
   public InternalStorageDO getInternalStorageDOById(Long id, String languageCode) {
-    InternalStorageTranslation internalStorageTranslation =
-        internalStorageTranslationRepo.getInternalStorageById(id, languageCode);
-    if (internalStorageTranslation != null)
+
+    InternalStorage internalStorage = internalStorageRepo.findById(id);
```
rekt-hard, line 72:
```suggestion
    InternalStorage internalStorage = internalStorageRepo.findById(id);
    if (internalStorage == null) {
      return null;
    }
```

Having the null check and return first allows to reduce logic and unnecessary calls later.

src/main/java/org/damap/base/rest/dmp/service/DmpConsistencyUtility.java
```diff
@@ -45,6 +50,66 @@ public void enforceDmpConsistency(DmpDO dmpDO) {
...
+
+    if (oldDmp == null) {
+      for (StorageDO storageDO : newDmp.getStorage()) {
+        InternalStorage storage = InternalStorage.findById(storageDO.getInternalStorageId());
+        if (!storage.isActive()) {
+          throw new ValidationException("Storage is not active");
+        }
+      }
+      return;
+    }
```
rekt-hard: Instead of reading one by one, maybe a search makes more sense. Easy to check if the result has the same length as the provided storages. Additional checks can be performed then as well.

src/main/java/org/damap/base/rest/dmp/service/DmpConsistencyUtility.java
```diff
@@ -45,6 +50,66 @@ public void enforceDmpConsistency(DmpDO dmpDO) {
...
+      // TODO: From here on, the entire update breaks if we access the distribution list of a
+      // dataset
+      // that is not in the current session. This is because the distribution list is lazy loaded
+      // and the session is closed. We need to find a way to load the distribution list in the
+      // current session.
+
+      /*
+      List<String> newUsedInDatasets = new ArrayList<>(List.copyOf(storageDO.getDatasets()));
+
+      for (String datasetReferenceHash : storageDO.getDatasets()) {
```
rekt-hard: Could I ask you to open an issue to switch from reference hash to actual ids? AFAIK, reference hashes are never validated and could be set/changed by a client.

src/main/java/org/damap/base/rest/dmp/service/DmpConsistencyUtility.java
```diff
@@ -45,6 +50,66 @@ public void enforceDmpConsistency(DmpDO dmpDO) {
...
+          throw new ValidationException("Storage is not active");
+        }
+      }
+      return;
+    }
+
+    List<Dataset> oldDatasets = oldDmp.getDatasetList();
+
+    for (StorageDO storageDO : newDmp.getStorage()) {
+      // For every dataset from the newDMP, we fetch the dataset from the persistent storage
```
rekt-hard: In general, an inactive storage may be used on any dataset, if the storage has been used before. We do not have to check if it is the same dataset. 

Maybe it makes sense to approach this whole storage check differently. Have a set of the old storage ids and a set of new storage ids. Subtract the old storage ids from the new ids - this leaves us with storage ids which have not been used before. Only perform the active check on the completely new storage.

For getting the old storage ids, maybe something like this could work (not sure if the assign check will actually work).
```java
    var existingStorageIds = dmp.getHostList()
        .stream()
        .filter(host -> Storage.class.isAssignableFrom(host.getClass()))
        .map(host -> ((Storage) host).getInternalStorageId().id)
        .collect(Collectors.toSet());
```

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
```diff
@@ -127,10 +132,17 @@ public DmpDO create(@Valid DmpDO dmpDO, String editedBy) {
...
+  public DmpDO update(@Valid DmpDO dmpDO) throws ClientErrorException {
     log.info("Updating DMP with id " + dmpDO.getId());
     DmpConsistencyUtility.enforceDmpConsistency(dmpDO);
     Dmp dmp = dmpRepo.findById(dmpDO.getId());
+
+    try {
+      DmpConsistencyUtility.enforceActiveStorage(dmpDO, dmp);
+    } catch (ClientErrorException e) {
+      throw e;
+    }
```
rekt-hard: No need to re-throw here

src/main/java/org/damap/base/rest/madmp/mapper/MaDmpMapper.java
```diff
@@ -481,15 +482,40 @@ public Host.SupportVersioning getSupportVersioning(Yesno versioning) {
...
+              + DEFAULT_LANGUAGE_CODE
+              + " for internal storage with id "
+              + internalStorageDO.getId());
+      for (InternalStorageTranslationDO translationDO : internalStorageDO.getTranslations()) {
+        host.setTitle(translationDO.getTitle());
+        host.setDescription(translationDO.getDescription());
+        host.setBackupFrequency(translationDO.getBackupFrequency());
+        break;
+      }
+    }
```
rekt-hard: Can the duplicated logic be reduced? First getting a translation and then set the fields.

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -26,7 +48,169 @@ public List<InternalStorageDO> getAllByLanguage(String languageCode) {
...
+    InternalStorage internalStorage =
+        InternalStorageDOMapper.mapDOToEntity(data, new InternalStorage());
+    internalStorage.persistAndFlush();
+
+    for (InternalStorageTranslationDO translationDO : data.getTranslations()) {
+      InternalStorageTranslation translation =
+          InternalStorageTranslationDOMapper.mapDOToTranslationEntityForCreation(
+              translationDO, internalStorage);
+      translation.setInternalStorageId(internalStorage);
+      translation.persistAndFlush();
```
rekt-hard: Why do we need to manually persist and flush? Does the ORM not take care of setting the ID accordingly?

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -26,7 +48,169 @@ public List<InternalStorageDO> getAllByLanguage(String languageCode) {
...
+  public InternalStorageDO read(String id, MultivaluedMap<String, String> queryParams) {
+
+    InternalStorage internalStorage = internalStorageRepo.findById(Long.parseLong(id));
+    List<InternalStorageTranslation> translations =
+        internalStorageTranslationRepo.getAllInternalStorageTranslationsByStorageId(
+            Long.parseLong(id));
+
+    if (internalStorage == null) {
+      throw new NotFoundException("No internal storage with ID " + id + " found");
+    }
```
rekt-hard:
```suggestion
    InternalStorage internalStorage = internalStorageRepo.findByIdOptional(Long.parseLong(id)).orElseThrow(NotFoundException("No internal storage with ID " + id + " found"));
    List<InternalStorageTranslation> translations =
        internalStorageTranslationRepo.getAllInternalStorageTranslationsByStorageId(
            Long.parseLong(id));

```

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -26,7 +48,169 @@ public List<InternalStorageDO> getAllByLanguage(String languageCode) {
...
+   * @param id a {@link java.lang.String} object
+   */
+  @Override
+  @Transactional
+  public void delete(String id) {
+    InternalStorage internalStorage = internalStorageRepo.findById(Long.parseLong(id));
+
+    if (internalStorage == null) {
+      throw new NotFoundException("No internal storage with ID " + id + " found");
+    }
```
rekt-hard:
```suggestion
    InternalStorage internalStorage = internalStorageRepo.findByIdOptional(Long.parseLong(id)).orElseThrow(NotFoundException("No internal storage with ID " + id + " found"));
```

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -26,7 +48,169 @@ public List<InternalStorageDO> getAllByLanguage(String languageCode) {
...
+    for (InternalStorage internalStorage : internalStorageList) {
+      List<InternalStorageTranslation> translations =
+          internalStorageTranslationRepo.getAllInternalStorageTranslationsByStorageId(
+              internalStorage.id);
+      internalStorageDOList.add(
+          InternalStorageDOMapper.mapEntityToDO(
+              internalStorage, new InternalStorageDO(), translations));
+    }
+
+    Search search = Search.fromMap(queryParams);
```
rekt-hard: See general review comment on search

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationService.java
```diff
@@ -0,0 +1,143 @@
...
+  public List<InternalStorageTranslationDO> getAllByStorageId(String storageId) {
+
+    if (!InternalStorageValidator.storageIdExists(Long.valueOf(storageId))) {
+      throw new NotFoundException("No internal storage found for id " + storageId);
+    }
+
+    List<InternalStorageTranslation> translations =
+        internalStorageTranslationRepo.getAllInternalStorageTranslationsByStorageId(
+            Long.valueOf(storageId));
+    log.info("Translations: " + translations);
```
rekt-hard: debug remnant?

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationValidator.java
```diff
@@ -0,0 +1,101 @@
...
+        || internalStorageTranslationDO.getDescription().isEmpty()) {
+      throw new ClientErrorException("Description must not be empty", Response.Status.BAD_REQUEST);
+    }
+
+    if (internalStorageTranslationDO.getBackupFrequency() == null
+        || internalStorageTranslationDO.getBackupFrequency().isEmpty()) {
+      throw new ClientErrorException(
+          "Backup frequency must not be empty", Response.Status.BAD_REQUEST);
+    }
+  }
```
rekt-hard: See general review comment on validation. Most of this can be achieved via Class Annotations on the DO Object.
Validating the existence is important as well, though it will be done in the service level when trying to read the Storage entity.

src/main/java/org/damap/base/rest/storage/InternalStorageValidator.java
```diff
@@ -0,0 +1,80 @@
...
+  public void validateSearchParameters(MultivaluedMap<String, String> queryParams) {
+    List<String> allowedParams =
+        Arrays.asList("url", "active", "storageLocation", "backupLocation");
+
+    for (String key : queryParams.keySet()) {
+      if (!allowedParams.contains(key)) {
+        throw new ClientErrorException(
+            "Invalid search parameter: " + key, Response.Status.BAD_REQUEST);
+      }
+    }
```
rekt-hard: Nice approach. See also the general review comment on search.
Throwing an exception might be too restrictive. It would suffice to simply remove this attribute from the filters.

src/main/resources/org/damap/base/db/changeLog-4.x/storage_procedures_2.sql
rekt-hard, line 1: The idea of the procedure file is that it can run every time liquibase runs the changelogs. So if we have to change a procedure, we can update the same file. Therefore, we should be able to just update one file and not have an additional changelog.

rekt-hard (commented):
Massive work - thank you for digging into this and also coming up with a first approach for the search.

**Search**
This shows how we can be very explicit about the search yet the modularity is missing. If this were to applied to another resource/service/repo, all the logic would have to be duplicated. I have created an issue detailing the general approach: https://github.com/tuwien-csd/damap-backend/issues/249

**Validation**
We can perform object validation on the service level: https://quarkus.io/guides/validation#service-method-validation. `*DO` objects used in the application are the ones for the API. Constraints can be defined on these classes and then be validated in the service.

**Exceptions**
We do not have to re-throw most exceptions on the resource level. If the service level throws, the resource will propagate it to the client. A client also has enough context to know on which request this happened (e.g. it knows that it tried to update a DMP and something went wrong).

**Common error response**
In order to easily show errors and warnings in the front end, we should come up with a common error interface. There is an extra issue for this now: https://github.com/tuwien-csd/damap-backend/issues/250

src/main/java/org/damap/base/repo/InternalStorageTranslationRepo.java
GeoffreyKarnbach: Removed

docker/docker-compose.yaml
GeoffreyKarnbach: Done

src/main/java/org/damap/base/rest/dmp/mapper/DmpDOMapper.java
GeoffreyKarnbach: Indeed, removed

src/main/java/org/damap/base/rest/dmp/mapper/MapperService.java
GeoffreyKarnbach: Done

src/main/java/org/damap/base/rest/madmp/mapper/MaDmpMapper.java
GeoffreyKarnbach: Done, we first get the list and preset the first element, then we potentially replace it.

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
GeoffreyKarnbach: Does not work, also not with "new"

> reason: no instance(s) of type variable(s) X exist so that NotFoundException conforms to Supplier<? extends X>

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
GeoffreyKarnbach: Same as above, does not work.

GeoffreyKarnbach: InternalStorage internalStorage = internalStorageRepo.findByIdOptional(Long.parseLong(id))
            .orElseThrow(() -> new NotFoundException("No internal storage with ID " + id + " found"));

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
GeoffreyKarnbach: InternalStorage internalStorage = internalStorageRepo.findByIdOptional(Long.parseLong(id))
            .orElseThrow(() -> new NotFoundException("No internal storage with ID " + id + " found"));

    List<InternalStorageTranslation> translations =
            internalStorageTranslationRepo.getAllInternalStorageTranslationsByStorageId(
                    Long.parseLong(id));

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationService.java
GeoffreyKarnbach: The log.info has been removed. The List is needed for the return

src/main/resources/org/damap/base/db/changeLog-4.x/storage_procedures_2.sql
GeoffreyKarnbach: Removed the old changeLog 4.0.3.1 and update storage_procedures.sql (necessary in that order, else active is defined in the procedurs but not in the schema).

src/main/java/org/damap/base/rest/dmp/service/DmpConsistencyUtility.java
GeoffreyKarnbach: https://github.com/tuwien-csd/damap-frontend/issues/303

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
GeoffreyKarnbach: Removed

src/main/java/org/damap/base/rest/storage/InternalStorageValidator.java
GeoffreyKarnbach: Solved and updated test case

src/main/java/org/damap/base/rest/DataManagementPlanResource.java
GeoffreyKarnbach: Can be removed indeed.

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
GeoffreyKarnbach: Can be removed indeed.

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationValidator.java
GeoffreyKarnbach: Added annotation on validator

src/main/java/org/damap/base/repo/InternalStorageRepo.java
GeoffreyKarnbach: Implemented!

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
GeoffreyKarnbach: Done!

src/main/java/org/damap/base/rest/dmp/service/DmpConsistencyUtility.java
GeoffreyKarnbach: Solved

src/main/java/org/damap/base/rest/dmp/service/DmpConsistencyUtility.java
GeoffreyKarnbach: Implemented

src/main/java/org/damap/base/repo/base/RepoSearch.java
```diff
@@ -0,0 +1,48 @@
...
+
+    for (Map.Entry<String, List<Object>> entry : queryParams.entrySet()) {
+      String key = entry.getKey();
+      List<Object> values = entry.getValue();
+      if (values.size() == 1) {
+        if (counter != 0) {
+          query.append(" AND ");
+        }
+        query.append(key).append(" = :").append(key);
+        params.put(key, values.get(0));
```
rekt-hard: Dealing with this case on its own adds quite some extra code for only omitting parenthesis. Would it be ok to remove this and just go with the general case?

With this case we would get `a=x AND b=y AND c=z` instead of `a=x AND (b=y) AND (c=z)`

src/main/java/org/damap/base/repo/base/RepoSearch.java
```diff
@@ -0,0 +1,48 @@
...
+      } else {
+        if (counter != 0) {
+          query.append(" AND ( ");
+        } else {
+          query.append(" ( ");
+        }
```
rekt-hard:
```suggestion
        if (counter != 0) {
          query.append(" AND "); // the extra white space at the end should not matter. if it does, it can also be removed as the next parenthesis will have a leading white space anyway
        }
        query.append(" ( ");
```

src/main/java/org/damap/base/repo/base/RepoSearch.java
```diff
@@ -0,0 +1,48 @@
...
+        query.append(key).append(" = :").append(key).append("0");
+        params.put(key + "0", values.get(0));
```
rekt-hard: What happens if an empty list is passed?

src/main/java/org/damap/base/repo/base/RepoSearch.java
```diff
@@ -0,0 +1,48 @@
...
+
+import io.quarkus.hibernate.orm.panache.PanacheRepository;
+import jakarta.ws.rs.core.MultivaluedMap;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public interface RepoSearch<T> extends PanacheRepository<T> {
+
+  default List<T> searchByParameters(MultivaluedMap<String, Object> queryParams) {
```
rekt-hard, line 20: How about we also pass in a Pagination object and use it for pagination as well as sort the results?

rekt-hard, line 20: Java docs and a short description of how the function constructs the query would be helpful

src/main/java/org/damap/base/rest/InternalStorageResource.java
```diff
@@ -17,18 +23,126 @@
...
+
   /**
    * getAllByLanguage.
    *
    * @param languageCode a {@link java.lang.String} object
    * @return a {@link java.util.List} object
    */
   @GET
-  @Path("/{languageCode}")
+  @Path("/languageCode/{languageCode}")
```
rekt-hard: Would it make sense to remove this and use the `search` endpoint instead?
Clients would have to request something like `/api/storages?languageCode=eng`

src/main/java/org/damap/base/rest/InternalStorageTranslationResource.java
```diff
@@ -0,0 +1,106 @@
...
+  @Override
+  @RolesAllowed("Damap Admin")
+  public InternalStorageTranslationDO create(@Valid InternalStorageTranslationDO data) {
+    log.debug("Create internal storage translation");
+    log.debug(data);
+
+    try {
+      return internalStorageTranslationService.create(data);
+    } catch (NotFoundException e) {
+      throw e;
```
rekt-hard: Why do we manually have to catch and re-throw the `NotFoundException`? Would assume this is handled automatically.

src/main/java/org/damap/base/rest/InternalStorageTranslationResource.java
```diff
@@ -0,0 +1,106 @@
...
+    } catch (ClientErrorException e) {
+      throw new ClientErrorException(
+          Response.status(Response.Status.BAD_REQUEST)
+              .entity(MESSAGE_START + e.getMessage() + "\"}")
+              .type(MediaType.APPLICATION_JSON)
+              .build());
+    }
+  }
```
rekt-hard, line 46: Why do we re-throw? Just for the front end?

src/main/java/org/damap/base/rest/InternalStorageTranslationResource.java
```diff
@@ -0,0 +1,106 @@
...
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.core.UriInfo;
+import java.util.List;
+import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.rest.base.resource.*;
+import org.damap.base.rest.storage.InternalStorageTranslationDO;
+import org.damap.base.rest.storage.InternalStorageTranslationService;
+
+@Path("/api/storages/translations")
```
rekt-hard: This path could also be a valid storage id, which might be confusing.
Usually API endpoints usually follow a schema of `/entityPlural/id/coupledEntityPlural1/id`

src/main/java/org/damap/base/rest/InternalStorageTranslationResource.java
```diff
@@ -0,0 +1,106 @@
...
+@Authenticated
+@Produces(MediaType.APPLICATION_JSON)
+@JBossLog
+public class InternalStorageTranslationResource
```
rekt-hard: Maybe it makes more sense to not have a separate resource for storage translations. Instead, combine the two resources as storage translations are tightly coupled with storage. We are not going to operate on storage translations alone but always in the context of a certain storage.

So on the storage resource, there are the common storage endpoints:
- `/api/storage`
- `/api/storage/{storageID}`

and then the storage translation endpoints:
- `/api/storage/{storageID}/translations`
- `/api/storage/{storageID}/translations/{storageTranslationID}`

src/main/java/org/damap/base/rest/base/service/ServiceSearch.java
```diff
@@ -12,4 +13,20 @@ public interface ServiceSearch<E> {
...
+    } else if (type == String.class) {
+      return value;
+    } else {
+      throw new IllegalArgumentException("Unsupported type: " + type);
+    }
+  }
+
+  default MultivaluedMap<String, Class> getEntityFields() {
+    return new MultivaluedHashMap<>();
+  }
```
rekt-hard, line 31: Fine for now.
In the long run, it might make sense to simply inject a service config, which specifies these attributes instead of having methods  on the service itself.

src/main/java/org/damap/base/rest/dmp/service/DmpConsistencyUtility.java
```diff
@@ -45,6 +57,67 @@ public void enforceDmpConsistency(DmpDO dmpDO) {
     enforceDatasetConsistency(dmpDO);
   }
 
+  public void enforceActiveStorage(
+      DmpDO newDmp, Dmp oldDmp, InternalStorageService internalStorageService)
+      throws ValidationException {
+
```
rekt-hard, line 73: Java docs would be helpful here.

src/main/java/org/damap/base/rest/dmp/service/DmpConsistencyUtility.java
```diff
@@ -45,6 +57,67 @@ public void enforceDmpConsistency(DmpDO dmpDO) {
...
+      }
+
+      for (InternalStorageDO storageDO : storages) {
+        if (!storageDO.getActive()) {
+          throw new ValidationException(STORAGE_NOT_ACTIVE);
+        }
+      }
+
+      return;
+    }
```
rekt-hard: Do we have to treat this case separately? A possibility would be to use the filter approach below (setting old storage ids to an empty set) and then retrieve the remaining storages (either one by one or all together with one search).

This could reduce logic duplication and complexity.

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
```diff
@@ -127,10 +131,13 @@ public DmpDO create(@Valid DmpDO dmpDO, String editedBy) {
    * @return a {@link org.damap.base.rest.dmp.domain.DmpDO} object
    */
   @Transactional
-  public DmpDO update(@Valid DmpDO dmpDO) {
+  public DmpDO update(@Valid DmpDO dmpDO) throws ClientErrorException {
```
rekt-hard: Is the throws clause necessary? Might be handled by the framework automatically

src/main/java/org/damap/base/rest/storage/InternalStorageDO.java
```diff
@@ -10,12 +13,18 @@
...
+  @NotBlank(message = "name cannot be blank")
+  @NotNull(message = "name cannot be null") private String url;
+
+  @NotBlank(message = "storage location cannot be blank")
+  @NotNull(message = "storage location cannot be null") private String storageLocation;
+
+  @NotBlank(message = "backup location cannot be blank")
+  @NotNull(message = "backup location cannot be null") private String backupLocation;
+
+  @NotNull(message = "backup frequency cannot be null") private Boolean active;
```
rekt-hard: Explicit messages might not be necessary, if they do not add context.
Here are the default messages for the constraints: https://github.com/hibernate/hibernate-validator/blob/main/engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties

Later on when creating the response, the property is mentioned (e.g. `{"path": "storageLocation", "message": "can not be null"}`.

If we do keep them, please ensure they are correct. Right now, there are some copy/paste issues.

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -26,7 +56,183 @@ public List<InternalStorageDO> getAllByLanguage(String languageCode) {
             storageTranslation ->
                 internalStorageDOList.add(
                     InternalStorageDOMapper.mapEntityToDO(
-                        storageTranslation, new InternalStorageDO())));
+                        storageTranslation.getInternalStorageId(),
```
rekt-hard: shelve: Not your fault. This should be `.getInternalStorage` instead. Had to manually look into the code to understand that this will retrieve the internal storage and not only the id :smiling_face_with_tear:

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -26,7 +56,183 @@ public List<InternalStorageDO> getAllByLanguage(String languageCode) {
...
+    for (InternalStorageTranslationDO translationDO : data.getTranslations()) {
+      translationDO.setStorageId(internalStorage.id);
+      InternalStorageTranslationValidator.validateForCreation(
+          translationDO, internalStorageRepo, internalStorageTranslationRepo);
+      InternalStorageTranslation translation =
+          InternalStorageTranslationDOMapper.mapDOToTranslationEntityForCreation(
+              translationDO, internalStorage);
+      translation.setInternalStorageId(internalStorage);
+      translation.persistAndFlush();
+    }
```
rekt-hard: This method runs inside a transaction. If anything goes wrong, the transaction should be rolled back automatically.

We should not have to manually persist and flush when we:
- create a storage
- create translations
As we might encounter an issue later on. Then we are left with a weird database state and the client can not simply retry the request, as some entities might already have been flushed.

Not sure about this: Would `.persist()` be enough without flushing? I think this would just try to register it on the transaction and set the id field without actually flushing it to the database, which the transaction should do at the end.

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -26,7 +56,183 @@ public List<InternalStorageDO> getAllByLanguage(String languageCode) {
...
+   * update a specific internal storage option.
+   *
+   * @param id a {@link java.lang.String} object
+   * @param data a {@link org.damap.base.rest.storage.InternalStorageDO} object
+   * @return a {@link org.damap.base.rest.storage.InternalStorageDO} object, updated
+   *     <p>Note: The internalStorageTranslation has to be updated through a separate endpoint
+   */
+  @Override
+  @Transactional
+  public InternalStorageDO update(String id, InternalStorageDO data) throws NotFoundException {
```
rekt-hard: Is the throws exception necessary?

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -26,7 +56,183 @@ public List<InternalStorageDO> getAllByLanguage(String languageCode) {
...
+            .orElseThrow(
+                () -> new NotFoundException("No internal storage with ID " + id + " found"));
+
+    List<Storage> storages = storageRepo.findByInternalStorageId(internalStorage);
+
+    if (!storages.isEmpty()) {
+      throw new ClientErrorException(
+          "Internal storage with ID " + id + " is still in use by " + storages.size() + " storages",
+          Response.Status.CONFLICT);
+    }
```
rekt-hard, line 138: Are we using internal storage somewhere else?

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -26,7 +56,183 @@ public List<InternalStorageDO> getAllByLanguage(String languageCode) {
...
+          internalStorageTranslationRepo.getAllInternalStorageTranslationsByStorageId(
+              internalStorage.id);
+      internalStorageDOList.add(
+          InternalStorageDOMapper.mapEntityToDO(
+              internalStorage, new InternalStorageDO(), translations));
+    }
+
+    Search search = Search.fromMap(queryParams);
+
+    return ResultList.fromItemsAndSearch(internalStorageDOList, search);
```
rekt-hard: Fine for now.
In the long run, we can think about a default implementation for a service which handles this general logic:
- convert values
- perform search
- create result list

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -26,7 +56,183 @@ public List<InternalStorageDO> getAllByLanguage(String languageCode) {
...
+  }
+
+  @Transactional
+  public InternalStorageDO getInternalStorageById(long internalStorageId) {
+    return InternalStorageDOMapper.mapEntityToDO(
+        internalStorageRepo.findById(internalStorageId),
+        new InternalStorageDO(),
+        internalStorageTranslationRepo.getAllInternalStorageTranslationsByStorageId(
+            internalStorageId));
+  }
```
rekt-hard: Looks like the same as the `read` method.

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationDO.java
```diff
@@ -0,0 +1,27 @@
...
+  @NotNull(message = "language code cannot be null") private String languageCode;
+
+  @NotBlank(message = "title cannot be blank")
+  @NotNull(message = "title cannot be null") private String title;
+
+  @NotBlank(message = "description cannot be blank")
+  @NotNull(message = "description cannot be null") private String description;
+
+  @NotBlank(message = "backup frequency cannot be blank")
+  @NotNull(message = "backup frequency cannot be null") private String backupFrequency;
```
rekt-hard: Maybe the default messages suffice.

src/main/java/org/damap/base/rest/InternalStorageResource.java
```diff
@@ -17,18 +23,126 @@
...
+  /**
+   * search for internal storage options.
+   *
+   * @param uriInfo a {@link jakarta.ws.rs.core.UriInfo} object, representing the query parameters
+   * @return a ResultList of {@link org.damap.base.rest.storage.InternalStorageDO} objects, that
+   *     match the criteria
+   */
+  @Override
+  @GET
+  @Path("/search")
```
rekt-hard: endpoint should just be `/storages`

src/main/java/org/damap/base/rest/storage/InternalStorageDO.java
```diff
@@ -10,12 +13,18 @@
...
-  private String storageLocation;
-  private String backupLocation;
+
+  @NotBlank(message = "name cannot be blank")
+  @NotNull(message = "name cannot be null") private String url;
+
+  @NotBlank(message = "storage location cannot be blank")
+  @NotNull(message = "storage location cannot be null") private String storageLocation;
+
+  @NotBlank(message = "backup location cannot be blank")
```
rekt-hard: I think this can be empty

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationDO.java
```diff
@@ -0,0 +1,27 @@
...
+  private long id;
+  private long storageId;
+
+  @NotBlank(message = "language code cannot be blank")
+  @NotNull(message = "language code cannot be null") private String languageCode;
+
+  @NotBlank(message = "title cannot be blank")
+  @NotNull(message = "title cannot be null") private String title;
+
+  @NotBlank(message = "description cannot be blank")
```
rekt-hard: Can be empty

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationDO.java
```diff
@@ -0,0 +1,27 @@
...
+  @NotNull(message = "description cannot be null") private String description;
+
+  @NotBlank(message = "backup frequency cannot be blank")
```
rekt-hard: Can be empty

rekt-hard (commented):
Massive work - thank you!

### Exception Handling
We have a separate issue for this so will try to keep it short.
What would be desired is to remove the exception handling from the resource level. There should be a possibility to register a handler for certain exception types, which then takes care of generating an appropriate response. If an exception is raised on the service level, this should get propagated to the resource level and the application wide handlers kick in.

### Storage Resource
Resources for storage and storage translations can be merged, as they are tightly coupled. See comments for more details.

### Entity Deletion
We should think about if we want to soft delete by default. This way, accidental deletions can be restored easily. For removing an entity completely, a purge method could be introduced.

src/main/java/org/damap/base/rest/InternalStorageTranslationResource.java
GeoffreyKarnbach: Yes, should we keep it?

src/main/java/org/damap/base/rest/InternalStorageTranslationResource.java
GeoffreyKarnbach: Makes sense, I thought about it as well, but due to our Resource Interfaces (ResourceCreate<T>) this would require the same interface to be implemented for different types.

src/main/java/org/damap/base/rest/storage/InternalStorageDO.java
GeoffreyKarnbach: They allow for a cleaner error message, when remaping them. Without the message:

```
{
  "violations": [
    {
      "path": "update.data.backupFrequency",
      "invalidValue": "",
      "message": "darf nicht leer sein"
    },
    {
      "path": "update.data.description",
      "invalidValue": "",
      "message": "darf nicht leer sein"
    }
  ],
  "exception": "ConstraintViolationException"
}
```


With the message:

```
{
  "violations": [
    {
      "path": "update.data.description",
      "invalidValue": "",
      "message": "description cannot be blank"
    },
    {
      "path": "update.data.backupFrequency",
      "invalidValue": "",
      "message": "backup frequency cannot be blank"
    }
  ],
  "exception": "ConstraintViolationException"
}
```

(The example for backupFrequency and description is only to showcase, validation on them has been removed as requested in the frontend).

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
GeoffreyKarnbach: Yes, inside of DMPs, if the DMP contains the given internal storage, it should not be able to be deleted but only disabled i guess?

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationDO.java
GeoffreyKarnbach: As for above, those messages allow better messages when mapped by the "ConstraintViolationExceptionMapper", i would keep it like that. The constraints for the description and frequency have been removed, according to the frontend conversation.

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationDO.java
GeoffreyKarnbach: I suppose you refer to the description, has been changed

src/main/java/org/damap/base/rest/storage/InternalStorageDO.java
GeoffreyKarnbach: For the backup location? Can be empty not

src/main/java/org/damap/base/rest/InternalStorageResource.java
GeoffreyKarnbach: Removed this and replaced the call in the frontend service (that actually never gets used anyways).

src/main/java/org/damap/base/rest/InternalStorageTranslationResource.java
GeoffreyKarnbach: Changed

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
GeoffreyKarnbach: Removed and replaced by read call

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
GeoffreyKarnbach: Valid point, works perfectly fine only with persist. Thanks!

src/main/java/org/damap/base/repo/base/RepoSearch.java
```diff
@@ -0,0 +1,55 @@
...
+      if (counter != 0) {
+        query.append(" AND ");
+      }
+
+      query.append(" ( ");
+
+      query.append(key).append(" = :").append(key).append("0");
+      params.put(key + "0", values.get(0));
+      for (int i = 1; i < values.size(); i++) {
+        query.append(" OR ").append(key).append(" = :").append(key).append(i);
```
rekt-hard: minor: for readability reasons, would it make sense to use a formatted string here?
```java
query.append(" OR %s = :%s".formatted(key, key + i));
```

src/main/java/org/damap/base/rest/InternalStorageTranslationResource.java
```diff
@@ -0,0 +1,104 @@
...
+      throw new ClientErrorException(
+          Response.status(Response.Status.BAD_REQUEST)
+              .entity(MESSAGE_START + e.getMessage() + "\"}")
+              .type(MediaType.APPLICATION_JSON)
+              .build());
+    }
+  }
+
+  @GET
+  @Path("/all")
```
rekt-hard: major: To be consistent with other endpoints, I think we can omit the `/all` prefix for retrieving all entries or searching.
Also `/all` could be misunderstood as an identifier.

src/main/java/org/damap/base/rest/dmp/service/DmpConsistencyUtility.java
```diff
@@ -45,6 +53,60 @@ public void enforceDmpConsistency(DmpDO dmpDO) {
...
+            oldDmpStorageIds.add(internalStorage.id);
+          }
+        }
+      }
+    }
+
+    // Difference between new and old DMP storage IDs
+    newDmpStorageIds.removeAll(oldDmpStorageIds);
+
+    if (!newDmpStorageIds.isEmpty()) {
```
rekt-hard: minor: conditional can be removed

src/main/java/org/damap/base/rest/dmp/service/DmpConsistencyUtility.java
```diff
@@ -45,6 +53,60 @@ public void enforceDmpConsistency(DmpDO dmpDO) {
...
+      for (Long id : newDmpStorageIds) {
+        if (id == null) {
+          continue;
+        }
```
rekt-hard: question: why can this be null?
can we filter null values beforehand? e.g. `newDmpStorageIds.removeIf(s -> s == null);`

src/main/java/org/damap/base/rest/storage/InternalStorageService.java
```diff
@@ -2,31 +2,219 @@
...
-   * @param languageCode a {@link java.lang.String} object
-   * @return a {@link java.util.List} object
+   * @param id a {@link java.lang.String} object
+   * @return
    */
-  public List<InternalStorageDO> getAllByLanguage(String languageCode) {
+  @Override
+  public InternalStorageDO read(String id) {
+    return this.read(id, null);
+  }
```
rekt-hard: minor: do we have to override this? The default is to call read with an empty map.

rekt-hard (commented):
Generally looking good. Some comments that should be rather quick to address.


One major issue:
I am getting an error on the admin page when fetching the internal storage.

```
HTTP Request to /api/storages failed, error id: e7ed2b7f-73ad-4422-8192-d2b31b219114-12: org.hibernate.PropertyAccessException: Null value was assigned to a property [class org.damap.base.domain.InternalStorage.active] of primitive type: 'org.damap.base.domain.InternalStorage.active' (setter)
	at org.hibernate.property.access.spi.SetterFieldImpl.set(SetterFieldImpl.java:63)
	at org.hibernate.property.access.spi.EnhancedSetterImpl.set(EnhancedSetterImpl.java:42)
	at org.hibernate.persister.entity.AbstractEntityPersister.setPropertyValues(AbstractEntityPersister.java:4459)
	at org.hibernate.sql.results.graph.entity.AbstractEntityInitializer.initializeEntityInstance(AbstractEntityInitializer.java:809)
	at org.hibernate.sql.results.graph.entity.AbstractEntityInitializer.initializeEntity(AbstractEntityInitializer.java:769)
	at org.hibernate.sql.results.graph.entity.AbstractEntityInitializer.initializeInstance(AbstractEntityInitializer.java:761)
	at org.hibernate.sql.results.internal.InitializersList.initializeInstance(InitializersList.java:73)
	at org.hibernate.sql.results.internal.StandardRowReader.coordinateInitializers(StandardRowReader.java:113)
	at org.hibernate.sql.results.internal.StandardRowReader.readRow(StandardRowReader.java:87)
	at org.hibernate.sql.results.spi.ListResultsConsumer.consume(ListResultsConsumer.java:205)
	at org.hibernate.sql.results.spi.ListResultsConsumer.consume(ListResultsConsumer.java:33)
	at org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl.doExecuteQuery(JdbcSelectExecutorStandardImpl.java:211)
	at org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl.executeQuery(JdbcSelectExecutorStandardImpl.java:83)
	at org.hibernate.sql.exec.spi.JdbcSelectExecutor.list(JdbcSelectExecutor.java:76)
	at org.hibernate.sql.exec.spi.JdbcSelectExecutor.list(JdbcSelectExecutor.java:65)
	at org.hibernate.query.sqm.internal.ConcreteSqmSelectQueryPlan.lambda$new$2(ConcreteSqmSelectQueryPlan.java:139)
	at org.hibernate.query.sqm.internal.ConcreteSqmSelectQueryPlan.withCacheableSqmInterpretation(ConcreteSqmSelectQueryPlan.java:382)
	at org.hibernate.query.sqm.internal.ConcreteSqmSelectQueryPlan.performList(ConcreteSqmSelectQueryPlan.java:302)
	at org.hibernate.query.sqm.internal.QuerySqmImpl.doList(QuerySqmImpl.java:526)
	at org.hibernate.query.spi.AbstractSelectionQuery.list(AbstractSelectionQuery.java:423)
	at org.hibernate.query.Query.getResultList(Query.java:120)
	at io.quarkus.hibernate.orm.panache.common.runtime.CommonPanacheQueryImpl.list(CommonPanacheQueryImpl.java:312)
	at io.quarkus.hibernate.orm.panache.runtime.PanacheQueryImpl.list(PanacheQueryImpl.java:149)
	at io.quarkus.hibernate.orm.panache.runtime.JpaOperations.list(JpaOperations.java:24)
	at io.quarkus.hibernate.orm.panache.runtime.JpaOperations.list(JpaOperations.java:10)
	at io.quarkus.hibernate.orm.panache.common.runtime.AbstractJpaOperations.list(AbstractJpaOperations.java:253)
	at org.damap.base.repo.InternalStorageRepo.list(InternalStorageRepo.java)
	at org.damap.base.repo.base.RepoSearch.searchByParameters(RepoSearch.java:56)
	at org.damap.base.repo.InternalStorageRepo_ClientProxy.searchByParameters(Unknown Source)
	at org.damap.base.rest.storage.InternalStorageService.search(InternalStorageService.java:180)
	at org.damap.base.rest.storage.InternalStorageService_ClientProxy.search(Unknown Source)
	at org.damap.base.rest.InternalStorageResource.search(InternalStorageResource.java:132)
	at org.damap.base.rest.InternalStorageResource_Subclass.search$$superforward(Unknown Source)
	at org.damap.base.rest.InternalStorageResource_Subclass$$function$$11.apply(Unknown Source)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:73)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext$NextAroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:97)
	at io.quarkus.security.runtime.interceptor.SecurityHandler.handle(SecurityHandler.java:27)
	at io.quarkus.security.runtime.interceptor.AuthenticatedInterceptor.intercept(AuthenticatedInterceptor.java:29)
	at io.quarkus.security.runtime.interceptor.AuthenticatedInterceptor_Bean.intercept(Unknown Source)
	at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:70)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:62)
	at io.quarkus.resteasy.reactive.server.runtime.StandardSecurityCheckInterceptor.intercept(StandardSecurityCheckInterceptor.java:44)
	at io.quarkus.resteasy.reactive.server.runtime.StandardSecurityCheckInterceptor_AuthenticatedInterceptor_Bean.intercept(Unknown Source)
	at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:30)
	at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:27)
	at org.damap.base.rest.InternalStorageResource_Subclass.search(Unknown Source)
	at org.damap.base.rest.InternalStorageResource$quarkusrestinvoker$search_959072baf5d22d5060f26385d4774d1f0c6ee23b.invoke(Unknown Source)
	at org.jboss.resteasy.reactive.server.handlers.InvocationHandler.handle(InvocationHandler.java:29)
	at io.quarkus.resteasy.reactive.server.runtime.QuarkusResteasyReactiveRequestContext.invokeHandler(QuarkusResteasyReactiveRequestContext.java:141)
	at org.jboss.resteasy.reactive.common.core.AbstractResteasyReactiveContext.run(AbstractResteasyReactiveContext.java:147)
	at io.quarkus.vertx.core.runtime.VertxCoreRecorder$14.runWith(VertxCoreRecorder.java:599)
	at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2516)
	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2495)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1521)
	at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:11)
	at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:11)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.lang.IllegalArgumentException: Can not set boolean field org.damap.base.domain.InternalStorage.active to null value
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:167)
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:171)
	at java.base/jdk.internal.reflect.UnsafeBooleanFieldAccessorImpl.set(UnsafeBooleanFieldAccessorImpl.java:80)
	at java.base/java.lang.reflect.Field.set(Field.java:799)
	at org.hibernate.property.access.spi.SetterFieldImpl.set(SetterFieldImpl.java:57)

```

src/main/java/org/damap/base/repo/base/RepoSearch.java
GeoffreyKarnbach: Done

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationValidator.java
```diff
@@ -0,0 +1,79 @@
...
+    InternalStorageTranslation internalStorageTranslation =
+        internalStorageTranslationRepo.findById(Long.parseLong(id));
+
+    if (internalStorageTranslation == null) {
+      throw new NotFoundException("No internal storage translation with ID " + id + " found");
+    }
+
+    InternalStorageTranslationValidator.validationCommon(
+        internalStorageTranslationDO, internalStorageRepo);
+  }
```
rekt-hard: major: missing validation to check if another translation with the same language key already exists.

src/main/java/org/damap/base/rest/storage/InternalStorageTranslationValidator.java
```diff
@@ -0,0 +1,90 @@
...
+      throw new NotFoundException(
+          "No internal storage found for id " + internalStorageTranslationDO.getStorageId());
+    }
+
+    // Check if language code is 'deu' or 'eng'
+    if (!internalStorageTranslationDO.getLanguageCode().equals("deu")
+        && !internalStorageTranslationDO.getLanguageCode().equals("eng")) {
+      throw new ClientErrorException(
+          "Language code must be 'deu' or 'eng'", Response.Status.BAD_REQUEST);
+    }
```
rekt-hard, line 88: minor/shelve: Might make sense to fetch available languages from the database and compare with them. Probably not really relevant right now.

rekt-hard (commented):
major: ouch. we have a hard dependency on an English translations being present when exporting a document. So when providing only a German translation, the export fails. Could I ask you to update the relevant code and try to fetch the English Translation. If this does not exist, just use the first one available?
https://github.com/tuwien-csd/damap-backend/blob/ec1b0a9b9739b3ccf1b5d655e835a7a2ef4a6dc9/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java#L409-L411

```
java.lang.NullPointerException: Cannot invoke "org.damap.base.domain.InternalStorageTranslation.getDescription()" because the return value of "org.damap.base.repo.InternalStorageTranslationRepo.getInternalStorageById(java.lang.Long, String)" is null
	at org.damap.base.conversion.AbstractTemplateExportScienceEuropeComponents.storageInformation(AbstractTemplateExportScienceEuropeComponents.java:411)
	at org.damap.base.conversion.AbstractTemplateExportScienceEuropeComponents.loadScienceEuropeContent(AbstractTemplateExportScienceEuropeComponents.java:38)
	at org.damap.base.conversion.ExportFWFTemplate.exportTemplate(ExportFWFTemplate.java:42)
	at org.damap.base.conversion.ExportFWFTemplate_ClientProxy.exportTemplate(Unknown Source)
	at org.damap.base.conversion.ExportTemplateBroker.exportTemplateByType(ExportTemplateBroker.java:69)
	at org.damap.base.conversion.ExportTemplateBroker_ClientProxy.exportTemplateByType(Unknown Source)
	at org.damap.base.rest.DmpDocumentResource.exportTemplate(DmpDocumentResource.java:56)

```

rekt-hard (commented):
Thank you for incorporating all the feedback :rocket: 
Could I ask you to rebase and update the change log file, so it matches the next release version (4.3.0)?

rekt-hard (approved):
Tested locally with clean database as well as existing database. Good job :clap:

PR #258, opened 2024-09-23
Template: Move reused dataset description bellow table for HorizonEurope and ScienceEurope

### LINKED ISSUE #243: Export: Dataset description column is too slim
**Changes**
Right now, the description column in the produced and reused dataset column is too slim to fit a proper description. This is especially true for ScienceEurope and HorizonEurope. Due to different formating, FWF does not have this problem. This description column stretches over 2 pages:
image

Sadly, i dont think there is an efficient way to keep the description column in these tables. Therefore I suggest that the descriptions are listed below the tables, to save space for the other columns. Something similar is already being done with repository descriptions, so maybe you an get some inspirations there. Any other solution is also fine, if you have an idea.

GeoffreyKarnbach:
## Description

Template

#### What does this PR do?

The description of reused datasets got moved from inside of the table to bellow it for the HorizonEurope and ScienceEurope template, as the table column for the description was too slim and sometimes was several pages long.

#### Dependencies

Template changes:
[horizonEuropeTemplate.docx](https://github.com/user-attachments/files/17093852/horizonEuropeTemplate.docx)
[scienceEuropeTemplate.docx](https://github.com/user-attachments/files/17093853/scienceEuropeTemplate.docx)

### Checks

- [x] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-243


rekt-hard:
Could this be done for new datasets as well? The same issue will occur eventually

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -1017,7 +1030,7 @@ public void composeTableReusedDatasets(XWPFDocument document, XWPFTable xwpfTabl
           docVar.add("no");
         }
 
-        docVar.add(Optional.ofNullable(dataset.getDescription()).orElse(""));
```
rekt-hard, line 1020: Can this be removed?

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -319,6 +319,19 @@ public void datasetsInformation() {
     if (dmp.getTargetAudience() != null)
       addReplacement(replacements, "[targetaudience]", dmp.getTargetAudience());
     else addReplacement(replacements, "[targetaudience]", "");
+
+    StringBuilder reusedDescription = new StringBuilder();
+    List<Dataset> reusedDatasets = getReusedDatasets();
+    if (!reusedDatasets.isEmpty()) {
```
rekt-hard: The conditional check can be removed, as the for loop will have no effect if the datasets are empty

ValentinFutterer (commented):
It works fine, but i noticed that you missed the new dataset column description - could you do this table as well?

Also, in the SE template, there is this:
image
 
Which leads to this when exported:
image

GeoffreyKarnbach:
Done for produced datasets as well, new change documents:
[horizonEuropeTemplate.docx](https://github.com/user-attachments/files/17114079/horizonEuropeTemplate.docx)
[scienceEuropeTemplate.docx](https://github.com/user-attachments/files/17114080/scienceEuropeTemplate.docx)

ValentinFutterer:
If a dataset has no description, null is written in the export document, pls fix then i will merge.

image

ValentinFutterer (approved):
Works 👍

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -319,6 +319,32 @@ public void datasetsInformation() {
...
+
+    StringBuilder newDescription = new StringBuilder();
+    List<Dataset> newDatasets = getNewDatasets();
+    for (int i = 0; i < newDatasets.size(); i++) {
+      Dataset dataset = newDatasets.get(i);
+      if (dataset.getDescription() != null && !dataset.getDescription().isEmpty()) {
+        newDescription.append("Description for \"");
+        newDescription.append(dataset.getTitle()).append("\": ;");
+        newDescription.append(dataset.getDescription()).append(";");
+      }
```
rekt-hard, line 344: Sooner or later, someone will probably be like: `Why is there no description exported for dataset x/y` :smile: 

Might make sense to add an empty string, if the description is null or empty.

PR #260, opened 2024-09-30
Preview: Add preview of exported PDF in the frontend using gotenberg for DOCX to PDF conversion

### LINKED ISSUE #252: Preselection of the template, Real-time view
In many cases, the researchers already know which template they need. It would probably make sense to adapt the interface or query to these templates. Alternatively, a real-time view that exists in parallel to the query would be useful to see how it would look in the exported document (e.g. online latex editor Overleaf).

### LINKED ISSUE #259: Live DMP Export Preview
**Description**

When working on a DMP (either during development to try out template changes, or as a user), to check out the current export state, a Word Document has to be exported everytime.

In order to enhance productivity, a Live Preview of the DMP Export would be very useful.

**Proposed solution**

There are two approaches, that may both be explored for displaying the preview in the Frontend: 
- A split screen view (such as in overleaf) 
image
- An overlay approach (triggered through a button press to open a popup with the preview) 
image

Both approaches have advantages and disadvantages:
- The split screen view enables the user to always see the state of the preview (updated for instance when saving the DMP after making changes), but the usability on small screens might be impacted.
- The overlay allows better usability on smaller screens (especially for laptops) but the user has to press an additional button to be able to preview the current state.

In addition to the location in the frontend, different technical challenges regarding the rendering and backend exist:
- In the frontend, a DOCX to HTML renderer can be used to display (https://github.com/VolodymyrBaydalka/docxjs), no data has to be exposed to any third party services and the file can be kept private (behind an authentication wall), as it will only be used by an authenticated user within the frontend. Main issue with that approach is, that the rendering is not perfect and has some small flaws
image
- Alternatively, a live preview iframe vom Microsoft can be used. This renders the Document in a perfect manner, but leads to a public (see the next paragraph) available word document and it exposes it to Microsoft Services. Furthermore, this option is only available in a deployed environment, where the backend has a static URL (not localhost) that can be resolved my Microsoft services.
image

The approach to offer in localhost and in a deployed environment the live preview feature would be:
- Use the JS library to render the DOCX to HTML in the dev environment (not perfect, but probably good enough to check for quick updates made in the DMP Form)
- Use the Microsoft IFrame in deployment, ensuring a reliable rendering (and possible as the backend is hosted).

As mentioned before, the files need to be "publicly" available for the Microsoft Iframe to be able to access them. As the IFrame may also take API requests as input (as long as a Word Document is returned), following approach would lead to better security:
- Persist the generated preview Word Documents on the Server with a UUID as name (32 characters for instance)
- Add a OTP / Password (64 characters) that is passed as query param to the Backend call.
- Ensure that not used export files are deleted as soon as possible (keep track of the current filename in a DB entity, remove the file before generating a new export file for a given project).

Through this approach, the "public" files, are protected by a 32 characters UUID name (that would have to be guessed) and a 64 characters password within a "short" timeframe, making it effectively not feasible for an attacker (same would apply to any other endpoint if a JWT is leaked, the password has to be treated as an authentication/access token).

This approach would furthermore allow users to generate exported documents, that can be shared using an URL and the password as query param to any non authenticated user.

An approach to persisting those temporary files would be to the "temp" environment of the operating system (%TEMP% for windows, /var/tmp or /tmp for Unix).

**Alternative solution**
Generate a PDF on Server Side from the Word Document (no feasible approach was found, possible with libraries, but requires Microsoft Word to be installed in the Server Environment => not easily possible in Linux). This approach would make the entire approach easier, as rendering a PDF in the Frontend is trivial.

GeoffreyKarnbach:
## Description

PDF File generation based on Docx file export using the Gotenberg Service (new Docker Service).

#### What does this PR do?
Add new endpoint, include Gotenberg into our docker-compose file.

#### Dependencies
https://github.com/tuwien-csd/damap-frontend/pull/320

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-252


rekt-hard (commented):
Awesome feature - thank you! :rocket: 

About the API endpoints: We do not have a dedicated document/file entity and this resource is about DMPs specifically, we may want to align the endpoints as well as the functionality a bit.

Export and preview essentially provide the same output in the back end: DMP as a document. While right now, the export is a DOCX and the preview is a PDF, we will most probably allow to "export" a PDF as well. The only difference for the back end should be the value of the `Content-Disposition header`. When downloading, this shall be `attachment`. For preview, this should be `inline`.

How the exported document will be handled is up to the client.
In the front end, when clicking on
- `Export`:  a download dialog for the user should be shown
- `Preview`: the file should be rendered in the browser to preview

With this in mind, we can rearrange the endpoints a bit.

Base API: `/api/document`
- DMP specific path: `/{dmpId}`
  - `GET` `/export` (export/preview)
    - Query Param: template type
    - Query Param: download (true by default)
    - Query Param: file type (.docx by default)
  - `GET` `/template_type` (funder specific export template)

The current export endpoint would then be deprecated. In order to not have to release a major version, we could
- refactor its logic to use the new service method
- add a deprecated annotation
- remove it completely at a later point (i.e. when a new major version is released)


Examples requests:
Preview DMP with FWF template as docx: `/api/document/1/export?template=FWF`
Download/Export DMP with SCIENCE_EUROPE as pdf: `/api/document/1/export?template=SCIENCE_EUROPE&file_type=pdf&download=1`
We may even move the maDMP export over here, as it follows the same logic to some degree (template = maDMP, file_type = json). Can be shelved for now and is up for a future discussion.

For the service level, the steps might be:
- get the DMP
- export the DMP with the specified template
- convert DMP to specified file type

We could go fancy here and introduce some kind of resolvers for the template type and file type. They would know which exact steps they have to perform to export a document and how to convert into a certain file format. This could allow institutions to more easily specify additional export formats and template types. Can be shelved for now and is up for future discussion


To summarize:
- Align endpoints based on the entity and then a certain action (i.e. `api/document/{dmpId}/export`
- Align service level implementation so it can be reused for export/preview
- Allow setting query params:
  - template type (already there)
  - file type
  - download

src/main/java/org/damap/base/rest/DmpDocumentResource.java
```diff
@@ -41,34 +43,74 @@ public class DmpDocumentResource {
    */
   @GET
   @Path("/{dmpId}")
+  @Deprecated(since = "4.3.0")
```
rekt-hard: To provide more info for other developers (at institutions mostly), if they are using this endpoint. Maybe we can also add an info about removal in the java docs, as it should probably be removed with v5
```suggestion
  @Deprecated(since = "4.3.0", forRemoval = true)
```

src/main/java/org/damap/base/rest/DmpDocumentResource.java
```diff
@@ -41,34 +43,74 @@ public class DmpDocumentResource {
...
+    return this.export(dmpId, template, true, "docx");
+  }
+
+  @GET
+  @Path("/{dmpId}/export")
+  @Produces(MediaType.APPLICATION_OCTET_STREAM)
+  public Response export(
+      @PathParam("dmpId") long dmpId,
+      @QueryParam("template") ETemplateType template,
+      @QueryParam("download") Boolean download,
```
rekt-hard: By specifying a default value, we may be able to get rid of the null check later.
```suggestion
      @QueryParam("download") @DefaultValue("true") boolean download,
```

src/main/java/org/damap/base/rest/DmpDocumentResource.java
```diff
@@ -41,34 +43,74 @@ public class DmpDocumentResource {
...
+      @QueryParam("filetype") String filetype) {
```
rekt-hard: minor/shelve: Changing this to an enum allows to shift the value check to the library.
Would also have to ensure that institutions are able to override this somehow, so they can provide their own file types. Can be shelved for now though.

rekt-hard: Might make sense to set a default value here as well

src/main/java/org/damap/base/rest/document/service/DocumentService.java
```diff
@@ -0,0 +1,95 @@
...
+      return getWordDocumentOf(document);
+    }
+
+    throw new WebApplicationException("Invalid file type: " + filetype);
+  }
+
+  private StreamingOutput getPdfOf(XWPFDocument xwpfDocument) {
+    File tempFile;
+    try {
+      tempFile = File.createTempFile("dmp", ".docx");
```
rekt-hard: major: in order to keep possible file clashes and file overrides to a minimum, adding the dmp id and current timestamp might be a good idea.

src/main/java/org/damap/base/rest/DmpDocumentResource.java
```diff
@@ -41,34 +43,74 @@ public class DmpDocumentResource {
...
+
+    if (download == null) {
+      download = true;
+    }
+
+    if (filetype == null || (!filetype.equals("pdf") && !filetype.equals("docx"))) {
+      filetype = "docx";
+    }
```
rekt-hard: major: this should be handled on the service level as there we should know how to deal with different values.

rekt-hard (commented):
Thank you for refactoring. Just some comments on smaller issues - then it should be good to go :)

src/main/java/org/damap/base/rest/DmpDocumentResource.java
GeoffreyKarnbach: Done

src/main/java/org/damap/base/rest/DmpDocumentResource.java
GeoffreyKarnbach: Done

src/main/java/org/damap/base/rest/DmpDocumentResource.java
GeoffreyKarnbach: Set to docx default

src/main/java/org/damap/base/rest/document/service/DocumentService.java
GeoffreyKarnbach: Done

src/main/java/org/damap/base/rest/DmpDocumentResource.java
GeoffreyKarnbach: Done

rekt-hard (approved):
Good job! LGTM

PR #268, opened 2024-10-09
Export: correct identifier types in a comparison

### LINKED ISSUE #266: Export: IdentifierTypes are not compared correctly
**Describe the bug**
In AbstractTemplateExportScienceEuropeComponent , this line 130 `if (personIdentifier.getIdentifierType().toString().equals("orcid")) {` causes problems. This comparison doesnt work, since the Enum is ORCID and not orcid. The same happens for other identifiers, like ROR. This leads to the identifiers missing in the contributor section of the export.
image

To fix this, do the comparison using the actual enums.

**Is this a regression?**
Yes.

**Expected behavior**
Comparisons should work like intended, see screenshot above.

**Environment**
 - Damap version [e.g. 4.1.0]

andresTabiTuwien:
## Description

Fix the problems in the comparison using the enums for them, ORCID and ROR

https://github.com/tuwien-csd/damap-backend/issues/266

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-266

### DIFF
--- a/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -127,7 +127,7 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
 
     if (personIdentifier != null) {
       String contactIdentifierId = personIdentifier.getIdentifier();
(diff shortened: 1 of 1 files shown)


rekt-hard (approved):
good catch, thank you!

PR #271, opened 2024-10-14
🔗 licenses: bring URLs up to date

SotosTsepe:
* fix dead links
* use opensource and creativecommons as main urls with spdx as fallback
* keeps URLs compatible with InvenioRDM v12+ releases


ValentinFutterer (approved):
Thanks for the help, url's match with tthe ones in the backend.

PR #277, opened 2024-11-04
Version: Update DmpListItemDO and add versionCount and latestVersionName for any given DMP

### LINKED ISSUE #346: Make DMP version info more accessible to users
**Description**
Accessing the DMP versions requires clicking the elipses button and selecting "View versions".  The process is not intuitive and requires users to explore the elipses button to know this is possible

**Describe the solution you'd like**
Add a "Version" column to the left of the "Created" column with the version number.  The version number should be hyperlinked to the existing versions view.

**Describe alternatives you've considered**
It might also be possible to add some or all of the 4 columns of the versions view to the main DMP view, so that a versions view is no longer necessary.  However, this could make the DMP view too cluttered.

**Additional context**
An info box with instructions on the versions column header could be helpful, but by having the version number hyperlinked it should be intuitive to the user what they will get by clicking on the number. 

**Screenshot**
Image

GeoffreyKarnbach:
## Description
Feature / Bugfix

#### What does this PR do?
Added two fields to the DMPListItemDO (for the information to be displayed in the Frontend DMP Table).

#### Breaking changes
Fixed major bug, disabling the versioning feature in the DB by checking for a max length for the name of a new versioning of a DMP.

#### Dependencies
Frontend PR:
https://github.com/tuwien-csd/damap-frontend/pull/354

Issue:
https://github.com/tuwien-csd/damap-frontend/issues/346

### DIFF
--- a/src/main/java/org/damap/base/rest/version/VersionService.java
+++ b/src/main/java/org/damap/base/rest/version/VersionService.java
@@ -45,6 +45,12 @@ public class VersionService {
    */
   @Transactional
   public VersionDO createOrUpdate(VersionDO versionDO) {
+
+    if (versionDO.getVersionName().length() > 255) {
+      throw new IllegalArgumentException(
+          "Version name must be less than or equal to 255 characters");
+    }
+
     if (versionDO.getId() != null) return update(versionDO);
     else return create(versionDO);
   }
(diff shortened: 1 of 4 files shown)


src/main/java/org/damap/base/rest/version/VersionService.java
```diff
@@ -45,6 +45,11 @@ public List<VersionDO> getDmpVersions(long dmpId) {
    */
   @Transactional
   public VersionDO createOrUpdate(VersionDO versionDO) {
+
+    if (versionDO.getVersionName().length() > 255) {
+      throw new IllegalArgumentException("Version name must be less than 255 characters");
```
ValentinFutterer: nitpick: should be `less than or equal to`

ValentinFutterer (approved):
Nice work, lets merge.

PR #279, opened 2024-11-11
Export: Remove storage description from export if it null in the translation

### LINKED ISSUE #273: Null exported for storage
**Describe the bug**
When exporting, `null` is exported in the template. This concerns all three templates.

**Is this a regression?**
No.

**Steps to Reproduce**
1. Login as an admin
2. Go to admin panel
3. Create a storage and for the translation, only enter a title and select a language
4. Create a DMP
5. Specify a new dataset
6. Go to step 5
7. Click on the `Storage` tab
8. Select the previously created storage
9. Select the previously defined dataste
10. Export with any template
11. Open the exported document
12. Find the section about the storage

Screenshot from 2024-10-23 09-35-33

**Expected behavior**
`null` should not be exported. Depending on what the real value should be, either an empty string or a default value may be exported.

**Screenshots**

**Environment**
 - Damap version 4.2.2

**Additional context**

GeoffreyKarnbach:
## Description
Remove storage description from export if it is null (to avoid ": null" in exported Word document)

#### What does this PR do?
Remove storage description from export if it is null (to avoid ": null" in exported Word document)

closes GH-273

### DIFF
--- a/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -424,9 +424,10 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
                       + " "
                       + loadResourceService.loadVariableFromResource(prop, "distributionStorage")
                       + " "
-                      + hostVar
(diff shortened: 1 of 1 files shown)


ValentinFutterer (approved):
Fixed it

PR #285, opened 2024-11-20
DMP: Add project coordinator and principal investigator roles for contributors

### LINKED ISSUE #326: In Step 2, add Project Coordinator and Principal Investigator
**Changes**
In Step 2, add “Project Coordinator and Principal Investigator”.  If a Principal Investigator is selected, it should be populated in the “PI” field in the first table for the FWF template. If a Principal Investigator or Project Coordinator is selected, it should be populated in the “Project coordinator, Principal Investigator” field under the Project details table for the SE template. 

**Screenshot**
Image

**Changes**
Coud break into separate issue as this one is less important: For the Horizon Europe template the name and institution of the person creating the DMP should go in the table “Delivery Slip”

**Screenshot**
Image

GeoffreyKarnbach:
## Description
Added "Principle Investigator" and "Project Coordinator" to the enum of roles for individual contributors in Section 2.
Added proper display of users with either PC or PI role in the corresponding tables in the different export templates.

### DIFF
--- a/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -9,7 +9,9 @@ import lombok.extern.jbosslog.JBossLog;
 import org.apache.poi.xwpf.usermodel.*;
 import org.damap.base.domain.*;
 import org.damap.base.enums.*;
+import org.damap.base.rest.dmp.domain.ContributorDO;
 import org.damap.base.rest.dmp.domain.ProjectDO;
+import org.damap.base.rest.dmp.mapper.ContributorDOMapper;
 import org.openxmlformats.schemas.wordprocessingml.x2006.main.CTVMerge;
 
 /**
@@ -21,6 +23,7 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
     extends AbstractTemplateExportSetup {
 
   protected Map<Long, String> datasetTableIDs = new HashMap<>();
+  private List<String> projectCoordinatorOrInvestigatorIds = new ArrayList<>();
 
   /** {@inheritDoc} */
   @Override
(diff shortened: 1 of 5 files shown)


src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -363,11 +386,24 @@ public void datasetsInformation() {
 
   /** storageIntroInformation */
   public void storageIntroInformation() {
-    String coordinatorFullName;
+    String coordinatorFullName = null;
+
+    if (!projectCoordinator.isEmpty()) {
+      // Search for Principal Investigator
+      for (ContributorDO contributor : projectCoordinator) {
+        if (contributor.getRole().equals(EContributorRole.PRINCIPAL_INVESTIGATOR)) {
```
ValentinFutterer: Missing null check here

ValentinFutterer:
Something must have happened while rewriting the code. This input
image

leads to this HE export:
image

As i understand it, contributors already listed under project coordinator and principal investigator should not be in contributors. 

Could you share what Christiane told you about how to display those two roles.

ValentinFutterer:
This input causes an error from the backend. Probably connected to the fact that contributors fetched dynamically during export have no roles assigned. Code needs to be rewritten to handle this null, wherever these fetched contributors are used.

image

src/main/java/org/damap/base/conversion/AbstractTemplateExportSetup.java
```diff
@@ -38,7 +38,7 @@ public abstract class AbstractTemplateExportSetup extends AbstractTemplateExport
   protected Properties prop = null;
   protected List<XWPFParagraph> xwpfParagraphs = null;
   protected List<XWPFTable> xwpfTables = null;
-  protected ContributorDO projectCoordinator = null;
+  protected List<ContributorDO> projectCoordinator = new ArrayList<>();
```
ValentinFutterer: This list should be renamed to projectCoordinators, since there can be multiple coordinators.

src/main/java/org/damap/base/conversion/AbstractTemplateExportSetup.java
```diff
@@ -53,18 +53,27 @@ protected void exportSetup(long dmpId) {
...
-          ContributorDOMapper.mapEntityToDO(projectLeaderOpt.get(), new ContributorDO());
-    else
+                contributor ->
+                    contributor.getContributorRole() == EContributorRole.PROJECT_LEADER
+                        || contributor.getContributorRole() == EContributorRole.PROJECT_COORDINATOR
+                        || contributor.getContributorRole()
+                            == EContributorRole.PRINCIPAL_INVESTIGATOR)
+            .toList();
+
+    if (!projectLeaderOpt.isEmpty()) {
```
ValentinFutterer: Java streams .toList() already return an empty list when no element was found, there is no reason for this intermediate list.

src/main/java/org/damap/base/conversion/AbstractTemplateExportSetup.java
```diff
@@ -53,18 +53,27 @@ protected void exportSetup(long dmpId) {
...
+      projectCoordinator = new ArrayList<>();
+      for (Contributor contributor : projectLeaderOpt) {
+        projectCoordinator.add(ContributorDOMapper.mapEntityToDO(contributor, new ContributorDO()));
+      }
+    } else
       try {
         if (dmp.getProject() != null && dmp.getProject().getUniversityId() != null)
-          projectCoordinator = projectService.getProjectLeader(dmp.getProject().getUniversityId());
+          projectCoordinator =
+              List.of(projectService.getProjectLeader(dmp.getProject().getUniversityId()));
```
ValentinFutterer: The contributor returned from the project service should be assigned the role project leader, if he/she has no role assigned from the service.

ValentinFutterer:
Another small request, can you please rewrite the commit message from `DMP:` to `Contributor:` . `DMP:` is a little too generic.

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -363,11 +391,26 @@ public void datasetsInformation() {
 
   /** storageIntroInformation */
   public void storageIntroInformation() {
-    String coordinatorFullName;
+    String coordinatorFullName = null;
```
ValentinFutterer, line 394: I have queried Christiane and Tomasz for the behaviour we want in this part. 
 The first contributor choice should be the data manager role. So we need to query all contributors and not only the coordinators.
 Second choice is the project leader, then prinvcipal investigator then project coordinator. 
 If none of these roles exist, the fallback will be the contact person.
 
As a last resort, you should choose the string `"the project leader`.

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -363,11 +391,26 @@ public void datasetsInformation() {
...
+      if (coordinatorFullName == null) {
+        // If no Principal Investigator is found, use the first coordinator
+        coordinatorFullName =
+            projectCoordinator.get(0).getFirstName()
+                + " "
+                + projectCoordinator.get(0).getLastName();
+      }
     } else {
       coordinatorFullName = "Coordinator";
     }
```
ValentinFutterer: I have queried Christiane and Tomasz for the behaviour we want in this part. 
 The first contributor choice should be the data manager role. So we need to query all contributors and not only the coordinators.
 Second choice is the project leader, then prinvcipal investigator then project coordinator. 
 If none of these roles exist, the fallback will be the contact person.
 
As a last resort, you should choose the string `"the project leader`.

src/main/java/org/damap/base/conversion/AbstractTemplateExportSetup.java
GeoffreyKarnbach: Agree

src/main/java/org/damap/base/conversion/AbstractTemplateExportSetup.java
GeoffreyKarnbach: Done

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
GeoffreyKarnbach: Implemented :)

ValentinFutterer:
There is still one small bug. With no contributor added, the contributor table looks like this:

image

image

Since the missing project leader is fetched from the project-db.

But this info os not used in the  `storageIntroInformation()` function, leading to this:

image

The fallback is triggered, although a name for the project leader would be available. A possible solution could be to use the 'projectCoordinators' list in `storageIntroInformation()`, instead of only the `contributors` list.

src/main/java/org/damap/base/conversion/AbstractTemplateExportSetup.java
```diff
@@ -53,21 +53,31 @@ protected void exportSetup(long dmpId) {
...
       try {
-        if (dmp.getProject() != null && dmp.getProject().getUniversityId() != null)
-          projectCoordinator = projectService.getProjectLeader(dmp.getProject().getUniversityId());
+        if (dmp.getProject() != null && dmp.getProject().getUniversityId() != null) {
+          ContributorDO projectLeader =
+              projectService.getProjectLeader(dmp.getProject().getUniversityId());
+          if (projectLeader.getRole() == null) {
+            projectLeader.setRole(EContributorRole.PROJECT_LEADER);
+          }
+          projectCoordinators = List.of(projectLeader);
```
ValentinFutterer, line 75: This can be slightly refactored, projectCorrdinators is already a list at that point.

src/main/java/org/damap/base/conversion/AbstractTemplateExportSetup.java
```diff
@@ -53,21 +53,31 @@ protected void exportSetup(long dmpId) {
...
+        } else projectCoordinators = new ArrayList<>();
```
ValentinFutterer: This can be slightly refactored, since projectCoordinators is already a list at that point.

ValentinFutterer (approved):
Mergable 👍

PR #296, opened 2024-11-27
Live Preview: Add possibility to disable the feature in the config file sent to the frontend

### LINKED ISSUE #287: Preview: Make feature optional
**Changes**
Zeno requested this feature to be optional, like the fits service, which seems like a good idea to maintain consistency in the application regarding external services.

**Additional notes**
Zeno also mentioned that he knows of a few libraries, that could provide the same functionality as our current external container guttenberg does. get in touch with him and see what he knows about it. It would be a nice change, if we could handle this with a dependency. 

makign the feature optional will require frontend-changes too!

**Acceptance Criteria**
Preview feature is optional via configuration option in the backend (like fits service)
Guttenberg service functionality is taken over by a library (if possible)

GeoffreyKarnbach:
## Description
Add to config file sent to the client whether or not the Live Preview Feature is available (by checking for a Gotenberg URL in the application.yaml file).

closes GH-287

### DIFF
--- a/src/main/java/org/damap/base/rest/ConfigResource.java
+++ b/src/main/java/org/damap/base/rest/ConfigResource.java
@@ -39,6 +39,9 @@ public class ConfigResource {
   @ConfigProperty(name = "damap.fits-url")
   Optional<URL> fitsUrl;
 
+  @ConfigProperty(name = "damap.gotenberg-url")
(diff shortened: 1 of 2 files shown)


ValentinFutterer (approved):
Nice work!

PR #298, opened 2024-12-02
Global: Add sorting to DMPs, Storage, Datasets and other entities by ID to preserve order

### LINKED ISSUE #293: DB: Preserve sequence of elements in lists
**Changes**
Users have complained about thesequence of elements changing, e.g. datasets. The sequence of lists should be preserved by the db, so that list elements in the frontend do not change places continously. This also affects the export. Ascertain how this could be achieved in the db and how much effort it would be. 

**Assumptions**
This is probbaly a lot of work - if its too much move this into discussion and discuss with the team if its worth it.

**Acceptance Criteria**
Ordering is preserved.

GeoffreyKarnbach:
## Description

Feature / Functionality

#### What does this PR do?
Add sorting by ID for DMPs (getAll and get for logged in user), Storages, Datasets, Contributors and Costs within a DMP.

closes GH-293

### DIFF
--- a/src/main/java/org/damap/base/rest/dmp/service/DmpService.java
+++ b/src/main/java/org/damap/base/rest/dmp/service/DmpService.java
@@ -5,11 +5,7 @@ import jakarta.inject.Inject;
 import jakarta.transaction.Transactional;
 import jakarta.validation.Valid;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 import lombok.extern.jbosslog.JBossLog;
 import org.damap.base.domain.Access;
 import org.damap.base.domain.Contributor;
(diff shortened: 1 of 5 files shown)


ValentinFutterer:
Hey, this works fine, but wouldnt it be simpler to do the sorting in the Repo classes. The you dont need to distribute the code over Services and Mappers. For example in DMPRepo:
```
public List<Dmp> getAll() {
    return listAll(Sort.by("id"));
  }
```
 [Read here for more info](https://quarkus.io/guides/hibernate-orm-panache#sorting).

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
```diff
@@ -67,6 +63,9 @@ public List<DmpListItemDO> getAll() {
             dmpListItemDOList.add(
                 DmpListItemDOMapper.mapEntityToDO(
                     null, dmp, new DmpListItemDO(), versionService.getDmpVersions(dmp.id))));
+
+    // dmpListItemDOList.sort(Comparator.comparing(DmpListItemDO::getId));
```
ValentinFutterer: Minor: Forgotten commented out code

GeoffreyKarnbach: done

ValentinFutterer (approved):
Approved

PR #302, opened 2024-12-11
Ethical: Add ethical report URI to datamodel and DO, that can be disabled throught the config file

### LINKED ISSUE #292: Ethics: Add ethical issues report to DAMAP
**Description**
From the [RDA-Common-standard](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard?tab=readme-ov-file#ethical_issues_report:~:text=yes-,ethical_issues_report,-To%20indicate%20where). This report is not yet in DAMAP, eventhough it is in the common standard. Med Uni Graz requires this, therefore we should implement it. Find a way to add it to our data model. Also ensure that it is in the form of an URI. It would also be a good idea to only allow this field when `yes` was chosen in this question:
image

But run this by Christiane and Tomek first, since this yes/no question is not part of the common standard.

**Proposed solution**
Expand our datamodel with the ethical issues report. Also make sure that the field has the form of an URI and is only filled when yes was chosen for the above question (ask Tomek and Christiane first). You also have to add this new value to our export templates, discuss with Christiane how to do this.

See for the [frontend issue here](https://github.com/tuwien-csd/damap-frontend/issues/382).

GeoffreyKarnbach:
## Description
Feature

#### What does this PR do?
Add new Ethical Report URI field in the DB, DO. Possibility to disable it using the application.yaml file using the ConfigDO sent to the frontend.

#### Dependencies
[Related Fronted PR](https://github.com/tuwien-csd/damap-frontend/pull/397)

closes GH-292

### DIFF
--- a/src/main/resources/org/damap/base/template/horizonEuropeTemplate.resource
+++ b/src/main/resources/org/damap/base/template/horizonEuropeTemplate.resource
@@ -55,6 +55,7 @@ legalRights.contact.default=Our institution has rights to the produced data and 
 
 ethical.no=No particular ethical issue is foreseen with the data to be used or produced by the project. This section will be updated if issues arise.
 ethicalReviewed.avail=The research plan of the project was reviewed by an ethics committee / a similar body.
+ethicalReportIntro=Ethical issue report number:
 ethicalStatement=Ethical issues in the project have been identified and discussed with the Ethics Coordinator. They are described in detail in separate documents.
 
 closeddatasetreasons.intro=Closed Dataset Reasons:
--- a/src/main/resources/org/damap/base/template/scienceEuropeTemplate.resource
+++ b/src/main/resources/org/damap/base/template/scienceEuropeTemplate.resource
@@ -49,6 +49,7 @@ legalRights.contact.default=Our institution has rights to the produced data and 
 
 ethical.no=No particular ethical issue is foreseen with the data to be used or produced by the project. This section will be updated if issues arise.
 ethicalReviewed.avail=The research plan of the project was reviewed by an ethics committee / a similar body.
+ethicalReportIntro=Ethical issue report number:
 ethicalStatement=Ethical issues in the project have been identified and discussed with the Ethics Coordinator. They are described in detail in separate documents.
 
 closeddatasetreasons.intro=Closed Dataset Reasons:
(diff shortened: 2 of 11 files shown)


src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.3.0_4.yaml
```diff
@@ -0,0 +1,24 @@
+databaseChangeLog:
+  - changeSet:
+      id: 18
+      author: Geoffrey Karnbach
```
ValentinFutterer: The file should be renamed, we are in version 4.3.1

ValentinFutterer:
Looks pretty good overall, but i have a few things to add.

First off, we have changed the dataytpe to string, so i would remove all mentions of URL or URI and just rename it to ethical issues report, like in the standard. 

I dont know why you touched the HE and SE templates, please add the review documents.
These changes might have broken something, i get these warnings:
image

In the same vein, could you add Changelog statements to the PR, since this PR touched the templates, the resource files and added new functionality that can be disabled (all interesting things for maintainers)

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.3.1_1.yaml
```diff
@@ -0,0 +1,24 @@
+databaseChangeLog:
+  - changeSet:
+      id: 19
```
ValentinFutterer: 4.3.1_1 is now taken, should be 4.3.1_2

src/main/resources/org/damap/base/template/horizonEuropeTemplate.resource
```diff
@@ -55,6 +55,7 @@ legalRights.contact.default=Our institution has rights to the produced data and
 
 ethical.no=No particular ethical issue is foreseen with the data to be used or produced by the project. This section will be updated if issues arise.
 ethicalReviewed.avail=The research plan of the project was reviewed by an ethics committee / a similar body.
+ethicalReportIntro=Ethical report URI:
```
ValentinFutterer: Rename to "Ethical issue report number:"

src/main/resources/org/damap/base/template/scienceEuropeTemplate.resource
```diff
@@ -49,6 +49,7 @@ legalRights.contact.default=Our institution has rights to the produced data and
 
 ethical.no=No particular ethical issue is foreseen with the data to be used or produced by the project. This section will be updated if issues arise.
 ethicalReviewed.avail=The research plan of the project was reviewed by an ethics committee / a similar body.
+ethicalReportIntro=Ethical report URI:
```
ValentinFutterer: Rename to "Ethical issue report number:"

ValentinFutterer (approved):
Approved

PR #312, opened 2024-12-23
Bugfix: Add null check when updating the project lead with manual contributors

### LINKED ISSUE #305: Its is impossible to switch projects when a manual contributor was added
**Describe the bug**
If you try to switch projects, you get an error when a manual contributor is assigned to the project. This happens, because the `updateProjectLead` method is called, when the project is switched. It tries to use the universityId of the contributor, but its null since he/she was added manually. See line 300 in DMPService. Add a null check, to prevent this from happening

**Is this a regression?**
No.

**Steps to Reproduce**
Add a manual project.
Add a manual contributor.
Delete the project, save.
Add another project and try to save.

**Expected behavior**
Switching projects should be no issue.

**Environment**
 - Damap version 4.3.1

GeoffreyKarnbach:
## Description
Bugfix

#### What does this PR do?
See PR Title

closes GH-305

### DIFF
--- a/src/main/java/org/damap/base/rest/dmp/service/DmpService.java
+++ b/src/main/java/org/damap/base/rest/dmp/service/DmpService.java
@@ -297,7 +297,10 @@ public class DmpService {
 
     Optional<Contributor> alreadyExistingContributorLeader =
         dmpContributors.stream()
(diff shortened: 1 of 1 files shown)


ValentinFutterer (approved):
Works 👍

PR #313, opened 2024-12-23
Bugfix: Add null check in export method for dataset identifier

### LINKED ISSUE #306: Missing null check during existing dataset export
**Describe the bug**
Starting at line 1164 in `composeTableReusedDatasets` in `AbstractTemplateExportScienceEuropeComponents` :
```
        if (dataset.getDatasetIdentifier() != null) {
          docVar.add(dataset.getDatasetIdentifier().getIdentifier());
        } else {
          docVar.add("");
        }
```

This code only checks if the identifier is null, but it does not check if the identifier field is null. This causes an error in `AbstractTemplateExportFunctions`, `insertTableCells` : 
```
List<String> lines = cellContent.get(i).lines().toList();
```

Both parts of the code should receive according null checks.

The bug causes certain legacy dmps in our demo instance to be unexportable, since they have identifiers where onyl the type is set. But this could theoretically also happen in production, when OpenAire returns an empty identifer (didnt check if this is possible)
You can also trigger it, when you manually add an existing dataset with a source and then delete the source. Then try to export. You should look into why this exact behaviour happens.

**Is this a regression?**
Yes.

**Steps to Reproduce**
Log into the db and add a dataset identifer with only the type set and the identifer string null. Then try to export. Or try to export DMP: 2146 in our demo instance.

Alternatively manually create an existing dataset with a source. Then delete the source and try to export.

**Expected behavior**
Exporting should still work.

**Environment**
 - Damap version 4.3.1

GeoffreyKarnbach:
## Description
Bugfix

#### What does this PR do?
See PR title

closes GH-306

### DIFF
--- a/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -1175,7 +1175,8 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
         docVar.add(datasetTableIDs.get(dataset.id));
         docVar.add(Optional.ofNullable(dataset.getTitle()).orElse(""));
 
-        if (dataset.getDatasetIdentifier() != null) {
+        if (dataset.getDatasetIdentifier() != null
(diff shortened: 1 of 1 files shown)


ValentinFutterer (approved):
👍

PR #314, opened 2024-12-23
Export: Use Objects.equals to perform a null-safe enum check

### LINKED ISSUE #308: Faulty enum comparison
**Describe the bug**
In `AbstractTemplateExportScienceEuropeComponents`, enums are compared using toString. They should be compared with a null safe Objects.equal though. This error causes the coordinator export to not export the id.

image

Also try to find other instances of this wrong comparison, if they still occur.

**Is this a regression?**
Yes.

**Expected behavior**
Correct comparison should be done, fixing the bug.

**Environment**
 - Damap version 4.3.1

GeoffreyKarnbach:
## Description
Bugfix

#### What does this PR do?
See PR Title

closes GH-308


ValentinFutterer (approved):
Good 👍

PR #315, opened 2024-12-23
Export: Fix bug with null ORCID in exported documented, remove ORCID if value is null from export

### LINKED ISSUE #310: Manually added contributor without ORCId is exported as ORCID = null
**Describe the bug**
A manually added contributor without an ORCID is exported like this
image

And on a sidenote, it says ORCID iD, with a lowercase i and highercase D. It should probably say just ORCID.

**Is this a regression?**
Yes.

**Expected behavior**
It should not show the ORCID at all, when none is given through the frontend.

**Environment**
 - Damap version 4.3.1

GeoffreyKarnbach:
## Description
Bugfix

#### What does this PR do?
See PR title

closes GH-310


ValentinFutterer (approved):
Solves the issue 👍

PR #317, opened 2024-12-30
Mock Service: Add empty check before accessing first element of a response list

### LINKED ISSUE #304: MockProjectServiceImpl crashes when you try to read a project that doesnt exist
**Describe the bug**
This bug can be observed on our demo instance. If you try to export a dmp with the project `Recommended Project x`, it fails. This happens due to a bug in `MockProjectServiceImpl.read()`. It calls get(0) on an empty list.

You should check the complete implementation for similar errors. Same for the other mock services.
 
**Is this a regression?**
No.

**Steps to Reproduce**
Go to the demo instance and try to export a dmp with a recommended project. Stacktrace: 
[sotirios.tsepelakis - Mon Dec 16 2024 13_44_21 GMT+0100 (Central European Standard Time).txt](https://github.com/user-attachments/files/18150290/sotirios.tsepelakis.-.Mon.Dec.16.2024.13_44_21.GMT%2B0100.Central.European.Standard.Time.txt)

**Expected behavior**
Should export without error.

**Environment**
 - Damap version 4.3.1

GeoffreyKarnbach:
## Description
Bugfix

#### What does this PR do?
Ensure that get is not used on an empty list with a hardcoded access index of 0.

closes GH-304

### DIFF
--- a/src/main/java/org/damap/base/rest/persons/MockUniversityPersonServiceImpl.java
+++ b/src/main/java/org/damap/base/rest/persons/MockUniversityPersonServiceImpl.java
@@ -3,6 +3,7 @@ package org.damap.base.rest.persons;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.MultivaluedMap;
+import java.util.List;
 import org.damap.base.rest.base.ResultList;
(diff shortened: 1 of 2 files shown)


ValentinFutterer (approved):
Fixes the bug nicely :)

PR #322, opened 2025-01-20
Version: Change naming mechanism for editor

### LINKED ISSUE #291: Version-View: Editor names are not human readable
**Describe the bug**
The names of editors are taken from the SecurityService, which in turn takes them from the OidcJwtCallerPrincipal which gets the name from the Jwt token. This library tries different fields in the token. First the upn (User Principal Name), which is structured similar to an email. If it cannot get the upn it tries the preferred_username, which is most likely a human-readable username in a system. Lastly, it tries to use the sub (Subject), which is an identifier number. Depending on the identity provider, different values are returned here, and most of them are not very human readable.

**Is this a regression?**
No.

**Expected behavior**
The editor should be at least the human name of the editor (firstname lastname). The maybe we can add some other differential features to avoid naming conflicts. Maybe ask Zeno how he would handle this, since he works with the tiss API, he could have some good ideas. Additionally, think about a solution of how to apply this change retroactively to the editors of old versions. Otherwise, there will be an abrupt break in naming scheme. Also think about name changes in the CRIS systems and how to handle those.

**Screenshots**
This is from the TU Wien DMP Tool. in DAMAP it would just say `user`.
image

**Environment**
 - Damap version 4.3.0

GeoffreyKarnbach:
## Description
Bugfix

#### What does this PR do?
Add new method to security service with claims on JWT to use correct display name for editor in the version view.

closes GH-291

### DIFF
--- a/src/main/java/org/damap/base/security/SecurityService.java
+++ b/src/main/java/org/damap/base/security/SecurityService.java
@@ -55,6 +55,33 @@ public class SecurityService {
     return ((OidcJwtCallerPrincipal) principal).getName();
   }
 
+  public String getDisplayName() {
+    final Principal principal = securityIdentity.getPrincipal();
+    if (!(principal instanceof OidcJwtCallerPrincipal)) {
+      return null;
+    }
+
+    OidcJwtCallerPrincipal oidcPrincipal = (OidcJwtCallerPrincipal) principal;
+
+    String name = getClaimValueAsString(oidcPrincipal, "name");
+    if (name != null) {
+      return name;
+    }
+
+    String firstName = getClaimValueAsString(oidcPrincipal, "given_name");
+    String lastName = getClaimValueAsString(oidcPrincipal, "family_name");
+    if (firstName != null && lastName != null) {
+      return firstName + " " + lastName;
+    }
+
+    return getClaimValueAsString(oidcPrincipal, "email");
+  }
+
+  private String getClaimValueAsString(OidcJwtCallerPrincipal oidcPrincipal, String claimKey) {
+    Object claimValue = oidcPrincipal.getClaims().getClaimValue(claimKey);
+    return claimValue != null ? claimValue.toString() : null;
+  }
+
   /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/damap/base/security/SecurityService.java
```diff
@@ -55,6 +55,29 @@ public String getUserName() {
...
+        ((OidcJwtCallerPrincipal) principal).getClaims().getClaimValue("email").toString();
+
+    String displayName = name;
+    if (name == null) {
+      if (firstName != null && lastName != null) {
+        displayName = firstName + " " + lastName;
+      } else {
+        displayName = email;
+      }
+    }
```
ValentinFutterer: Checking if name == null  will never be true, since in this case `getClaimValue("name").toString()` would throw an exception. I would first check (null safe) if name exists. If yes, use it as displayName. If no, check for givenName and family_name...

ValentinFutterer:
There should also be a plan on how to handle this retroactively, since the editor names for older dmps are just unreadable hashes (at least for TU Wien). Check with TU Graz and Med Uni Graz on how to proceed, if they all have the same problem, i would set old editor names to null, sicne seeing nothing is better than the users seeing random hashes with no visible meaning behind it.

src/main/java/org/damap/base/security/SecurityService.java
GeoffreyKarnbach: solved

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.3.0_4.yaml
```diff
@@ -0,0 +1,12 @@
+databaseChangeLog:
+  - changeSet:
+      id: 17
```
ValentinFutterer: This id should be 18
Also we are on 4.3.1

ValentinFutterer (approved):
Ok lets merge

PR #332, opened 2025-02-05
Add the null safe comparation 

### LINKED ISSUE #309: Make export null safe
**Changes**
Almost the whole backend uses .equals() to chekc for equality, but this causes an error, when it is called on null (which happens a lot). To remove future bugs and spare us a lot of hassle, we should replace all instances of .equals with Objects.equals. 

Some lists are accessed without checking, if it is empty or if the content is null. You should also find those instances and deal with them. (try search for .get() or similar to find them)

**Additional notes**
Sometimes `x != null` are used beforehand. Check the logic and see if they can be removed after adding Objects.equals. 

Also be mindful, that Objects.equals returns true when bove inputs are null. If this behaviour is unwanted in that instance, think of an alternative.

andresTabiTuwien:
https://github.com/tuwien-csd/damap-backend/issues/309

#### What does this PR do?

Add the function to compare the nullability and some functions to get the nullability of some properties.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-39

### DIFF
--- a/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -1,5 +1,7 @@
 package org.damap.base.conversion;
 
+import static org.damap.base.utils.EqualityUtils.nullExclusiveEquals;
+
 import java.text.NumberFormat;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
@@ -52,10 +54,10 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
     int reuseIDprogression = 0;
 
     for (Dataset dataset : datasets) {
-      if (dataset.getSource().equals(EDataSource.NEW)) {
+      if (nullExclusiveEquals(dataset.getSource(), EDataSource.NEW)) {
         datasetTableIDs.put(dataset.id, "P" + ++newIDprogression);
       }
-      if (dataset.getSource().equals(EDataSource.REUSED)) {
+      if (nullExclusiveEquals(dataset.getSource(), EDataSource.REUSED)) {
         datasetTableIDs.put(dataset.id, "R" + ++reuseIDprogression);
       }
     }
@@ -111,15 +113,12 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
       fundingItems.add(projectCRIS.getFunding().getFundingProgram());
     }
     // add grant number to funding item variables
-    if (project.getFunding() != null
-        && project.getFunding().getGrantIdentifier().getIdentifier() != null) {
-      fundingItems.add(project.getFunding().getGrantIdentifier().getIdentifier());
+    if (project.getFundingGrantIdentifierIdentifier() != null) {
+      fundingItems.add(project.getFundingGrantIdentifierIdentifier());
     }
 
-    if (project.getFunding() != null
-        && project.getFunding().getFunderIdentifier().getIdentifier() != null) {
-      addReplacement(
-          replacements, "[funderid]", project.getFunding().getFunderIdentifier().getIdentifier());
+    if (project.getFundingFunderIdentifierIdentifier() != null) {
+      addReplacement(replacements, "[funderid]", project.getFundingFunderIdentifierIdentifier());
     } else {
       addReplacement(replacements, "[funderid]", "");
     }
@@ -139,7 +138,7 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
 
     if (personIdentifier != null && personIdentifier.getIdentifier() != null) {
       String contactIdentifierId = personIdentifier.getIdentifier();
-      if (personIdentifier.getIdentifierType().equals(EIdentifierType.ORCID)) {
+      if (nullExclusiveEquals(personIdentifier.getIdentifierType(), EIdentifierType.ORCID)) {
         identifier = "ORCID: " + contactIdentifierId;
       }
     }
@@ -153,7 +152,7 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
 
     if (affiliationIdentifier != null) {
       String contactAffiliationIdentifierId = affiliationIdentifier.getIdentifier();
-      if (affiliationIdentifier.getIdentifierType().equals(EIdentifierType.ROR)) {
+      if (nullExclusiveEquals(affiliationIdentifier.getIdentifierType(), EIdentifierType.ROR)) {
         identifier = "ROR: " + contactAffiliationIdentifierId;
       }
     }
@@ -233,7 +232,7 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
       if (contributor.getPersonId() != null) {
         coordinatorIdentifierId = contributor.getPersonId().getIdentifier();
 
-        if (Objects.equals(contributor.getPersonId().getType(), EIdentifierType.ORCID)) {
+        if (nullExclusiveEquals(contributor.getPersonId().getType(), EIdentifierType.ORCID)) {
           String coordinatorId = "ORCID iD: " + coordinatorIdentifierId;
           coordinatorProperties.add(coordinatorId);
         }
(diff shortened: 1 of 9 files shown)


src/main/java/org/damap/base/utils/EqualityUtils.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class EqualityUtils {
+  /**
+   * Checks the nullability of two objects and compares them for equality.
+   *
+   * @param o1 the first object
+   * @param o2 the second object
+   * @return false if both are null. if they are equal return true, false otherwise
+   */
+  public static boolean nullSafeEquals(Object o1, Object o2) {
```
ValentinFutterer: A better name for the class would maybe be NullExclusiveEquals, to show how it differs from normal Objects.equals(), since it is already null safe.

Open for other naming suggestions :)

src/main/java/org/damap/base/utils/PropertyValidator.java
```diff
@@ -0,0 +1,74 @@
+package org.damap.base.utils;
+
+import org.damap.base.domain.Dataset;
+import org.damap.base.domain.Dmp;
+import org.damap.base.domain.Project;
+
+public class PropertyValidator {
```
ValentinFutterer: I would remove this class and move the functionality inside the classes you give as method arguments. The methods should return the wanted object or null, if not possible. To make it clear what they do, maybe rename them to `getFundingGrantidentifierIdentifier` or something similar. Or just `getGrantIdentifierIdentifier`?

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -316,9 +318,12 @@ private void dmpContributorInformation() {
...
-                  || contributor.getContributorRole().equals(EContributorRole.PROJECT_COORDINATOR)))
+              && (nullSafeEquals(
+                      contributor.getContributorRole(), EContributorRole.PRINCIPAL_INVESTIGATOR)
+                  || nullSafeEquals(
+                      contributor.getContributorRole(), EContributorRole.PROJECT_LEADER)
+                  || nullSafeEquals(
+                      contributor.getContributorRole(), EContributorRole.PROJECT_COORDINATOR)))
           || this.projectCoordinatorOrInvestigatorIds.contains(contributor.getUniversityId())) {
         continue;
       }
```
ValentinFutterer: The if is unnecessary here, if you use the null safe comparison

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -455,7 +461,8 @@ public void storageIntroInformation() {
       for (Contributor contributor : contributorPool) {
         if (contributor != null
             && contributor.getContributorRole() != null
-            && contributor.getContributorRole().equals(EContributorRole.PROJECT_COORDINATOR)) {
+            && nullSafeEquals(
+                contributor.getContributorRole(), EContributorRole.PROJECT_COORDINATOR)) {
           coordinatorFullName = contributor.getFirstName() + " " + contributor.getLastName();
           break;
         }
```
ValentinFutterer, line 464: `contributor.getContributorRole() != null` is unnecessary, due to the null safe check

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -1114,7 +1121,7 @@ public void composeTableNewDatasets(XWPFTable xwpfTable) {
...
 
         if (dataset.getSensitiveData() != null) {
```
ValentinFutterer: if is superflous now

andresTabiTuwien: What this means ?

src/main/java/org/damap/base/domain/Dataset.java
```diff
@@ -135,4 +135,17 @@ public List<Repository> getRepositories() {
...
     return id;
   }
+
+  /**
+   * Validates that the final value in the chain dataset.getDmp().getTargetAudience() is not null.
+   *
+   * @return the value of the targetAudience if it's not null.
+   * @return null if any part of the chain is null.
+   */
+  public String getNonNullDatasetTargetAudience() {
```
ValentinFutterer: You should keep the naming scheme consistent across these methods. See `getFundingGrantIdentifierIdentifier()` in Project.

src/main/java/org/damap/base/domain/Dataset.java
```diff
@@ -135,4 +135,17 @@ public List<Repository> getRepositories() {
...
+    if (this.getDmp() == null || this.getDmp().getTargetAudience() == null) {
```
ValentinFutterer: The second part after the || is not necessary.

src/main/java/org/damap/base/domain/Dataset.java
ValentinFutterer: The same is true for the 3 other methods of this kind.

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -316,9 +315,12 @@ private void dmpContributorInformation() {
...
-                  || contributor.getContributorRole().equals(EContributorRole.PROJECT_COORDINATOR)))
+              && (nullExclusiveEquals(
+                      contributor.getContributorRole(), EContributorRole.PRINCIPAL_INVESTIGATOR)
+                  || nullExclusiveEquals(
+                      contributor.getContributorRole(), EContributorRole.PROJECT_LEADER)
+                  || nullExclusiveEquals(
+                      contributor.getContributorRole(), EContributorRole.PROJECT_COORDINATOR)))
           || this.projectCoordinatorOrInvestigatorIds.contains(contributor.getUniversityId())) {
         continue;
       }
```
ValentinFutterer: `contributor.getContributorRole() != null` is not necessary, since this case is covered by the new equals method.

src/main/java/org/damap/base/domain/Dataset.java
```diff
@@ -135,4 +135,17 @@ public List<Repository> getRepositories() {
   public Long getId() {
     return id;
   }
+
+  /**
+   * Validates that the final value in the chain dataset.getDmp().getTargetAudience() is not null.
+   *
+   * @return the value of the targetAudience if it's not null.
+   * @return null if any part of the chain is null.
+   */
```
ValentinFutterer: The description needs to be updated, its still like it was before the refactoring.

src/main/java/org/damap/base/domain/Dmp.java
```diff
@@ -209,4 +209,17 @@ public Contributor getContact() {
             .findFirst();
     return contact.orElse(null);
   }
+
+  /**
+   * Validates that the final value in the chain dmp.getProject().getUniversityId() is not null.
+   *
+   * @return the value of the UniversityID if it's not null.
+   * @return null if any part of the chain is null.
+   */
```
ValentinFutterer: The description needs to be updated, its still like it was before the refactoring.

src/main/java/org/damap/base/domain/Project.java
```diff
@@ -41,4 +41,36 @@ public class Project extends PanacheEntity {
...
   @Column(name = "acronym")
   private String acronym;
+
+  /**
+   * Validates that the final value in the chain
+   * project.getFunding().getGrantIdentifier().getIdentifier() is not null.
+   *
+   * @return the value of the chain if it's not null.
+   * @return null if any part of the chain is null.
+   */
```
ValentinFutterer: The description needs to be updated, its still like it was before the refactoring.

src/main/java/org/damap/base/domain/Project.java
```diff
@@ -41,4 +41,36 @@ public class Project extends PanacheEntity {
...
+    return this.getFunding().getGrantIdentifier().getIdentifier();
+  }
+
+  /**
+   * Validates that the final value in the chain
+   * project.getFunding().getFunderIdentifier().getIdentifier() is not null.
+   *
+   * @return the value of the chain if it's not null.
+   * @return null if any part of the chain is null.
+   */
```
ValentinFutterer: The description needs to be updated, its still like it was before the refactoring.

ValentinFutterer (approved):
Thanks for doing this - should help mitigate future bugs 👍

PR #337, opened 2025-02-19
Dataset: Add technical resource

### LINKED ISSUE #295: Dataset: Add technical resource
**Description**
From the [RDA-Common-standard](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard?tab=readme-ov-file#properties-in-technical_resource). This field is not yet in DAMAP, eventhough it is in the common standard. Med Uni Graz requires this, therefore we should add it to the dataset data model. Take care to also add this to the madmp export and the template.

Inside the template, add a new column to the produced datasets table with the heading "technical resource (method, device, software)" Since a dataset can ahve multiple datasets, you need to find a way to export it neatly into the template. You can consult Christiane for ideas/to verify your ideas.

**Proposed solution**
Expand the dataset data model, templates and madmp export.

See the [partner frontend issue](https://github.com/tuwien-csd/damap-frontend/issues/403)

GeoffreyKarnbach:
## Description
Feature

#### What does this PR do?
Add technical resource to datasets, in export and maDMP.

#### Change documents Word
[FWFTemplate.docx](https://github.com/user-attachments/files/18863698/FWFTemplate.docx)
[horizonEuropeTemplate.docx](https://github.com/user-attachments/files/18863699/horizonEuropeTemplate.docx)
[scienceEuropeTemplate.docx](https://github.com/user-attachments/files/18863700/scienceEuropeTemplate.docx)

closes GH-295


src/main/java/org/damap/base/domain/TechnicalResource.java
```diff
@@ -0,0 +1,26 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+import org.hibernate.envers.Audited;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ToString
+@Entity
+@Inheritance(strategy = InheritanceType.JOINED)
```
ValentinFutterer: You onyl need the `@Inheritance` annotation when your class has subclasses, so it should be save to remove.

src/main/java/org/damap/base/domain/TechnicalResource.java
```diff
@@ -0,0 +1,26 @@
...
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import jakarta.persistence.*;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+import org.hibernate.envers.Audited;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ToString
```
ValentinFutterer: `@ToString` is superflous, since its included in `@Data`.

src/main/java/org/damap/base/rest/dmp/mapper/DatasetDOMapper.java
```diff
@@ -119,6 +131,17 @@ public Dataset mapDOtoEntity(DatasetDO datasetDO, Dataset dataset, MapperService
     dataset.setType(typeList);
     dataset.setFileFormat(datasetDO.getFileFormat());
 
+    List<TechnicalResource> existingResources = dataset.getTechnicalResources();
```
ValentinFutterer, line 134: I am not 100% sure, but i think if you send a dataset over the API, technical resources can in theory be null, so we should add check here, just to be safe. Or you create a new list and override the old one, which would be cleaner imo.

src/main/java/org/damap/base/rest/madmp/mapper/MaDmpMapper.java
```diff
@@ -241,9 +241,26 @@ public Dataset mapToMaDmp(
...
 
+    dataset.setTechnicalResource(getTechnicalResource(datasetDO));
+
     return dataset;
   }
 
+  private List<TechnicalResource> getTechnicalResource(DatasetDO datasetDO) {
+    List<TechnicalResource> technicalResources = new ArrayList<>();
+    for (TechnicalResourceDO technicalResourceDO : datasetDO.getTechnicalResources()) {
+      TechnicalResource technicalResource = new TechnicalResource();
```
ValentinFutterer, line 252: The if is always true, like we discussed.
Also please add a comment on why you coded this in.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.4.0_1.yaml
```diff
@@ -0,0 +1,84 @@
+databaseChangeLog:
+  - changeSet:
+      id: 22
```
ValentinFutterer: id should be 21

src/main/java/org/damap/base/domain/TechnicalResource.java
GeoffreyKarnbach: done

src/main/java/org/damap/base/domain/TechnicalResource.java
GeoffreyKarnbach: done

src/main/java/org/damap/base/rest/madmp/mapper/MaDmpMapper.java
GeoffreyKarnbach: done

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.4.0_1.yaml
GeoffreyKarnbach: done

src/main/java/org/damap/base/rest/dmp/mapper/DatasetDOMapper.java
GeoffreyKarnbach: done

src/main/java/org/damap/base/domain/TechnicalResource.java
```diff
@@ -0,0 +1,23 @@
...
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@Entity
+@Table(name = "technical_resource")
+@Audited
+public class TechnicalResource extends PanacheEntity {
+
+  @Column(name = "name")
+  private String name;
```
ValentinFutterer: This needs a `@NotBlank` and `@NotNull` annotation imo

src/main/java/org/damap/base/rest/dmp/domain/TechnicalResourceDO.java
```diff
@@ -0,0 +1,16 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import jakarta.validation.constraints.Size;
+import lombok.Data;
+
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class TechnicalResourceDO {
+
+  @Size(max = 255)
```
ValentinFutterer, line 13: This also needs a `@NotBlank` and `@NotNull`

src/main/java/org/damap/base/rest/dmp/domain/TechnicalResourceDO.java
```diff
@@ -0,0 +1,16 @@
...
+  private String name;
+
+  @Size(max = 255)
+  private String description;
```
ValentinFutterer: Descriptions are longer than names,i would use the standard 4000 here, or was there a reason to make it 255

ValentinFutterer:
Seems like some random run collapsed again or something else is wrong with the export to HE.
image
Can you please fix it real quick, thank you :)

ValentinFutterer (approved):
Rdy to be merged

PR #343, opened 2025-03-24
FWF: Add new resource file specifically for the FWF template

### LINKED ISSUE #342: SE and FWF have different texts for [personaldata] placeholder
**Describe the bug**

SE and FWF have different texts to replace the [personaldata] placeholder with, but they both share the `scienceEuropeTemplate.resource` file. FWF should get its own resource file, to solve this bug. (Right now, the bug is in the SE export. Before https://github.com/tuwien-csd/damap-backend/pull/264, it was in FWF)

SE: `personal.avail=In this project, we will process personal data (see section 1a)`
FWF: `personal.avail=In this project, we will process personal data (see section II.1).`

**Is this a regression?**
No.

**Steps to Reproduce**
Choose yes here:

Image

And watch the SE export at section 4 and the FWF at section V.

**Expected behavior**
Export should be correct in both templates.

**Screenshots**

**Environment**
 - Damap version 4.4.0

GeoffreyKarnbach:
## Description
FWF resource file is almost the same, except for `personal.avail`.

closes GH-342

### DIFF
--- a/src/main/java/org/damap/base/conversion/TemplateFileBrokerServiceImpl.java
+++ b/src/main/java/org/damap/base/conversion/TemplateFileBrokerServiceImpl.java
@@ -32,7 +32,7 @@ public class TemplateFileBrokerServiceImpl implements TemplateFileBrokerService 
   /** {@inheritDoc} */
   @Override
   public Properties getFWFTemplateResource() {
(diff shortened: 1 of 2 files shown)


ValentinFutterer (approved):
Looks good 👍

PR #345, opened 2025-03-26
Template: Add cover pages to templates

### LINKED ISSUE #284: Add an Info Sheet/Cover Page for exports to explain colored text
**Description**
Colored text is unclear to users.  An info sheet or cover letter is needed to explain what the color(s) signify

**Describe the solution you'd like**
Create sheet with table explaining colors used in text

**Describe alternatives you've considered**
A table within the existing templates could also be used, but the concern is that users might easily overlook the table.  By using an entire sheet there is a higher chance of the user reading the text

**Additional context**
Feedback from Tomasz' DMP class.

GeoffreyKarnbach:
## Description
Feature

closes GH-284


ValentinFutterer (approved):
Well done adjusting the fonts to the template

PR #352, opened 2025-04-07
DMP Filename handling

### LINKED ISSUE #350: DMP export filenames show "DMP_null_someID" when title fields are null
**Describe the bug**
When exporting DMPs, the filename generation logic in DmpService.getDefaultFileName() doesn't properly handle cases where both the DMP title and project title are null. This results in filenames like "DMP_null_someID" in the download list...

**Is this a regression?**
Yes/No.

**Steps to Reproduce**

**Expected behavior**

**Screenshots**

**Environment**
 - Browser [e.g. firefox, chrome]
 - Damap version [e.g. 2.1.0]

**Additional context**

lpandath:
## Description
The getDefaultFileName() method in DmpService was not handling null titles consistently, potentially resulting in filenames like "DMP_null_someID" for older DMPs.
Test cases for the getDefaultFileName() method were incomplete, missing scenarios for older DMPs with null titles and special characters

#### What does this PR do?
Always prefix filenames with "DMP_" for consistency
Handle null titles by using the DMP ID as a fallback
Added test case 

### Checks

- [ ] Export updated
- [ ] Documentation added
- [x] Tests added
- [ ] E2e tests created

closes GH-350

### DIFF
--- a/src/main/java/org/damap/base/rest/dmp/service/DmpService.java
+++ b/src/main/java/org/damap/base/rest/dmp/service/DmpService.java
@@ -241,27 +241,19 @@ public class DmpService {
         }
       }
 
-      // This condition is met if:
-      // - no project (dmp.getProject() == null).
-      // - a project that contains neither an acronym nor a title
-      // - if dmp.getTitle() is available, we use it directly.
-      if (namePart == null && dmp.getTitle() != null) {
+      if (dmp.getTitle() != null && !dmp.getTitle().trim().isEmpty()) {
         namePart = dmp.getTitle();
         usePrefixAndSuffix = false;
       }
     }
 
     String finalName;
-    if (namePart != null) {
-      if (usePrefixAndSuffix) {
-        finalName = "DMP_" + namePart + "_" + formattedDate;
-      } else {
-        finalName = namePart;
-      }
+    if (namePart != null && !namePart.trim().isEmpty()) {
+      finalName = usePrefixAndSuffix ? "DMP_" + namePart + "_" + formattedDate : namePart;
     } else {
-      // Fallback to default if no namePart could be determined from DMP or Project
       finalName = "DMP_" + id + "_" + formattedDate;
     }
+
     return finalName.replaceAll("[\"',\s]+", "_");
   }
 
--- a/src/main/java/org/damap/base/rest/dmp/service/DmpService.java
+++ b/src/main/java/org/damap/base/rest/dmp/service/DmpService.java
@@ -16,6 +16,7 @@ import org.damap.base.enums.EIdentifierType;
 import org.damap.base.repo.AccessRepo;
 import org.damap.base.repo.DmpRepo;
 import org.damap.base.rest.dmp.domain.*;
+import org.damap.base.rest.dmp.domain.ProjectDO;
 import org.damap.base.rest.dmp.mapper.ContributorDOMapper;
 import org.damap.base.rest.dmp.mapper.DmpDOMapper;
 import org.damap.base.rest.dmp.mapper.DmpListItemDOMapper;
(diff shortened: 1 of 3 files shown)


src/main/java/org/damap/base/rest/dmp/service/DmpService.java
```diff
@@ -233,11 +233,20 @@ public String getDefaultFileName(long id) {
...
-        if (dmp.getTitle() != null) filename = dmp.getTitle();
+        if (dmp.getTitle() != null) {
+          filename = dmp.getTitle();
+        } else {
+          filename = "DMP_" + id + "_" + formatter.format(date);
+        }
       }
+    } else {
+      filename = "DMP_" + id + "_" + formatter.format(date);
     }
```
ValentinFutterer: This code works, but by adding more logic it became very hard to read. For example, there are three instances of:
```
else {
   filename = "DMP_" + id + "_" + formatter.format(date);
 }
```

Since this is the new default value, you could assign `filename = "DMP_" + id + "_" + formatter.format(date);` at the beginning, then you dont need to have all of these else branches. 
The you could turn this into an if else branch:
```
} else {
        if (dmp.getTitle() != null) {
          filename = dmp.getTitle();
```

Another thing of the top of my head would be something like this as a kind of preprocessing to avoid codelines that stretch over multiple lines.:
```
String universityId = dmp.getProject().getUniversityId();
Project universityProject = (universityId != null) ? projectService.read(universityId) : null;
```

I also noticed that when a project exists, but acronym and title are null, it will go to the default name, without even trying dmp.title().

ValentinFutterer:
Good work on the tests 👍

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
lpandath: Makes sense, thanks for your review!

ValentinFutterer:
Seems like a git mishap has occured :)  can you please fix it, then I will continue the review

ValentinFutterer:
There is still something wrong with the git history - look at the changed files, they include changes from the last two patches.
Also, this change should only be one commit.

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
```diff
@@ -212,37 +213,40 @@ private void removeVersions(Dmp dmp) {
...
       }
     }
 
-    filename = filename.replaceAll("[\"',\\s]+", "_");
+    if (dmp.getTitle() != null) {
+      return dmp.getTitle();
+    }
 
-    return filename;
+    return filename.replaceAll("[\"',\\s]+", "_");
```
ValentinFutterer: There is some functionality lost from the refactor - the replaceAll postprocessing step is only done when no if is entered. You should use else ifs, to make the control flow only enter one conditional that sets the filename variable (instead of instantly returning). The last line here will then execute as normal and do the replaceAll.

ValentinFutterer: Or, another idea, you scrap the global filename variable and do the postprocessing inline with the return statements, to improve readability of the if else constructs

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
```diff
@@ -212,37 +213,56 @@ private void removeVersions(Dmp dmp) {
...
+    if (namePart != null) {
+      if (usePrefixAndSuffix) {
+        finalName = "DMP_" + namePart + "_" + formattedDate;
+      } else {
+        finalName = namePart;
+      }
+    } else {
+      // Fallback to default if no namePart could be determined from DMP or Project
+      finalName = "DMP_" + id + "_" + formattedDate;
+    }
```
ValentinFutterer: This part is weird. `usePrefixAndSuffix` can only be false if `namePart = null`. But `usePrefixAndSuffix` only matters in a block that is entered when `namePart != null`...

Could you please fix this? Otherwise, I will do it myself. I have already spent more time reviewing than it would take me to solve this issue myself.

lpandath: @ValentinFutterer thank you, i will check why is weird. What is your opinion on this as reviwer? otherwise i will come up with different solution

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
```diff
@@ -212,37 +213,48 @@ private void removeVersions(Dmp dmp) {
...
       }
     }
 
-    filename = filename.replaceAll("[\"',\\s]+", "_");
+    String finalName;
+    if (namePart != null && !namePart.trim().isEmpty()) {
+      finalName = usePrefixAndSuffix ? "DMP_" + namePart + "_" + formattedDate : namePart;
+    } else {
+      finalName = "DMP_" + id + "_" + formattedDate;
+    }
```
ValentinFutterer: Problem with this is, that namePart is now always overriden. The original solution was way to o complicated to be easily understood. I am not a big fan of flags like `useprefixAndSuffix`, especially when its set at only one place - its just one more thing to track. The original intention was to handle a weird exception to the naming algorithm, but why use a flag at the first place. Something like this would be easier too read imo:

```
String finalName;
    if (namePart == null) {
      if (dmp != null && dmp.getTitle() != null) {
        finalName = dmp.getTitle();
      } else {
        finalName = "DMP_" + id + "_" + formattedDate;
      }
    } else {
      finalName = "DMP_" + namePart + "_" + formattedDate;
    }
```

But thats still a lot of code and if and elses. To make it even cleaner, we could just treat `dmp.getTitle()` like a normal namePart with prefix and suffix. To do this, we can just replace all empty spaces with underscores. What do you think of that idea, would that work for you and Graz?

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
ValentinFutterer: Its weird to read because when namepart is null, its set to the dmp title and the `usePrefixAndSuffix`. And then in the next block, where it uses `usePrefixAndSuffix` as a condition, it checks if `namePart != null`. And i missed the logical connection where namePart was set to the title, so its not null and the block is entered. You need to make a lot of logical jumps to understand it.

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
lpandath: Thank you for your thorough review! Let me address this:
the current implementation has some merits - it provides a clear separation between name gathering and name formatting, with the namePart variable acting as a single point of truth for the filename's base. The usePrefixAndSuffix flag explicitly documents our rule where DMP titles are treated differently.

However, you are right about the readability issues. The code evolved from a simpler version into a complex structure with nested if-else statements, which is exactly why I created this issue and attempted to refactor it.

I'll restructure the code to remove the usePrefixAndSuffix flag and handle each case separately with more clear methods where this will make the priority order of name selection while maintaining the same rules but with cleaner code.

p.s `squash` after review.

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
lpandath: replied in the other comment.

ValentinFutterer:
It works fine now and the code is way more readable 👍 
The banch is out of date though, can you rebase and squash, then i can merge.

lpandath:
> It works fine now and the code is way more readable 👍 The banch is out of date though, can you rebase and squash, then i can merge.

you can do that now, thanks!

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
```diff
@@ -212,37 +212,61 @@ private void removeVersions(Dmp dmp) {
...
+    return formatDefaultName(id, formattedDate);
+  }
+
+  private boolean hasDmpTitle(Dmp dmp) {
+    return dmp.getTitle() != null && !dmp.getTitle().trim().isEmpty();
+  }
+
+  private String extractProjectName(Dmp dmp) {
+    if (dmp.getProject() == null) {
+      return null;
```
bitmotivator, line 248: In the old code this would have created a filename based on `My Data Management Plan`. The new code seems to create an artificial filename based on the DMP date. Is this intentional?

bitmotivator (approved):
The old code took me a while to wrap my head around, but the new code looks fine. @ValentinFutterer please take a second look and then I'm happy to merge this.

src/main/java/org/damap/base/rest/dmp/service/DmpService.java
lpandath: The old implementation had issues with older DMPs, were resulting in filenames like "DMP_null_someID" for older DMPs. The "My Data Management Plan" filename is now only used as the ultimate fallback when the DMP doesn't exist in the database. For existing DMPs, users get more detailed filenames based on their actual DMP title or project information with date...

bitmotivator: I have no strong opinion on file names, maybe @ValentinFutterer does, I just wanted to note that the logic seems to be different.

ValentinFutterer: I also dont have strong opinions, these are edge cases that shouldnt even come up in normal deployments and users dont pay that much attention to these filenames anyways. I just want to merge this PR already 😅

PR #354, opened 2025-04-08
Templates: remove placeholders when exporting without project

### LINKED ISSUE #351: Exporting with no project leads to placeholders staying in the document
**Describe the bug**
Exporting without a project leads to placeholders being left in the docuemnt, while it shiuld be relaced with an empty string to remove placeholders. (Title stying as a placeholder is fine)

Image

**Is this a regression?**
Yes.

**Expected behavior**
Placeholders should not appear in the exported document.

**Environment**
 - Damap version 4.5.0

BlueTechie:
## Description

Bugfix
#### What does this PR do?

Removed 5 placeholders when exporting without project (startdate, enddate, funderid, grantid and projectid)
Minor refactoring

closes GH-351

### DIFF
--- a/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -67,6 +67,11 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
   public void titlePage() {
     Project project = dmp.getProject();
     if (project == null) {
+      addReplacement(replacements, "[startdate]", "");
+      addReplacement(replacements, "[enddate]", "");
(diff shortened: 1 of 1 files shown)


ValentinFutterer:
Simple enough fix, works well 👍 

Minor note: When starting a commit message in uppercase, I would also start the word after the : in uppercase, so "Template: Remove..." in this case. You can read [this article](https://cbea.ms/git-commit/#capitalize) on commit messages if you are interested.

I would also remark that "Templates" is not correct here, since the changes did not occur in the templates, but in the export. You could just write "Export" instead.

BlueTechie:
Thanks, I fixed the commit message

ValentinFutterer (approved):
Thanks for taking time to refactor 😄

PR #375, opened 2025-04-30
Export: Fix incompatible column number warning

### LINKED ISSUE #371: Export: Shows weird error messages
**Describe the bug**
When you create a new DMP, choose a project and then export, these error messages are displayed:

Image

But apparently, everything works fine? This needs to be checked out.

**Is this a regression?**
Yes.

**Expected behavior**
No error messages and no wrong behaviour.

**Environment**
 - Damap version 4.5.2

BlueTechie:
## Description

Bugfix
#### What does this PR do?

Fixes the warnings of the form "Number of provided values for table row does not match number of row cells. num provided/num row cells: 7/6"

This had 2 causes:
- For the 1st table (_produced datasets_): the description column is always passed along to fill the table even though only the FWF template actually uses such a column. This does not break functionality (but still leads to the warning) since the minimum of numCells and numContent is taken  (see line 173 in _AbstractTemplateExportFunctions_) as the number of columns to fill, i.e. the minimum of expected and provided column numbers. At first, I tried fixing it with overrides of the composeTableDataAccess method from the abstract _AbstractTemplateExportScienceEuropeComponents_ class in the export classes but this is cumbersome. This solution using instanceof is more elegant, as it simply checks if the _ExportFWFTemplate_ class is instantiated and if so passes along 7 columns instead of 6 (both if there are datasets or if there are none).

- For the 3rd table (_Data publication and access conditions_): if no dataset is specified, an array of 7 empty strings was created but it should  be 6 since that table has 6 columns in all 3 templates.
#### Breaking changes

closes GH-371

### DIFF
--- a/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -1143,8 +1143,8 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
       }
       xwpfTable.removeRow(xwpfTable.getRows().size() - 1);
     } else {
-      // clean row
-      ArrayList<String> emptyContent = new ArrayList<>(Arrays.asList("", "", "", "", "", "", ""));
+      // clean row, the first FWF template table has the additional "description" column
+      ArrayList<String> emptyContent = new ArrayList<>(Collections.nCopies(7, ""));
       insertTableCells(
           xwpfTable, xwpfTable.getRows().get(xwpfTable.getRows().size() - 1), emptyContent);
     }
@@ -1376,8 +1376,7 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
       xwpfTable.removeRow(xwpfTable.getRows().size() - 1);
     } else {
       // clean row
-      ArrayList<String> emptyContent =
-          new ArrayList<String>(Arrays.asList("", "", "", "", "", "", ""));
+      ArrayList<String> emptyContent = new ArrayList<String>(Arrays.asList("", "", "", "", "", ""));
       insertTableCells(
           xwpfTable, xwpfTable.getRows().get(xwpfTable.getRows().size() - 1), emptyContent);
     }
--- a/src/main/java/org/damap/base/conversion/AbstractTemplateExportFunctions.java
+++ b/src/main/java/org/damap/base/conversion/AbstractTemplateExportFunctions.java
@@ -156,19 +156,6 @@ public abstract class AbstractTemplateExportFunctions {
     int numCells = cells.size();
     int numContent = cellContent.size();
 
-    // Log warning if numCells and numContent does not match
-    if (numCells != numContent) {
-      log.warn(
-          String.format(
-              "Table %s\n"
-                  + "Number of provided values for table row does not match number of row cells. num provided/num row cells: %2d/%2d",
-              String.join(
-                  " | ",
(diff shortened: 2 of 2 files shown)


ValentinFutterer:
I am not opposed to the solution, but it doesnt feel good either :)

Like we can merge it, but it makes the code more complicated. Maybe we can remove the code that prints the warnings altogether? What does it protect against exactly, how realistic are the chances that these warnings will catch a bug before it happens? Maybe also check the commit where this was introduced for more context.

Nice PR description btw 👍

BlueTechie:
I would recommend not removing the code generating the warning messages since the templates have changed in the past (see for instance https://github.com/tuwien-csd/damap-backend/pull/258) and/or it could be the case that the values passed to fill out the tables change (as was the case in https://github.com/tuwien-csd/damap-backend/pull/234 where the warning code was introduced) and in these cases debugging would be more difficult without the warnings.

ValentinFutterer:
Your changes broke something when exporting to FWF.
image

I am still not opposed to removing the warning code tbh :)

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -1122,14 +1122,17 @@ public void composeTableNewDatasets(XWPFTable xwpfTable) {
           docVar.add("no");
         }
 
-        docVar.add(Optional.ofNullable(dataset.getDescription()).orElse(""));
+        if (this instanceof ExportFWFTemplate) {
```
bitmotivator: Downcasts usually throw up a bit of a red flag. Could you please add a comment to the code why this is necessary and what happens in the other cases.

bitmotivator (dismissed):
Small nitpick

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
ValentinFutterer: At this point I would just remove the warning code, seems it produces more work than i saves :)

BlueTechie: I removed the warning code per @ValentinFutterer 's suggestion, which also removes the need for the hacky solution with "instanceof" which was done to avoid this warning. Since the code picks the minimum of the expected and provided number of columns anyways, the solution to always pass on 7 columns should be valid and does not merit an extra warning each time.

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -1361,8 +1361,7 @@ public void composeTableDatasetPublication(XWPFTable xwpfTable) {
       xwpfTable.removeRow(xwpfTable.getRows().size() - 1);
     } else {
       // clean row
-      ArrayList<String> emptyContent =
-          new ArrayList<String>(Arrays.asList("", "", "", "", "", "", ""));
+      ArrayList<String> emptyContent = new ArrayList<String>(Arrays.asList("", "", "", "", "", ""));
```
bitmotivator, line 1379: There are only 6 instead of 7 fields here now. Is this intentional? Also, Collections.nCopies looks kind of neat, could be used here.

ValentinFutterer: The 7 fields where a bug, since the table only has 6 columns, its mentioned in the PR description :)

ValentinFutterer (approved):
I am happy with just removing the code 👍

PR #376, opened 2025-05-13
DB: Generalize to also support Oracle DBs

BlueTechie:
## Description

Feature
#### What does this PR do?

_Old PR description:_
~~Test migration from Postgres to Oracle (with fixed comments).~~

~~This is only a test to see if switching to oracle works and should not be merged to next (apologies for the confusion)~~

_New PR description:_
Added support for Oracle DB both for the CI pipeline and dev environments by generalizing to not only support postgres databases.

closes GH-


Dockerfile
```diff
@@ -16,7 +16,10 @@ LABEL maintainer="zeno.casellato@tuwien.ac.at" \
         org.label-schema.docker.cmd="docker run -d -p 8080:8080 damap"
 
 ARG BUILD_HOME=/home/app
-ARG BUILD_PROFILE=postgres
+# FOR POSTGRES USE:
+#ARG BUILD_PROFILE=postgres
+# FOR ORACLE USE:
+ARG BUILD_PROFILE=oracle
```
ghost: What does this argument do? Shouldn't DAMAP be DB-agnostic?

docker/docker-compose-services.yaml
```diff
@@ -1,36 +1,36 @@
...
+    environment:
+      ORACLE_PASSWORD: pw4damap
+      APP_USER: damap
+      APP_USER_PASSWORD: pw4damap
+    healthcheck:
+      test: [ "CMD-SHELL", "healthcheck.sh" ]
+      interval: 30s
+      timeout: 15s
+      retries: 5
+
```
ghost: This change essentially removes Postgres from the DAMAP docker-compose altogether, which is not ideal. Very few people want to run with Oracle by default in a local test instance.

It would be better to create separate docker compose files for running with Postgres and Oracle and move the common parts to an included file.

ghost (dismissed):
Hey @BlueTechie this is in draft but still moved to the review column... Is it actually ready to review? At any rate, here's a provisional review. We definitely need to test for both Postgres and Oracle, but doing so shouldn't require code changes each time. We can work with include files, environment variables, etc and test both at the same time.

BlueTechie:
Apologies for the confusion! This draft PR was only meant as a test to see if switching to oracle works as expected and was not intended to be merged to next :)

BlueTechie:
Updated the PR description and title, this PR is now ready for review.

SotosTsepe (changes requested):
It seems that the backend still fails to start. I could only succeed by hardcoding `driver: oracle.jdbc.driver.OracleDriver`. Somehow the environment variable is not passed from the compose file.. Investigating further.

bitmotivator:
Hey @SotosTsepe @BlueTechie  I dug into this a little more and found this while enabling trace logging in Quarkus:

```
damap-be-1      | 2025-05-26 20:01:24,180 WARN  [io.qua.run.con.ConfigRecorder] (main) Build time property cannot be changed at runtime:
damap-be-1      |  - quarkus.datasource.jdbc.driver is set to 'oracle.jdbc.driver.OracleDriver' but it is build time fixed to 'null'. Did you change the property quarkus.datasource.jdbc.driver after building the application?
```

Looking at the [Quarkus data sources documentation](https://quarkus.io/guides/datasource#datasource-reference), it seems to me that `quarkus.datasources.db-kind` is a build-time configuration option and cannot be changed at runtime. This is... less than ideal and I should have checked earlier.

At any rate, the following options are still open:

1. Simply revert back to using build profiles and make the CI work that way. Also document that you can NOT change the DB kind at runtime.
2. Build and ship a separate container image for Oracle as a first-class citizen. This is not ideal as it will need extra tooling.
3. Set up separate data sources for Oracle and PostgreSQL in the application.yaml and switch using profiles. (This is hack and probably not very maintainable.)
4. Find a way to make Quarkus do what we want. ([This StackOverflow question](https://stackoverflow.com/questions/66426953/configure-datasource-type-at-runtime) seems to suggest that [re-augmentation may be the answer](https://quarkus.io/guides/reaugmentation).)

I think for the purposes of this PR, option 1 is probably a good solution. Option 4 seems  to be also relatively painless. I'm happy to contribute whichever solution you feel comfortable with to this PR.

> [!TIP]
> While debugging this I learned that you can override application.yaml settings simply by specifying the env variable in Quarkus. So `QUARKUS_LOG_LEVEL` is equal to `quarkus.log.level` in the config. You can run Quarkus with trace logging by specifying `QUARKUS_LOG_LEVEL: ALL` in the `docker-compose.oracle.yaml` file. See the [Quarkus Configuration Reference Guide](https://quarkus.io/guides/config-reference#environment-variables) for details.

SotosTsepe:
Hey @bitmotivator, thanks for the input! I had similar findings through the docs yesterday and your points are valid. One thing I tried out and worked is setting `ENV QUARKUS_DATASOURCE_JDBC_DRIVER="oracle.jdbc.driver.OracleDriver"` in the build (1st) stage of the Dockerfile.

Since we have 2 compose files, we could pass build arguments from the file.

We could have the following in the Oracle compose file:
```
      args:
        BUILD_PROFILE: oracle
        DB_KIND: oracle
        JDBC_DRIVER: oracle.jdbc.driver.OracleDriver
 ```
 
 and pass the args dynamically in the Dockerfile, which works:
```
RUN mvn -B clean package -DskipTests \
    -Dquarkus.datasource.jdbc.driver=${JDBC_DRIVER} \
    -Dquarkus.datasource.db-kind=${DB_KIND} \
    -Dquarkus.profile=${BUILD_PROFILE}
```
 
Does this solution make sense? I think it follows the logic of point 2, correct? If not suitable, I will start experimenting with points 1 and 4.

**Comment by bitmotivator** [comment]
@SotosTsepe yes, this would be solution 2 indeed. Given the administrative overhead of releasing multiple images (one for each DB vendor) trying solution 4 would probably be worth at least a little investigation as it is the most flexible for the users as well as for the devs.

**Comment by SotosTsepe** [comment]
Remaining checklist before merging:

- [x] Reaugment only if environment variables have changed.
- [x] Pipeline tests for Maven build are still running on dev profile (which means h2 and not PostgreSQL/Oracle). Address this.
- [x] Tests must succeed
- [x] Sanitize documentation

File: .github/workflows/ci.yml [hunk 3]
```diff
@@ -0,0 +1,126 @@
...
+  # The H2 database is used by default and does not require a Docker container.
+  test-maven:
+    name: test-maven (${{ matrix.database }})
+    runs-on: ubuntu-latest
+    needs: test-docker
+    strategy:
+      matrix:
+        include:
+          # The H2 database gets its settings from the dev profile.
+          - database: h2
```
bitmotivator, line 59: The H2 database engine has some quirks with Hibernate if I remember correctly. If it works with Oracle and PosgreSQL, but not with H2 we can proceed.

bitmotivator (approved):
Approving in the anticipation that the issues will be fixed, the code looks good. Please track down if the people writing the failing code are still around in order to speed up getting this merged.

.github/workflows/ci.yml
SotosTsepe: It worked before, when I supplied these:
```
  url: jdbc:h2:mem:testdb
  db_kind: h2
  driver: org.h2.Driver
```

If I run this command locally, it works properly, so I assume the test profile is activated automatically and retrieves the settings from `application.yml`. However, in the pipeline, this doesn't seem to be the case.

SotosTsepe (approved):
PR [checklist](https://github.com/damap-org/damap-backend/pull/376#issuecomment-2932065189) completed.

ghost (approved):

PR #355, opened 2025-04-09
Storage: Add boolean to indicate if an external storage is managed

GeoffreyKarnbach:
## Description
Feature

closes GH-349

PR #377, opened 2025-05-20
Gk/374 api key

### LINKED ISSUE #374: API KEY
Frontend Issue https://github.com/tuwien-csd/damap-frontend/issues/450
User should be able to generate an API-Key so machines can connect to our APIs using this key.
There should be some way to manage keys in a secure way (e.g. only show them once or  expration dates)

GeoffreyKarnbach:
## Description
Feature

closes GH-374

### DIFF
--- a/src/main/java/org/damap/base/domain/ApiKey.java
+++ b/src/main/java/org/damap/base/domain/ApiKey.java
@@ -1,0 +1,29 @@
+package org.damap.base.domain;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.Table;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ToString(exclude = "user")
+@Entity
+@Table(name = "apikey")
+public class ApiKey extends PanacheEntity {
+
+  @Column private String name;
+
+  @Column(nullable = false)
+  private String value;
+
+  @ManyToOne(fetch = FetchType.EAGER)
+  @JoinColumn(name = "user_id", nullable = false)
+  private User user;
+}
--- a/src/main/java/org/damap/base/domain/User.java
+++ b/src/main/java/org/damap/base/domain/User.java
@@ -1,0 +1,31 @@
+package org.damap.base.domain;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import jakarta.persistence.*;
+import java.util.List;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ToString(exclude = {"apiKeys", "roles"})
+@Entity
+@Table(name = "user_account")
+public class User extends PanacheEntity {
+
+  @Column(name = "user_id", nullable = false, unique = true)
+  private String userId;
+
+  @Column(name = "display_name")
+  private String displayName;
+
+  @Column(nullable = false, unique = true)
+  private String username;
+
+  @OneToMany(mappedBy = "user")
+  private List<ApiKey> apiKeys;
+
+  @OneToMany(mappedBy = "user", fetch = FetchType.EAGER)
+  private List<UserRole> roles;
+}
--- a/src/main/java/org/damap/base/rest/AccountResource.java
+++ b/src/main/java/org/damap/base/rest/AccountResource.java
@@ -1,0 +1,82 @@
+package org.damap.base.rest;
+
+import io.quarkus.security.Authenticated;
+import jakarta.annotation.security.RolesAllowed;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.*;
+import jakarta.ws.rs.core.MediaType;
+import java.util.List;
+import java.util.Set;
+import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.rest.account.domain.ApiKeyDO;
+import org.damap.base.rest.account.domain.UserPageDO;
+import org.damap.base.rest.account.service.AccountService;
+import org.damap.base.security.SecurityService;
+
+@Path("/api/account")
+@Produces(MediaType.APPLICATION_JSON)
+@Authenticated
+@JBossLog
+public class AccountResource {
+
+  @Inject AccountService accountService;
+
+  @Inject SecurityService securityService;
+
+  @GET
+  @Path("/api-keys")
+  public List<String> getApiKeys() {
+    log.info("GET /account/api-keys");
+    String userId = securityService.getUserId();
+    return accountService.getApiKeys(userId);
+  }
+
+  @POST
+  @Path("/api-keys")
+  public ApiKeyDO createApiKey(@QueryParam("keyName") String keyName) {
+    log.info("POST /account/api-keys");
+    String username = securityService.getUserId();
+    String displayName = securityService.getDisplayName();
+    String userId = securityService.getUserId();
+    Set<String> roles = securityService.getUserRoles();
+    return accountService.createApiKey(keyName, username, displayName, userId, roles);
+  }
+
+  @POST
+  @Path("/api-keys/reset-all")
+  @RolesAllowed("Damap Admin")
+  public boolean resetAllApiKeys(@QueryParam("userId") String userId) {
+    log.info("POST /account/api-keys/reset-all for userId: " + userId);
+    return accountService.resetAllApiKeys(userId);
+  }
+
+  @POST
+  @Path("/refresh-user-role")
+  public void refreshUserApiKeyRole() {
+    log.info("POST /account/refresh-user-role");
+    String username = securityService.getUserId();
+    String displayName = securityService.getDisplayName();
+    String userId = securityService.getUserId();
+    Set<String> roles = securityService.getUserRoles();
+    accountService.refreshUserApiKeyRole(username, displayName, userId, roles);
+  }
+
+  @DELETE
+  @Path("/api-keys/{key}")
+  public void deleteApiKey(@PathParam("key") String key) {
+    log.info("DELETE /account/api-keys/" + key);
+    String userId = securityService.getUserId();
+    accountService.deleteApiKey(key, userId);
+  }
+
+  @GET
+  @Path("/users")
+  @RolesAllowed("Damap Admin")
+  public UserPageDO getAllUsers(
+      @QueryParam("page") @DefaultValue("0") int page,
+      @QueryParam("size") @DefaultValue("10") int size) {
+    log.info("GET /account/users");
+    int offset = page * size;
+    return accountService.getAllUsers(offset, size);
+  }
+}
--- a/src/main/java/org/damap/base/rest/account/service/AccountService.java
+++ b/src/main/java/org/damap/base/rest/account/service/AccountService.java
@@ -1,0 +1,163 @@
+package org.damap.base.rest.account.service;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import jakarta.transaction.Transactional;
+import jakarta.ws.rs.ForbiddenException;
+import jakarta.ws.rs.NotFoundException;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.domain.ApiKey;
+import org.damap.base.domain.User;
+import org.damap.base.domain.UserRole;
+import org.damap.base.repo.ApiKeyRepo;
+import org.damap.base.repo.UserAccountRepo;
+import org.damap.base.rest.account.domain.ApiKeyDO;
+import org.damap.base.rest.account.domain.UserDO;
+import org.damap.base.rest.account.domain.UserPageDO;
+
+@ApplicationScoped
+@JBossLog
+public class AccountService {
+
+  @Inject private UserAccountRepo userAccountRepo;
+
+  @Inject private ApiKeyRepo apiKeyRepo;
+
+  public List<String> getApiKeys(String userId) {
+    log.info("GET /account/api-keys");
+    List<ApiKey> apiKeys = apiKeyRepo.findByUserId(userId);
+    return apiKeys.stream().map(ApiKey::getName).toList();
+  }
+
+  @Transactional
+  public ApiKeyDO createApiKey(
+      String keyName, String username, String displayName, String userId, Set<String> roles) {
+    log.info("POST /account/api-keys");
+
+    if (keyName == null || keyName.isEmpty()) {
+      throw new IllegalArgumentException("API key name cannot be null or empty");
+    }
+
+    Optional<User> optionalUser = userAccountRepo.findByUserId(userId);
+    User existingUser;
+
+    if (optionalUser.isPresent()) {
+      existingUser = optionalUser.get();
+    } else {
+      existingUser = new User();
+      existingUser.setUsername(username);
+      existingUser.setDisplayName(displayName);
+      existingUser.setUserId(userId);
+      existingUser.persist();
+
+      for (String role : roles) {
+        UserRole userRole = new UserRole();
+        userRole.setUser(existingUser);
+        userRole.setRole(role);
+        userRole.persist();
+      }
+    }
+
+    Optional<ApiKey> optionalApiKey = apiKeyRepo.findByKey(keyName);
+    if (optionalApiKey.isPresent()
+        && Objects.equals(optionalApiKey.get().getUser().getUserId(), existingUser.getUserId())) {
+      throw new IllegalArgumentException("API key already exists for this user with the same name");
+    }
+
+    ApiKey apiKey = new ApiKey();
+    apiKey.setName(keyName);
+    apiKey.setValue(java.util.UUID.randomUUID().toString());
+    apiKey.setUser(existingUser);
+    apiKey.persist();
+
+    return ApiKeyDO.builder().name(apiKey.getName()).value(apiKey.getValue()).build();
+  }
+
+  @Transactional
+  public void refreshUserApiKeyRole(
+      String username, String displayName, String userId, Set<String> roles) {
+    log.info("POST /account/refresh-user-role");
+    Optional<User> optionalUser = userAccountRepo.findByUserId(userId);
+    if (optionalUser.isPresent()) {
+      User existingUser = optionalUser.get();
+      existingUser.setUsername(username);
+      existingUser.setDisplayName(displayName);
+      existingUser.setRoles(
+          roles.stream()
+              .map(
+                  role -> {
+                    UserRole userRole = new UserRole();
+                    userRole.setUser(existingUser);
+                    userRole.setRole(role);
+                    return userRole;
+                  })
+              .toList());
+      existingUser.persist();
+    } else {
+      throw new NotFoundException("User not found");
+    }
+  }
+
+  @Transactional
+  public void deleteApiKey(String key, String userId) {
+    log.info("DELETE /account/api-keys/" + key);
+
+    if (key == null || key.isEmpty()) {
+      throw new IllegalArgumentException("API key cannot be null or empty");
+    }
+
+    Optional<ApiKey> optionalApiKey = apiKeyRepo.findByKey(key);
+    if (optionalApiKey.isPresent()) {
+      ApiKey apiKey = optionalApiKey.get();
+      if (Objects.equals(apiKey.getUser().getUserId(), userId)) {
+        apiKeyRepo.delete(apiKey);
+      } else {
+        throw new ForbiddenException("API key does not belong to the user");
+      }
+    } else {
+      throw new NotFoundException("API key not found");
+    }
+  }
+
+  public UserPageDO getAllUsers(int offset, int limit) {
+    log.info("GET /account/users page: " + offset + ", size: " + limit);
+    List<User> allUsers = userAccountRepo.findAllUsersPage(offset, limit);
+
+    List<UserDO> users =
+        allUsers.stream()
+            .map(
+                user ->
+                    UserDO.builder()
+                        .userId(user.getUserId())
+                        .displayName(user.getDisplayName())
+                        .username(user.getUsername())
+                        .roles(user.getRoles().stream().map(UserRole::getRole).toList())
+                        .apiKeys(user.getApiKeys().stream().map(ApiKey::getName).toList())
+                        .build())
+            .toList();
+
+    int totalElements = (int) userAccountRepo.count();
+    int totalPages = (int) Math.ceil((double) totalElements / limit);
+    return UserPageDO.builder()
+        .users(users)
+        .totalElements(totalElements)
+        .totalPages(totalPages)
+        .size(limit)
+        .number(offset / limit + 1)
+        .build();
+  }
+
+  @Transactional
+  public boolean resetAllApiKeys(String userId) {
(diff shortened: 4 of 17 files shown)


src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+public class AccountService {
+
+  @Inject private UserAccountRepo userAccountRepo;
+
+  @Inject private ApiKeyRepo apiKeyRepo;
+
+  public List<String> getApiKeys(String userId) {
+    log.info("GET /account/api-keys");
+    List<ApiKey> apiKeys = apiKeyRepo.findByUserId(userId);
+    return apiKeys.stream().map(ApiKey::getName).toList();
```
bitmotivator, line 33: As far as I can tell from the code, this returns the API key name as well as the secret part. API keys should _never_ be returned after they are created. This is because if one of the user's credentials is compromised, you don't want to compromise all of them at once.

Technically, we'd need two DO's, one with the "value" and one without. And yes, this absolutely makes moot the point of calling it a Domain Object, but hey, our "Domain Objects" are actually HTTP DTOs anyway, so that ship has sailed a long long time ago.

bitmotivator (dismissed):
One security-related nit-pick that I'm also happy to address in a follow-up PR. @BlueTechie can I get your quick opinion here? Otherwise, ship it!

src/main/java/org/damap/base/rest/account/service/AccountService.java
GeoffreyKarnbach: I am unsure, I only return the API Key Name and not the secret (allthough it would be available in the Service, but never returned to the user)?

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.6.0_1.yaml
```diff
@@ -0,0 +1,97 @@
...
+              startValue: 1
+              incrementBy: 1
+          - createSequence:
+              sequenceName: apikey_seq
+              startValue: 1
+              incrementBy: 1
+          - createSequence:
+              sequenceName: user_roles_seq
+              startValue: 1
+              incrementBy: 1
```
ValentinFutterer, line 17: Sequences should have an `incrementBy` value of 50, for more info see this: https://github.com/hibernate/hibernate-orm/blob/6.0/migration-guide.adoc#defaults-for-implicit-sequence-generators

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.6.0_1.yaml
```diff
@@ -0,0 +1,97 @@
+databaseChangeLog:
+  - changeSet:
+        id: 22-0
```
ValentinFutterer, line 3: Why is this changeset called 22-0? There is no 22-1.
Also the highest id right now is 23, so this should have 24. 
Careful, https://github.com/damap-org/damap-backend/pull/355 has the same problem. There needs to be some coordination happening, otherwise we will have conflicts.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.x.yaml
```diff
@@ -26,4 +26,6 @@ databaseChangeLog:
   - include:
       file: org/damap/base/db/changeLog-4.x/changeLog-4.5.2_1.yaml
   - include:
-      file: org/damap/base/db/changeLog-4.x/changeLog-4.5.2_2.yaml
\ No newline at end of file
+      file: org/damap/base/db/changeLog-4.x/changeLog-4.5.2_2.yaml
+  - include:
+      file: org/damap/base/db/changeLog-4.x/changeLog-4.6.0_1.yaml
```
ValentinFutterer, line 31: This file should be named after the current version. 
careful with naming, https://github.com/damap-org/damap-backend/pull/355 has the same problem.

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+
+    Optional<ApiKey> optionalApiKey = apiKeyRepo.findByKey(keyName);
+    if (optionalApiKey.isPresent()
+        && Objects.equals(optionalApiKey.get().getUser().getUserId(), existingUser.getUserId())) {
+      throw new IllegalArgumentException("API key already exists for this user with the same name");
+    }
+
+    ApiKey apiKey = new ApiKey();
+    apiKey.setName(keyName);
+    apiKey.setValue(java.util.UUID.randomUUID().toString());
```
BlueTechie, line 73: Api key values should not be generated with a UUID - see [this article](https://hackernoon.com/never-rely-on-uuid-for-authentication-generation-vulnerabilities-and-best-practices) and also section 6 of the [docs](https://www.rfc-editor.org/rfc/rfc4122#page-16) states the following: _Do not assume that UUIDs are hard to guess; they should not be used as security capabilities (identifiers whose mere possession grants access), for example._

Instead, we should use a secure random generator (for instance _SecureRandom_).

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+
+@ApplicationScoped
+@JBossLog
+public class AccountService {
+
+  @Inject private UserAccountRepo userAccountRepo;
+
+  @Inject private ApiKeyRepo apiKeyRepo;
+
+  public List<String> getApiKeys(String userId) {
```
BlueTechie, line 30: I believe blindly accepting userIds without checking for proper authorization could be a security risk, no? Injecting the caller identity and then using it to check for authorization could perhaps be more secure. (this comment applies to each method where userId is passed)

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+    apiKey.setName(keyName);
+    apiKey.setValue(java.util.UUID.randomUUID().toString());
+    apiKey.setUser(existingUser);
+    apiKey.persist();
+
+    return ApiKeyDO.builder().name(apiKey.getName()).value(apiKey.getValue()).build();
+  }
+
+  @Transactional
+  public void refreshUserApiKeyRole(
```
BlueTechie, line 81: I believe access to this method should be restricted (for instance with `@RolesAllowed("admin")`). Otherwise the caller can pass arbitrary roles -> privilege escalation.

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+    return apiKeys.stream().map(ApiKey::getName).toList();
+  }
+
+  @Transactional
+  public ApiKeyDO createApiKey(
+      String keyName, String username, String displayName, String userId, Set<String> roles) {
+    log.info("POST /account/api-keys");
+
+    if (keyName == null || keyName.isEmpty()) {
+      throw new IllegalArgumentException("API key name cannot be null or empty");
```
BlueTechie, line 42: Shouldn't there be more input validation (for example to check the length of the username etc.)?

src/main/java/org/damap/base/security/ApiKeyAuthenticationMechanism.java
```diff
@@ -0,0 +1,65 @@
...
+            em -> {
+              String header = context.request().getHeader("Authorization");
+
+              if (header != null && header.startsWith("ApiKey ")) {
+                String apiKey = header.substring("ApiKey ".length());
+                context
+                    .vertx()
+                    .executeBlocking(
+                        () -> {
+                          Optional<User> user = apiKeyRepo.findUserByApiKeyValue(apiKey);
```
BlueTechie, line 37: Api keys should not be stored in plaintext (instead store the secure hash of the key, for instance SHA-256). Also shouldn't there be a rate limit or brute force protection in place? Otherwise an attacker could in theory spam the endpoint either to deny service or brute force until a valid API key is found.

src/main/java/org/damap/base/security/ApiKeyIdentityProvider.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @Inject ApiKeyRepo apiKeyRepo;
+
+  @Override
+  public Uni<SecurityIdentity> authenticate(
+      ApiKeyAuthenticationRequest request, AuthenticationRequestContext context) {
+    String apiKey = request.getApiKey();
+
+    Optional<User> userOptional = apiKeyRepo.findUserByApiKeyValue(apiKey);
```
BlueTechie, line 32: A DB compromise would result in all API keys getting leaked (see my other comment). It would be more secure to store the hash value of the API key.

BlueTechie (commented):
Review summary: I have identified some potential security concerns, such as storing the API keys in plaintext and generating the value with a random UUID instead of a PRG. However, @bitmotivator is more qualified to judge my findings.

src/main/java/org/damap/base/rest/account/service/AccountService.java
bitmotivator: Nice catch!

src/main/java/org/damap/base/rest/account/service/AccountService.java
bitmotivator: @BlueTechie does this work in a service level (i.e. not in a resource) in Quarkus?

src/main/java/org/damap/base/security/ApiKeyIdentityProvider.java
bitmotivator: Holy macaroni, @BlueTechie you are right! Not only should the API keys be hashed, it should be done so using a proper password hashing that includes multiple rounds and salt (e.g. PDKDF2). I can't believe I missed this.

src/main/java/org/damap/base/security/ApiKeyAuthenticationMechanism.java
bitmotivator: See my comment below, I can't believe how I could have missed this. However, SHA-256 is not sufficient because it doesn't automatically include salting, so it would be susceptible to a rainbow table attack. Multiple rounds of SHA-256 + salt would be more appropriate. However, in the spirit of don't roll your own crypto an industry standard method such as PBKDF2 should be used.

BlueTechie: That's a very interesting point! I considered recommending salting in my review but as I understand it, using a cryptographically secure PRG for generating the API key values already addresses the security concern you mentioned. A PRG creates a high-entropy bitstring (computationally indistinguishable from a truly random one) and thus rainbow attacks should be highly unfeasible in this case - unlike with user passwords, which are low-entropy. I found the top answers to [1](https://security.stackexchange.com/questions/60691/length-of-passwords-that-are-rainbow-table-safe) and [2](https://security.stackexchange.com/questions/180345/do-i-need-to-hash-or-encrypt-api-keys-before-storing-them-in-a-database) to be interesting short reads on this subject for anyone interested.

 With that being said, however, using an industry standard such as PBKDF2 or _Argon2_ (the _latter_ being [more secure](https://community.bitwarden.com/t/pbkdf2-vs-argon2-which-is-better/59187/2)) like you recommend is not much extra effort and puts us on the more secure side of things so I would agree that this is a good idea.

bitmotivator: That may be true, especially for a low-risk situation like DAMAP, but given how cheap it became to create hashes with GPUs I wouldn't hold my breath for a single round of SHA-256 even if the underlying data is entirely random. This is not just for now, but also for the future where SHA-256 may have newly discovered weaknesses and generating hashes may have become magnitudes cheaper.

As you said, using an industry-standard password hashing method costs us basically nothing. Using multiple rounds buys us time to allow for a gradual rollover of keys instead of having to immediately reissue all API keys when a weakness is discovered.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.6.0_1.yaml
```diff
@@ -0,0 +1,97 @@
...
+        author: Geoffrey Karnbach
```
ValentinFutterer, line 4: Nitpick for the whole file: `nullable: false` is not necessary when you already specified `primaryKey: true` as a constraint, since it implies unique and non-nullable.

bitmotivator:
@GeoffreyKarnbach I have some capacity this week, do you prefer to finish this PR yourself or should I take this one over?

src/main/java/org/damap/base/rest/account/service/AccountService.java
BlueTechie: I found this in the quarkus [authorization docs](https://quarkus.io/guides/security-authorize-web-endpoints-reference): _Method-level security annotation must be placed on the endpoint implementation._ 
Looking at _AccountRessource.java_, the endpoint is currently defined as follows:
``` java 
  @POST
  @Path("/refresh-user-role")
  public void refreshUserApiKeyRole() {
    log.info("POST /account/refresh-user-role");
    String username = securityService.getUserId();
    String displayName = securityService.getDisplayName();
    String userId = securityService.getUserId();
    Set<String> roles = securityService.getUserRoles();
    accountService.refreshUserApiKeyRole(username, displayName, userId, roles);
  }
```
So we should put the annotation there to restrict access and not (as my previous comment indicated) in _AccountService.java_.

GeoffreyKarnbach:
> @GeoffreyKarnbach I have some capacity this week, do you prefer to finish this PR yourself or should I take this one over?

Hey @bitmotivator ,
Would be great if you could take it over, especially for a security related PR, it might be smarter, given the numerous feedback. Once you are done, can you please ping me, so that I can have a look at the code out of interest :)

src/main/java/org/damap/base/rest/account/service/AccountService.java
ValentinFutterer: If you could pass arbitrary roles here, you could essentially do so at any other endpoint in the application. The roles are read from the token and if users can manipulate that, its all over anyway, so I think this is save.

bitmotivator:
Thanks @GeoffreyKarnbach, will do!

src/main/java/org/damap/base/rest/account/service/AccountService.java
ValentinFutterer: I do not think it does, otherwise the whole thing is compromised (and many other java applciation also)

Also, you cannot enforce `@RolesAllowed("admin")` here, otherwise users without admin privileges cannot use it.

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+    apiKey.setUser(existingUser);
+    apiKey.persist();
+
+    return ApiKeyDO.builder().name(apiKey.getName()).value(apiKey.getValue()).build();
+  }
+
+  @Transactional
+  public void refreshUserApiKeyRole(
+      String username, String displayName, String userId, Set<String> roles) {
+    log.info("POST /account/refresh-user-role");
```
ValentinFutterer, line 83: Minor: This logging is already being done in the resource method.

src/main/java/org/damap/base/domain/User.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Column(nullable = false, unique = true)
+  private String username;
+
+  @OneToMany(mappedBy = "user")
+  private List<ApiKey> apiKeys;
+
+  @OneToMany(mappedBy = "user", fetch = FetchType.EAGER)
+  private List<UserRole> roles;
+}
```
ValentinFutterer, line 31: this is missing `cascade = CascadeType.ALL, orphanRemoval = true`, to properly propagate changes from user to the userRoles children.

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+          roles.stream()
+              .map(
+                  role -> {
+                    UserRole userRole = new UserRole();
+                    userRole.setUser(existingUser);
+                    userRole.setRole(role);
+                    return userRole;
+                  })
+              .toList());
+      existingUser.persist();
```
ValentinFutterer, line 99: This wasnt properly tested, since it doesnt add or remove userRoles when changes in permissions occur - see the comment in the User class. Hibernate doesnt deal well with replacing a whole collection, since it doesnt know what was removed (as opposed to modifying). I would sugest to reuse the old list and just clear it beforehand.

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+@ApplicationScoped
+@JBossLog
+public class AccountService {
+
+  @Inject private UserAccountRepo userAccountRepo;
+
+  @Inject private ApiKeyRepo apiKeyRepo;
+
+  public List<String> getApiKeys(String userId) {
+    log.info("GET /account/api-keys");
```
ValentinFutterer, line 31: Minor: Same log message as in the resource

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+    List<ApiKey> apiKeys = apiKeyRepo.findByUserId(userId);
+    return apiKeys.stream().map(ApiKey::getName).toList();
+  }
+
+  @Transactional
+  public ApiKeyDO createApiKey(
+      String keyName, String username, String displayName, String userId, Set<String> roles) {
+    log.info("POST /account/api-keys");
```
ValentinFutterer, line 39: Minor: Same log message as in the resource

src/main/java/org/damap/base/rest/account/service/AccountService.java
ValentinFutterer: True, this comment  should also be extended to all other methods in the service layer.

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+      existingUser.setDisplayName(displayName);
+      existingUser.setUserId(userId);
+      existingUser.persist();
+
+      for (String role : roles) {
+        UserRole userRole = new UserRole();
+        userRole.setUser(existingUser);
+        userRole.setRole(role);
+        userRole.persist();
+      }
```
ValentinFutterer, line 62: Minor: If you add the cascade all annotation to the user class, like discussed in another comment, you can add the userrole list to the user object and persist everything in one operation.

src/main/java/org/damap/base/rest/AccountResource.java
```diff
@@ -0,0 +1,82 @@
...
+    return accountService.createApiKey(keyName, username, displayName, userId, roles);
+  }
+
+  @POST
+  @Path("/api-keys/reset-all")
+  @RolesAllowed("Damap Admin")
+  public boolean resetAllApiKeys(@QueryParam("userId") String userId) {
+    log.info("POST /account/api-keys/reset-all for userId: " + userId);
+    return accountService.resetAllApiKeys(userId);
+  }
```
ValentinFutterer, line 51: Shouldnt this be a `@DELETE`, since it just deletes all keys for one user?

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+              .toList());
+      existingUser.persist();
+    } else {
+      throw new NotFoundException("User not found");
+    }
+  }
+
+  @Transactional
+  public void deleteApiKey(String key, String userId) {
+    log.info("DELETE /account/api-keys/" + key);
```
ValentinFutterer, line 107: Minor: Same log message as in the resource

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+    if (optionalApiKey.isPresent()) {
+      ApiKey apiKey = optionalApiKey.get();
+      if (Objects.equals(apiKey.getUser().getUserId(), userId)) {
+        apiKeyRepo.delete(apiKey);
+      } else {
+        throw new ForbiddenException("API key does not belong to the user");
+      }
+    } else {
+      throw new NotFoundException("API key not found");
+    }
```
ValentinFutterer, line 123: Is there a reason to use the name to search for the apikey to delete? Why not use the ID, seems a lot easier?

src/main/java/org/damap/base/rest/account/service/AccountService.java
```diff
@@ -0,0 +1,163 @@
...
+
+  @Transactional
+  public void deleteApiKey(String key, String userId) {
+    log.info("DELETE /account/api-keys/" + key);
+
+    if (key == null || key.isEmpty()) {
+      throw new IllegalArgumentException("API key cannot be null or empty");
+    }
+
+    Optional<ApiKey> optionalApiKey = apiKeyRepo.findByKey(key);
```
ValentinFutterer, line 113: Naming for key here confused me - I didnt know that it was the keyName until I went into the findBykey method to check. The same thing also happens at other locations. Sometimes it is called keyname though and soemthimes only key.

src/main/java/org/damap/base/domain/User.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ToString(exclude = {"apiKeys", "roles"})
+@Entity
+@Table(name = "user_account")
+public class User extends PanacheEntity {
```
ValentinFutterer, line 15: Is there a reason for the table to be named user_account and not just user. Seems like they are two different things.

src/main/java/org/damap/base/domain/ApiKey.java
```diff
@@ -0,0 +1,29 @@
...
+import lombok.ToString;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ToString(exclude = "user")
+@Entity
+@Table(name = "apikey")
+public class ApiKey extends PanacheEntity {
+
+  @Column private String name;
```
ValentinFutterer, line 21: In liquibase, this is non-nullable and unique.

ValentinFutterer:
There is documentation missing on how to use this feature :)

src/main/java/org/damap/base/security/SecurityService.java
```diff
@@ -51,30 +56,42 @@ public String getUserId() {
...
     final Principal principal = securityIdentity.getPrincipal();
-    if (!(principal instanceof OidcJwtCallerPrincipal)) return null;
-    return ((OidcJwtCallerPrincipal) principal).getName();
+    if ((principal instanceof OidcJwtCallerPrincipal)) {
+      return ((OidcJwtCallerPrincipal) principal).getName();
+    }
+    if (principal instanceof ApiKeyPrincipal) {
+      return ((ApiKeyPrincipal) principal).getName();
+    }
+    return null;
```
ValentinFutterer, line 65: Is there a reason for the instanceOf's here? Couldnt we just call getName on the principal?

src/main/java/org/damap/base/security/SecurityService.java
```diff
@@ -51,30 +56,42 @@ public String getUserId() {
...
-    String firstName = getClaimValueAsString(oidcPrincipal, "given_name");
-    String lastName = getClaimValueAsString(oidcPrincipal, "family_name");
-    if (firstName != null && lastName != null) {
-      return firstName + " " + lastName;
+    if (principal instanceof ApiKeyPrincipal) {
+      return ((ApiKeyPrincipal) principal).getDisplayName();
     }
 
-    return getClaimValueAsString(oidcPrincipal, "email");
+    return null;
```
ValentinFutterer, line 94: Why was this fallback to email removed?

bitmotivator:
As discussed in person, **we'll put this one on hold until we figure out how we are going to do authentication in DAMAP Cloud**. I will move the code into a branch prefixed `archive/` so we have it for later reference once I get around to it.

PR #381, opened 2025-06-04
Contributor: Select most recent ORCID affiliation

### LINKED ISSUE #144: Contributor affiliation selection: preferential treatment
The ORCIDRecord does not contain this information, but maybe there is an easy way to fetch the source of an affiliation and give preferential treatment to those that were not created by the orcid person. See the picture, right now DAMAP chooses "Interdisziplinäre Werteforschung"  instead of "Universität Wien" as the affiliation. Just a suggestion, if it is easily doable.
image

_Originally posted by @ValentinFutterer in https://github.com/tuwien-csd/damap-backend/pull/128#discussion_r1431086376_

GeoffreyKarnbach:
## Description
Feature

#### What does this PR do?
Update a given contributor to correct its affiliation and take the most recent one from ORCID.

closes GH-144

### DIFF
--- a/src/main/java/org/damap/base/rest/ORCIDPersonResource.java
+++ b/src/main/java/org/damap/base/rest/ORCIDPersonResource.java
@@ -1,0 +1,38 @@
+package org.damap.base.rest;
+
+import io.quarkus.security.Authenticated;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.enums.EIdentifierType;
+import org.damap.base.rest.dmp.domain.ContributorDO;
+import org.damap.base.rest.persons.orcid.ORCIDPersonServiceImpl;
+import org.eclipse.microprofile.openapi.annotations.parameters.RequestBody;
+
+/** PersonResource class. */
+@Path("/api/orcid")
+@Authenticated
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+@JBossLog
+public class ORCIDPersonResource {
+
+  @Inject ORCIDPersonServiceImpl orcidPersonService;
+
+  @POST
+  @Path("/affiliation")
+  public ContributorDO updateOrcidAffiliation(@RequestBody ContributorDO contributorDO) {
+
+    if (contributorDO.getPersonId().getType() != EIdentifierType.ORCID) {
+      throw new IllegalArgumentException(
+          "Cannot correct affiliations for contributor with identifier type: "
+              + contributorDO.getPersonId().getType());
+    }
+
+    return orcidPersonService.read(contributorDO.getPersonId().getIdentifier());
+  }
+}
--- a/src/main/java/org/damap/base/rest/PersonResource.java
+++ b/src/main/java/org/damap/base/rest/PersonResource.java
@@ -2,9 +2,7 @@ package org.damap.base.rest;
 
 import io.quarkus.security.Authenticated;
 import jakarta.inject.Inject;
-import jakarta.ws.rs.Consumes;
-import jakarta.ws.rs.Path;
-import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.*;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.UriInfo;
 import lombok.extern.jbosslog.JBossLog;


src/main/java/org/damap/base/rest/PersonServiceBroker.java
```diff
@@ -78,4 +79,8 @@ public PersonService getServiceForQueryParam(String searchServiceType) {
 
     return searchService;
   }
+
+  public ORCIDPersonServiceImpl getOrcidPersonService() {
+    return (ORCIDPersonServiceImpl) personServices.get("ORCID");
```
bitmotivator: I'm not a fan of putting this into the `PersonServiceBroker` as there is no brokering required to get this service. This is something where the ORCIDPersonServiceImpl could be injected directly into the target class.

src/main/java/org/damap/base/rest/PersonResource.java
```diff
@@ -57,4 +57,11 @@ public ResultList<ContributorDO> search(UriInfo uriInfo) {
 
     return result;
   }
+
+  @POST
+  @Path("/orcid/affiliation")
+  public ContributorDO updateOrcidAffiliation(@RequestBody ContributorDO contributorDO) {
+    ORCIDPersonServiceImpl orcidPersonService = personServiceBroker.getOrcidPersonService();
+    return orcidPersonService.getCorrectAffiliations(contributorDO);
```
bitmotivator: I believe that this should not be in the PersonResource. If we have ORCID-specific API calls, then this should be in a separate ORCIDPersonResource and that resource shouldn't call the PersonServiceBroker at all. (And I fully acknowledge that this is... less than ideal because it's a bunch of work.)

src/main/java/org/damap/base/rest/persons/orcid/ORCIDPersonServiceImpl.java
```diff
@@ -52,4 +53,18 @@ public ResultList<ContributorDO> search(MultivaluedMap<String, String> queryPara
...
+      throw new IllegalArgumentException(
+          "Cannot correct affiliations for contributor with identifier type: "
+              + contributorDO.getPersonId().getType());
+    }
+
+    String orcid = contributorDO.getPersonId().getIdentifier();
+    ContributorDO orcidRecord =
+        ORCIDMapper.mapRecordEntityToPersonDO(orcidRestClient.get(orcid), new ContributorDO());
+    return orcidRecord;
+  }
```
bitmotivator: I'm not quite sure I understand the reasoning behind this. Why isn't this logic included in the search() and read() functions above? Should we extend the ContributorDO to allow for multiple affiliations?

bitmotivator (commented):
Super unsure about this, so I'm just leaving a comment here.

src/main/java/org/damap/base/rest/persons/orcid/ORCIDPersonServiceImpl.java
GeoffreyKarnbach: Don't think we need multiple affiliations. The reasoning behind having it separately, is that it is only supposed to be used to "correct" the affiliation. Right now, we only can select the correct one, by making a specific get request for one ORCID, which is not the case in search, otherwise we would have 11 API calls per search request (so when for instance using the input field in the frontend).

GeoffreyKarnbach: Refactored the endpoint now to use read under the hood and removed "getCorrectAffiliations", you were right

src/main/java/org/damap/base/rest/PersonResource.java
GeoffreyKarnbach: Refactored into new resource

src/main/java/org/damap/base/rest/ORCIDPersonResource.java
```diff
@@ -0,0 +1,39 @@
...
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.enums.EIdentifierType;
+import org.damap.base.rest.dmp.domain.ContributorDO;
+import org.damap.base.rest.persons.orcid.ORCIDPersonServiceImpl;
+import org.eclipse.microprofile.openapi.annotations.parameters.RequestBody;
+
+/** PersonResource class. */
+@Path("/api/orcid")
```
ValentinFutterer, line 17: Not the same URL as in the frontend

src/main/java/org/damap/base/rest/PersonServiceBroker.java
ValentinFutterer: I think this comment wasnt addressed, or did you discuss this over other channels? @bitmotivator

bitmotivator: This is a one-liner and is only used in one place, this should be inlined to where it's used. Alternatively, the `ORCIDPersonServiceImpl` can be injected directly as @ValentinFutterer mentioned in person.

ValentinFutterer (approved):
👍

PR #385, opened 2025-06-11
Fixing container permissions for OpenShift deployment

bitmotivator:
This PR changes the `Dockerfile` to make it possible to run with a different user. This is necessary because OpenShift and hardened Kubernetes clusters typically impose additional restrictions and change the user ID.

Additionally, this PR also includes a change to make releases easier.


ghost (approved):
In the future, please do not include two different features/changes to PRs. Approving now since we're on the clock to deploy.

PR #392, opened 2025-06-12
Elsevier Pure implementation

### LINKED ISSUE #410: Pure integration
In order to make DAMAP Cloud viable, we must integrate Elsevier Pure into DAMAP natively.

bitmotivator:
This pull request implements parts of the [Elsevier Pure API](https://api.elsevierpure.com/ws/api/rapidoc.html) for the purposes of fetching projects and people in DAMAP.

Implementing the Pure API will allow several universities using Pure to integrate DAMAP without additional work required.

This initial implementation only reads data exports from Pure, a full implementation including authentication will follow later.

## TODO

- [x] Finish the project integration
- [x] Make the project integration configurable
- [x] Test if the Pure integration works when configured as a replacement for the mock services
- [x] Document how to set up the Pure integration using environment variables


lpandath:
@bitmotivator This PR does introduce breaking changes for institutional implementations. 

Our TU Graz PersonService currently implements:
`read(String id, MultivaluedMap<String, String> queryParams)`

But this method now throws RuntimeException by default. All institutional implementations need to update to the new signatures:
`read(String id)` 

Would it be possible to document this as a breaking change and include a migration guide? It would really help to understand how to update the existing implementations.

bitmotivator:
@lpandath thank you for your review!

Maybe I missed something, but the way it works in Java is that the [`default` implementation of an interface method is not used when you override it](https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html#extending). The default implementation for `read(String id)` calls `read(String id, MultivaluedMap<String,String> queryParams)`, which is your overridden method.

So, unless I made a mistake in the code, you should be able to continue implementing `read(String id, MultivaluedMap<String,String> queryParams)` and it should work the same. In addition, I will test this with the mock implementation before marking it ready for your review.

src/main/java/org/damap/base/integration/pure/PureAPI.java
```diff
@@ -0,0 +1,87 @@
...
+   *
+   * @return a list of all projects in the Pure database.
+   */
+  default List<PureAPIProject> listAllProjects() {
+    List<PureAPIProject> items = new ArrayList<>();
+    long offset = 0;
+    long pageSize = 10;
+    boolean finished = false;
+    while (!finished) {
+      PureAPIPaginatedProjectsResponse persons = listAllProjects(pageSize, offset);
```
lpandath: Variable named persons but contains projects?!

src/main/java/org/damap/base/integration/pure/PureAPI.java
```diff
@@ -0,0 +1,87 @@
...
+      items.addAll(persons.items);
+      if (persons.count <= persons.pageInformation.offset + persons.pageInformation.size) {
```
lpandath: The offset is set to the current offset instead of incrementing, which will cause an infinite loop no? 
Would it be better:
`offset += pageSize;` or `offset = persons.pageInformation.offset + persons.pageInformation.size;`

lpandath (commented):
Very well structured and clean interfaces! 
The bug would cause problems with large Pure datasets.

src/main/java/org/damap/base/integration/pure/PureAPI.java
```diff
@@ -0,0 +1,87 @@
...
+    boolean finished = false;
+    while (!finished) {
+      PureAPIPaginatedPersonsResponse persons = listAllPersons(pageSize, offset);
+      items.addAll(persons.items);
+      offset = persons.pageInformation.offset + persons.pageInformation.size;
+      if (persons.count <= offset) {
+        finished = true;
+      }
+    }
+    return items;
```
ValentinFutterer: I am not quite sure how this works:
What values do `persons.pageInformation.offset` and `persons.pageInformation.size` return?
If you call `listAllPersons` with offset = 0 and pageSize = 10, will they return 0 and 10 respectively? If yes, then `finsihed` will always be set to true.

bitmotivator: Unfortunately, I have no way to test this, but from the API docs I have to surmise that `persons.pageInformation.offset` will always be the offset passed in, while `persons.pageInformation.size` will either be the number of records currently returned, or the size passed in. However, `persons.count`, as far as I understand it, will always be the total number of records on the server, so `finished` should only be `true` once the `offset` variable reaches that number. @lpandath may be able to confirm if this works properly after testing.

lpandath: @bitmotivator  yes, i can test this that when integration in place.

lpandath: @bitmotivator 
https://github.com/damap-org/damap-backend/pull/405

src/main/java/org/damap/base/integration/pure/PureProjectService.java
```diff
@@ -0,0 +1,137 @@
...
+
+  @Override
+  public ProjectSupplementDO getProjectSupplement(String projectId) {
+    return new ProjectSupplementDO();
+  }
+
+  @Override
+  public ResultList<ProjectDO> getRecommended(Search search) {
+    return this.search(search);
+  }
```
ValentinFutterer: I think that this method here is broken, since the frontend passes an empty search by default - so it will always return all projects (or none, depending on how PURE is implemented). You need some custom code here that searches for all projects that includes the callers PURE ID.

ValentinFutterer (approved):
After much heartache, I think we can finally merge 🚀

PR #398, opened 2025-06-30
Refactoring to allow for configurable project services

bitmotivator:
This PR adds the ability to configure project services using environment variables. This is a requirement for #392 .

## TODO

- [x] Changelog?


CUSTOMISING.md
```diff
@@ -138,33 +138,37 @@ Also before version 3, all changesets where kept in db/changeLog.yaml. If you mi
...
 ### Configuring Project and Person API
-Provide your CRIS system and person database API addresses in the config:
 
-In your institutional project, write custom API services for retrieving project and person information from your
-services, as well as mappers, to map their information to damaps classes.
-To this end you need to implement the API services
-[ProjectService.class](src/main/java/org/damap/base/rest/projects/ProjectService.java) and
-[PersonService.class](src/main/java/org/damap/base/rest/persons/PersonService.java).
+DAMAP supports automatically populating project and person information from your CRIS system. It comes with a number of
+industry-standard integrations, such as ORCID or Elsevier Pure, but you can also create your own.
```
ValentinFutterer: This was split from the pure branch, so its not 100% up to date anymore.

CUSTOMISING.md
```diff
@@ -138,33 +138,37 @@ Also before version 3, all changesets where kept in db/changeLog.yaml. If you mi
...
-This will allow you to integrate several person services in parallel. The frontend will then 
-communicate with the service the user chooses. This component will provide the list of options to the frontend, 
-so no further changes are necessary there.
+<details><summary>application.yaml</summary>
 
 ```yaml
+## Custom project service:
+# project-service: your-service
+## Note: if left unconfigured, the natural Java CDI load order takes precedence. Use the
+## @Priority annotation to load your implementation first.
```
ValentinFutterer: I think the # in the third row was not intentional

its projects-service in the application.yaml

Also I do think the Note part is more confusing than it helps, without context i have no idea what steps i have to take and which to avoid. The application.yaml also doesnt help much, since there are no comments next to projects-service. I would make clear that the config id from the ProjectService implementation is expected here.

CUSTOMISING.md
```diff
@@ -138,33 +138,37 @@ Also before version 3, all changesets where kept in db/changeLog.yaml. If you mi
...
+
   person-services:
     - display-text: 'University'
       query-value: 'UNIVERSITY'
-      class-name: 'org.damap.base.rest.persons.MockUniversityPersonServiceImpl'
+      class-name: 'org.damap.base.integration.mock.MockUniversityPersonServiceImpl'
     - display-text: 'ORCID'
       query-value: 'ORCID'
-      class-name: 'org.damap.base.rest.persons.orcid.ORCIDPersonServiceImpl'
+      class-name: 'org.damap.base.integration.orcid.ORCIDPersonServiceImpl'
```
ValentinFutterer: Minor: Might be misleading to list here, since you probably shouldnt change that

ValentinFutterer (commented):
Looks good to me but disclaimer, I have never tested it with an actual old-style integration.

Maybe we should add some centralized documentation on how all of these person and project services integrations work in depth. A newer developer will have one hell of a time trying to understand all of this, even with previous knowledge it was confusing as hell. Though we can push this to a later date to release earlier.

CUSTOMISING.md
bitmotivator: I can live with this, we'll get rid of this document sooner rather than later. This needs to be moved to the website anyway since it's easier to update.

ValentinFutterer (approved):
👍

PR #399, opened 2025-06-30
Access: Add support for granting and revoking owner rights

### LINKED ISSUE #390: Enable multiple DMP owners
**Description**
The current DAMAP implementation allows only a single owner per DMP, limiting flexibility in real-world scenarios. When a consortium ends and data stewards need to transition DMP ownership, there's no mechanism to designate multiple owners or smoothly transfer ownership responsibilities.

**Proposed solution**
When a consortium ends, the original DMP owner _(e.g., consortium coordinator)_ needs to transfer ownership to data stewards or new data managers. However, during the transition period, multiple people may need owner-level permissions to be sure for continuity of data management.

**Alternative solution**
- allow multiple owners per DMP
- add UI checkbox/option to grant owner permissions 
- update backend validation to support multiple owners
- ensure at least one owner must always exist (prevent removing last owner)

**Additional context**

_If this issue requests multiple changes consider using checkboxes or lists._

ValentinFutterer:
## Description

 Feature

#### What does this PR do?
- When creating an access, old accesses for the user are automatically deleted, to avoid having multiple accesses floating around (there was not patch/put for access).
- When deleting accesses, it now makes sure there is always at least one owner left after deleting.
- Now, only the admin or owner of a DMP can view and interact with its accesses.

#### Code review focus
Security aspects. The surrounding code was not perfect and i refrained from a lot of otherwise necesary refactoring to save on time- this will be tackled on a later date.

#### Dependencies
https://github.com/damap-org/damap-frontend/issues/474

closes GH-390

### DIFF
--- a/src/main/java/org/damap/base/repo/AccessRepo.java
+++ b/src/main/java/org/damap/base/repo/AccessRepo.java
@@ -12,7 +12,7 @@ import org.damap.base.domain.Dmp;
 public class AccessRepo implements PanacheRepository<Access> {
 
   /**
-   * getAllByUniversityId.
+   * Returns all accesses by the institution/university-specific identifier of a person.
    *
    * @param universityId a {@link java.lang.String} object
    * @return a {@link java.util.List} object
--- a/src/main/java/org/damap/base/repo/AccessRepo.java
+++ b/src/main/java/org/damap/base/repo/AccessRepo.java
@@ -12,12 +12,12 @@ import org.damap.base.domain.Dmp;
 public class AccessRepo implements PanacheRepository<Access> {
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/damap/base/repo/AccessRepo.java
```diff
@@ -12,12 +12,12 @@
 public class AccessRepo implements PanacheRepository<Access> {
 
   /**
-   * getAllDmpByUniversityId.
+   * getAllByUniversityId.
```
bitmotivator:
```suggestion
   * Returns all accesses by the institution/university-specific identifier of a person.
```

bitmotivator (approved):

PR #400, opened 2025-06-30
Draft: production-grade Docker Compose deployment

bitmotivator:
This PR adds a production-grade DAMAP deployment using Docker Compose.


docker/docker-compose.dev.common.yaml
SotosTsepe, line 1: From personal experience: To ease the development process, I would move all the dev files to the root directory of the project.  This way, one can avoid changing to the `docker` directory each time a new session is used. A wrapper `docker-compose.yaml` file would also play nicely, as it would be a matter of just running `docker compose up -d`. To avoid confusion, the prod file should be renamed to `docker-compose.prod.yaml`

docker/prod/.env
SotosTsepe, line 1: I wouldn't recommend committing this file as `.env` but rather as `prod.example.env` to avoid possible confusion with a production-ready file (due to the values contained). When the developer needs to run explicitly:
```bash
cp prod.example.env .env
``` 
then (s)he knows that the values need to be changed. Leaving it as is could be dangerous for the impatient ones. Without the specific `.env` file, the guards in the `prod` compose files will ensure that the services won't spin up.

docker/prod/data/.gitkeep
SotosTsepe, line 1: Since `.gitignore` will create the `data` directory in version control, this `.gitkeep` should be redundant.

docker/prod/docker-compose.keycloak.db.yaml
SotosTsepe, line 1: Since this file is used only by Keycloak, is there a specific reason for the split and the extra profile?

docker/prod/docker-compose.keycloak.yaml
```diff
@@ -0,0 +1,44 @@
...
+    command:
+      - start
+      - --hostname=https://${DAMAP_HOSTNAME?Please provide a hostname for Keycloak}
+      - --hostname-backchannel-dynamic
+      - "true"
+      - --http-enabled=true
+      - --proxy-headers
+      - xforwarded
+      - --import-realm
+      - --hostname-debug=true
```
SotosTsepe, line 18: Not 100% sure about this flag, but I think it isn't suitable for a prod service.

docker/prod/docker-compose.keycloak.db.yaml
bitmotivator: Yes, the file is included by the main file. If it is not in an extra profile, there is no way to stop it from being applied. The idea is that you only call docker-compose once.

docker/prod/data/.gitkeep
bitmotivator: The data directory needs to exist, that's why the .gitkeep file is force-committed, but nothing in it will ever make it into git. If I remove .gitkeep, the folder will not exist.

docker/prod/generate-key.sh
SotosTsepe, line 1: Are we against committing a test `bundle.crt` and `private.pem` files? I think it would simplify the process by using them as bind mounts in the container. One would just need to replace them with the real files instead of running this script.

SotosTsepe (commented):
Solid work overall! I’d suggest adding some new lines to the yaml files for better readability, as well as at the end of the files. We must also make sure the CI pipeline works before merging. I can jump in and assist with this PR right away.

docker/prod/generate-key.sh
bitmotivator: Yes, because GitHub security scanning will scream bloody murder if we do. Generally, having anything that looks like a secret in Git is a bad idea because it's then easy to miss when the "real" certificates are accidentally committed.

SotosTsepe:
Superseded by #479

PR #404, opened 2025-07-14
Fixing failed Maven Central deployment

bitmotivator:


.github/workflows/release-maven.yml
```diff
@@ -15,7 +15,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-
+      - name: Change Maven version
+        if: startsWith(github.ref, 'refs/tags/')
```
ValentinFutterer, line 19: Minor question since I am a GitHub actions noob :)
Is this if necessary - the action should only trigger when a tag is pushed right?

.github/workflows/release-maven.yml
```diff
@@ -15,7 +15,11 @@ jobs:
...
+        run: |
+          VERSION=$(echo -n '${{ github.ref_name }}' | sed -e 's$refs/tags/$$' | sed -e 's/v//')
```
ValentinFutterer: `| sed -e 's$refs/tags/$$'` can be removed, since `github.ref_name` already omits it

ValentinFutterer:
Ah and please update the changelog :)

.github/workflows/release-maven.yml
bitmotivator: I'd like to keep it just in case this workflow gets edited. :)

.github/workflows/release-maven.yml
bitmotivator: Oh really? We have it like this in the other one too.

bitmotivator:
@ValentinFutterer this is not a user-visible change, this doesn't need to be in the changelog.

ValentinFutterer (approved):
Lets try if it works ;)

PR #419, opened 2025-08-07
Global: Add custom error handling for external services

ValentinFutterer:
## Description

Refactoring

#### What does this PR do?
Add custom error handling for all external clients using Quarkus ClientExceptionMapper. New exceptions carry a detailed message for a console log and a machine readable error code, that can be interpreted by the frontend. Now users see a actual cause when a service like Re3Data is down. 

I also added 10 second timeouts and fallbacks, that catch a service not responding.

I did not write the error handling for the mock project/person services since they should never be down while the application is still up and I will add the error handling to the Pure PR.

Connected [Frontend PR](https://github.com/damap-org/damap-frontend/pull/484) where the error codes are processed.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created


bitmotivator (approved):
I like this a lot.

src/main/java/org/damap/base/enums/EErrorCode.java
```diff
@@ -0,0 +1,20 @@
...
+  ORCID_NOT_AVAILABLE,
+  ORCID_UNEXPECTED_ERROR,
+  FITS_NOT_AVAILABLE,
+  FITS_UNEXPECTED_ERROR,
+  OPENAIRE_NOT_FOUND,
+  OPENAIRE_NOT_AVAILABLE,
+  OPENAIRE_UNEXPECTED_ERROR,
+  GOTENBERG_NOT_AVAILABLE,
+  GOTENBERG_UNEXPECTED_ERROR
+}
```
DanceaVlad, line 23:
```suggestion
public enum EErrorCode {
  RE3DATA_NOT_FOUND(
          "A repository couldn't be found in Re3Data",
          Response.Status.NOT_FOUND
  ),
  RE3DATA_NOT_AVAILABLE(
          "Re3Data service is currently not available",
          Response.Status.SERVICE_UNAVAILABLE
  ),
  RE3DATA_UNEXPECTED_ERROR(
          "Something unexpected happened while calling Re3Data",
          Response.Status.INTERNAL_SERVER_ERROR
  ),
  RE3DATA_RECOMMENDED_NOT_FOUND(
          "A recommended repository couldn't be found in Re3Data",
          Response.Status.NOT_FOUND
  ),
  RE3DATA_RECOMMENDED_NOT_AVAILABLE(
          "Re3Data service is currently not available for recommended repositories",
          Response.Status.SERVICE_UNAVAILABLE
  ),
  RE3DATA_RECOMMENDED_UNEXPECTED_ERROR(
          "Something unexpected happened while fetching recommended repositories from Re3Data",
          Response.Status.INTERNAL_SERVER_ERROR
  );

  private final String message;
  private final Response.Status httpStatus;

  EErrorCode(String message, Response.Status httpStatus) {
    this.message = message;
    this.httpStatus = httpStatus;
  }

  public String getMessage() {
    return message;
  }

  public Response.Status getHttpStatus() {
    return httpStatus;
  }
}

```

src/main/java/org/damap/base/r3data/RepositoriesRemoteResource.java
```diff
@@ -70,4 +83,65 @@ generated.List search(
...
+                  "Re3Data is currently not available, caused by " + response.getStatus()),
+              Response.Status.INTERNAL_SERVER_ERROR);
+      default ->
+          new DamapApiException(
+              new ErrorDto(
+                  EErrorCode.RE3DATA_UNEXPECTED_ERROR,
+                  "Something unexpected happened while calling Re3Data, caused by "
+                      + response.getStatus()),
+              Response.Status.INTERNAL_SERVER_ERROR);
+    };
```
DanceaVlad, line 111:
```suggestion
    EErrorCode errorCode = switch (response.getStatus()) {
      case 404 -> RE3DATA_NOT_FOUND;
      case 500, 502, 503, 504 -> RE3DATA_NOT_AVAILABLE;
      default -> RE3DATA_UNEXPECTED_ERROR;
    };

    return new DamapApiException(
            new ErrorDto(errorCode, errorCode.getMessage()),
            errorCode.getHttpStatus());
```

src/main/java/org/damap/base/r3data/RepositoriesResource.java
```diff
@@ -6,21 +6,21 @@
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
-import jakarta.ws.rs.core.Context;
-import jakarta.ws.rs.core.MediaType;
-import jakarta.ws.rs.core.MultivaluedMap;
-import jakarta.ws.rs.core.UriInfo;
+import jakarta.ws.rs.core.*;
```
DanceaVlad: At my last internship we had a rule to never use star (*) imports to avoid ambiguity when a class could be imported from 2 places. 
Like here you have **jakarta.ws.rs.core.*** and are using **Context** but you may also have it from **javax.naming.*** . Don't know how important that is, thought I would just mention it

src/main/java/org/damap/base/enums/EErrorCode.java
ValentinFutterer: Hey, in theory i like the suggestion. I didnt impleemnt it though, due to these reasons:

- Timeconstraints, since they do not add anything in terms of functionality
- Your suggestion would discard the information on which error originally caused the exception
- RE3DATA_NOT_AVAILABLE could be due to any 500 error or due to timeout - this nuance would also be lost with how you described the solution; and yes you could just add a different timeout error type, but see reason 1 :)

ValentinFutterer: We cna maybe refactor at a later date.

src/main/java/org/damap/base/r3data/RepositoriesResource.java
ValentinFutterer: Good catch

src/main/java/org/damap/base/integration/orcid/OrcidPersonService.java
```diff
@@ -1,20 +1,30 @@
 package org.damap.base.integration.orcid;
 
+import static org.reflections.Reflections.log;
```
lpandath: this is not an application logger?!  Use Logger.getLogger(ClassName.class) instead.          

Logger log = Logger.getLogger(OrcidPersonService.class.getName());

src/main/java/org/damap/base/integration/pure/HTTPBasedPureAPI.java
```diff
@@ -1,9 +1,18 @@
 package org.damap.base.integration.pure;
 
+import static org.reflections.Reflections.log;
```
lpandath: same

src/main/java/org/damap/base/rest/document/service/GotenbergRestService.java
```diff
@@ -1,19 +1,61 @@
 package org.damap.base.rest.document.service;
 
+import static org.reflections.Reflections.log;
```
lpandath: same

src/main/java/org/damap/base/rest/openaire/OpenAireRemoteResource.java
```diff
@@ -1,16 +1,25 @@
 package org.damap.base.rest.openaire;
 
+import static org.reflections.Reflections.log;
```
lpandath: same

src/main/java/org/damap/base/r3data/RepositoriesRemoteResource.java
```diff
@@ -1,10 +1,19 @@
 package org.damap.base.r3data;
 
+import static org.reflections.Reflections.log;
```
lpandath: same

src/main/java/org/damap/base/r3data/RepositoriesService.java
```diff
@@ -47,11 +51,21 @@ public List<RepositoryDetails> getRecommended() {
...
+          case RE3DATA_NOT_FOUND ->
+              errorPayload = new ErrorDto(RE3DATA_RECOMMENDED_NOT_FOUND, e.getPayload().details());
+          case RE3DATA_NOT_AVAILABLE ->
+              errorPayload =
+                  new ErrorDto(RE3DATA_RECOMMENDED_NOT_AVAILABLE, e.getPayload().details());
+          case RE3DATA_UNEXPECTED_ERROR ->
+              errorPayload =
+                  new ErrorDto(RE3DATA_RECOMMENDED_UNEXPECTED_ERROR, e.getPayload().details());
+        }
+        throw new DamapApiException(errorPayload, e.getStatus(), e);
```
lpandath, line 68: Before this change, if one recommended repo lookup failed, the rest still loaded. Now it throws and stops  so one failure breaks the whole list. Intended?

src/main/java/org/damap/base/integration/orcid/OrcidPersonService.java
ValentinFutterer: Yes good catch, I let my IDE import a logger without thinking about it 😅

src/main/java/org/damap/base/r3data/RepositoriesService.java
ValentinFutterer: True, but I believe its fine. Chances are if one lookup fails, the others will so too. With our new recommended repository input mask, its very unlikely that bad IDs get into the DB, so we can rule out one ID that always fails. And in the unliekly event that the lookup fails for only one randomly because the server is down in that moment, I think its better behaviour to fail everything.

PR #425, opened 2025-09-05
Generate sources package by default

bitmotivator:
This PR changes the pom.xml to generate the sources jar by default. This is useful in case we need to manually create Maven Central artifacts.


ValentinFutterer (approved):
👍

PR #426, opened 2025-09-11
lth/Vd/407 make frontend customizable with config api

### LINKED ISSUE #407: Make frontend customizations configurable using the config API
In order to make multi-tenancy viable, we need to make the frontend configurable though the config API we already have. This includes:

- Colors
- Logos
- Possibly the customizable texts

lpandath:
## Description
Allows the admin to save configurations for Colors, Images, and Recommended Repositories in the database.

#### What does this PR do?

_Color Theme_
Allows for colors to be saved in the database, from which the frontend generates the color scheme.

_Images_
Allows for images to be saved in the database, which the frontend loads instead of the default images located in the assets folder.

_Recommended Repositories_
Replaces hardcoded repositories from application.yaml with a database table.

#### Dependencies
Frontend: https://github.com/damap-org/damap-frontend/pull/488

@DanceaVlad  - Initial implementation of theme service and theme image resources, final cleanup of all three changes
@lpandath  - continued dev of image service and repository management system

closes GH-407

### DIFF
--- a/src/main/java/org/damap/base/domain/Image.java
+++ b/src/main/java/org/damap/base/domain/Image.java
@@ -1,0 +1,38 @@
+package org.damap.base.domain;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import jakarta.persistence.Basic;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.Lob;
+import jakarta.persistence.Table;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+
+@Data
+@EqualsAndHashCode(
+    callSuper = true,
+    exclude = {"data"})
+@AllArgsConstructor
+@NoArgsConstructor
+@Entity
+@Table(name = "image")
+public class Image extends PanacheEntity {
+
+  @Column(name = "image_key", nullable = false, unique = true)
+  private String imageKey;
+
+  @Lob
+  @Basic(fetch = FetchType.LAZY)
+  @Column(name = "data", nullable = false)
+  private byte[] data;
+
+  @Column(name = "filesize", nullable = false)
+  private Long filesize; // in bytes
+
+  @Column(name = "mime_type", nullable = false)
+  private String mimeType;
+}
--- a/src/main/java/org/damap/base/domain/RecommendedRepository.java
+++ b/src/main/java/org/damap/base/domain/RecommendedRepository.java
@@ -1,0 +1,22 @@
+package org.damap.base.domain;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Table;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.Size;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data
+@Entity
+@EqualsAndHashCode(callSuper = true)
+@Table(name = "recommended_repository")
+public class RecommendedRepository extends PanacheEntity {
+
+  @NotBlank
+  @Size(max = 255)
+  @Column(name = "repository_id", nullable = false, unique = true)
+  private String repositoryId;
+}
--- a/src/main/java/org/damap/base/r3data/RepositoriesService.java
+++ b/src/main/java/org/damap/base/r3data/RepositoriesService.java
@@ -1,17 +1,17 @@
 package org.damap.base.r3data;
 
 import io.quarkus.cache.CacheResult;
+import io.quarkus.panache.common.Sort;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.MultivaluedMap;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Optional;
 import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.domain.RecommendedRepository;
 import org.damap.base.enums.EIdentifierType;
 import org.damap.base.r3data.dto.RepositoryDetails;
 import org.damap.base.r3data.mapper.RepositoryMapper;
-import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
 import org.re3data.schema._2_2.Re3Data;
 
@@ -21,9 +21,6 @@ import org.re3data.schema._2_2.Re3Data;
 public class RepositoriesService {
 
   @Inject @RestClient RepositoriesRemoteResource repositoriesRemoteResource;
-
-  @ConfigProperty(name = "damap.repositories.recommendation")
-  Optional<String[]> repositoriesRecommendation;
 
   /**
    * getAll.
@@ -40,24 +37,24 @@ public class RepositoriesService {
    *
    * @return a {@link java.util.List} object
    */
-  @CacheResult(cacheName = "recommendedRepositories")
   public List<RepositoryDetails> getRecommended() {
-    List<RepositoryDetails> recommendedRepositories = new ArrayList<>();
-    if (repositoriesRecommendation.isEmpty()) {
-      return recommendedRepositories;
-    }
+    List<RecommendedRepository> recommendedRepositories =
+        RecommendedRepository.listAll(Sort.by("id").ascending());
+    List<RepositoryDetails> recommendedRepositoryDetails = new ArrayList<>();
 
-    for (String id : repositoriesRecommendation.get()) {
-      if (id.startsWith("r3d")) {
-        try {
-          Re3Data repo = this.getById(id);
-          recommendedRepositories.add(RepositoryMapper.mapToRepositoryDetails(repo, id));
-        } catch (Exception e) {
-          log.infov("Failed to retrieve repository for ID {0}, error: {1}", id, e.getMessage());
-        }
+    for (RecommendedRepository recommendedRepository : recommendedRepositories) {
+      String repositoryId = recommendedRepository.getRepositoryId();
+
+      try {
+        Re3Data repo = this.getById(repositoryId);
+        RepositoryDetails details = RepositoryMapper.mapToRepositoryDetails(repo, repositoryId);
+        recommendedRepositoryDetails.add(details);
+      } catch (Exception e) {
+        log.infov(
+            "Failed to retrieve repository for ID {0}, error: {1}", repositoryId, e.getMessage());
       }
     }
-    return recommendedRepositories;
+    return recommendedRepositoryDetails;
   }
 
   /**
--- a/src/main/java/org/damap/base/rest/AdminResource.java
+++ b/src/main/java/org/damap/base/rest/AdminResource.java
@@ -4,11 +4,29 @@ import io.quarkus.security.Authenticated;
 import jakarta.annotation.security.RolesAllowed;
 import jakarta.inject.Inject;
 import jakarta.validation.Valid;
-import jakarta.ws.rs.*;
+import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.DELETE;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.PUT;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.core.MediaType;
+import java.util.List;
 import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.domain.ColorTheme;
+import org.damap.base.domain.Image;
+import org.damap.base.domain.RecommendedRepository;
 import org.damap.base.rest.admin.domain.BannerDO;
+import org.damap.base.rest.admin.domain.RecommendedRepositoryDO;
 import org.damap.base.rest.admin.service.AdminService;
+import org.damap.base.rest.admin.service.RecommendedRepositoryService;
+import org.damap.base.rest.theme.service.ColorThemeService;
+import org.damap.base.rest.theme.service.ImageThemeService;
+import org.jboss.resteasy.reactive.RestForm;
+import org.jboss.resteasy.reactive.RestPath;
+import org.jboss.resteasy.reactive.multipart.FileUpload;
 
 /** AdminResource class. */
 @Path("/api/admin")
@@ -18,11 +36,14 @@ import org.damap.base.rest.admin.service.AdminService;
 public class AdminResource {
 
   @Inject AdminService adminService;
+  @Inject RecommendedRepositoryService recommendedRepositoryService;
+  @Inject ColorThemeService colorThemeService;
+  @Inject ImageThemeService imageThemeService;
 
   @GET
   @Path("/banner")
   public BannerDO getAppBanner() {
-    log.info("GET /admin/banner");
+    log.info("Retrieving application banner");
     return this.adminService.getAppBanner();
   }
 
@@ -30,7 +51,7 @@ public class AdminResource {
   @Path("/banner")
   @RolesAllowed("${damap.auth.admin-role-name}")
   public BannerDO createAppBanner(@Valid BannerDO bannerDO) {
-    log.info("POST /admin/banner");
+    log.info("Creating new application banner");
     log.info(bannerDO);
     return this.adminService.createAppBanner(bannerDO);
   }
@@ -39,7 +60,7 @@ public class AdminResource {
   @Path("/banner")
   @RolesAllowed("${damap.auth.admin-role-name}")
   public BannerDO updateAppBanner(@Valid BannerDO bannerDO) {
-    log.info("PUT /admin/banner");
+    log.info("Updating application banner");
     log.info(bannerDO);
     return this.adminService.updateAppBanner(bannerDO);
   }
@@ -48,7 +69,71 @@ public class AdminResource {
   @Path("/banner")
   @RolesAllowed("${damap.auth.admin-role-name}")
   public void deleteAppBanner() {
-    log.info("DELETE /admin/banner");
+    log.info("Deleting application banner");
     this.adminService.deleteAppBanner();
+  }
+
+  @GET
+  @Path("/color-theme")
+  @RolesAllowed("${damap.auth.admin-role-name}")
+  public ColorTheme getColorTheme() {
+    log.info("Retrieving color theme configuration");
+    return this.colorThemeService.getTheme();
+  }
+
+  @PUT
+  @Path("/color-theme")
+  @RolesAllowed("${damap.auth.admin-role-name}")
+  public ColorTheme uploadColorTheme(ColorTheme colorTheme) {
+    log.info("Updating color theme configuration");
+    return colorThemeService.uploadTheme(colorTheme);
+  }
+
+  @PUT
+  @Path("/image-theme")
+  @RolesAllowed("${damap.auth.admin-role-name}")
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
+  public Image uploadImage(
+      @RestForm("imageKey") String imageKey, @RestForm("file") FileUpload file) {
+    log.info("Updating image with key: " + imageKey);
+    return imageThemeService.uploadImage(imageKey, file);
+  }
+
+  @DELETE
+  @Path("/image-theme")
+  @RolesAllowed("${damap.auth.admin-role-name}")
+  public void deleteImage(@QueryParam("imageKey") String imageKey) {
+    log.info("Deleting image with key: " + imageKey);
+    imageThemeService.deleteImage(imageKey);
+  }
+
+  @GET
+  @Path("/recommended-repositories")
+  @RolesAllowed("${damap.auth.admin-role-name}")
+  public List<RecommendedRepositoryDO> getRecommendedRepositories() {
+    log.info("Retrieving all recommended repositories");
+    return this.recommendedRepositoryService.getRecommendedRepositoriesWithNames();
+  }
+
+  @POST
+  @Path("/recommended-repositories")
+  @RolesAllowed("${damap.auth.admin-role-name}")
+  public RecommendedRepositoryDO createRecommendedRepository(
+      @Valid RecommendedRepositoryDO recommendedRepositoryDO) {
+    log.info("Creating new recommended repository: " + recommendedRepositoryDO.getRepositoryId());
+    RecommendedRepository recommendedRepository = new RecommendedRepository();
+    recommendedRepository.setRepositoryId(recommendedRepositoryDO.getRepositoryId());
+    RecommendedRepository created =
+        this.recommendedRepositoryService.createRecommendedRepository(recommendedRepository);
+    recommendedRepositoryDO.setId(created.id);
+    return recommendedRepositoryDO;
+  }
+
+  @DELETE
+  @Path("/recommended-repositories/{id}")
+  @RolesAllowed("${damap.auth.admin-role-name}")
+  public void deleteRecommendedRepository(@RestPath Long id) {
+    log.info("Deleting recommended repository with id: " + id);
+    this.recommendedRepositoryService.deleteRecommendedRepository(id);
   }
 }
--- a/src/main/java/org/damap/base/rest/admin/domain/RecommendedRepositoryDO.java
+++ b/src/main/java/org/damap/base/rest/admin/domain/RecommendedRepositoryDO.java
@@ -1,0 +1,10 @@
+package org.damap.base.rest.admin.domain;
+
+import lombok.Data;
+
+@Data
+public class RecommendedRepositoryDO {
+  private Long id;
+  private String repositoryId;
+  private String name;
+}
--- a/src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
+++ b/src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
@@ -1,0 +1,103 @@
+package org.damap.base.rest.admin.service;
+
+import io.quarkus.panache.common.Sort;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import jakarta.transaction.Transactional;
+import jakarta.validation.Valid;
+import jakarta.ws.rs.BadRequestException;
+import jakarta.ws.rs.ClientErrorException;
+import jakarta.ws.rs.NotFoundException;
+import jakarta.ws.rs.core.Response;
+import java.util.ArrayList;
+import java.util.List;
+import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.domain.RecommendedRepository;
+import org.damap.base.r3data.RepositoriesService;
+import org.damap.base.rest.admin.domain.RecommendedRepositoryDO;
+import org.re3data.schema._2_2.Re3Data;
+
+@ApplicationScoped
+@JBossLog
+public class RecommendedRepositoryService {
+
+  @Inject RepositoriesService repositoriesService;
+
+  private boolean isValidRepositoryId(String repositoryId) {
+    if (repositoryId == null) {
+      return false;
+    }
+    if (!repositoryId.matches("^r3d\\d{6,}$")) { // not the official pattern
+      return false;
+    }
+    try {
+      repositoriesService.getById(repositoryId);
+      return true;
+    } catch (Exception e) {
+      return false;
+    }
+  }
+
+  public List<RecommendedRepositoryDO> getRecommendedRepositoriesWithNames() {
+    List<RecommendedRepository> recommendedRepositories =
+        RecommendedRepository.listAll(Sort.by("id").ascending());
+    List<RecommendedRepositoryDO> result = new ArrayList<>();
+
+    for (RecommendedRepository repo : recommendedRepositories) {
+      RecommendedRepositoryDO dto = new RecommendedRepositoryDO();
+      dto.setId(repo.id);
+      dto.setRepositoryId(repo.getRepositoryId());
+
+      try {
+        Re3Data re3Data = repositoriesService.getById(repo.getRepositoryId());
+        if (!re3Data.getRepository().isEmpty()) {
+          dto.setName(re3Data.getRepository().get(0).getRepositoryName().getValue());
+        }
+      } catch (Exception e) {
+        log.infov(
+            "Failed to retrieve repository name for ID {0}, error: {1}",
+            repo.getRepositoryId(), e.getMessage());
+      }
+
+      result.add(dto);
+    }
+
+    return result;
+  }
+
+  @Transactional
+  public RecommendedRepository createRecommendedRepository(
+      @Valid RecommendedRepository recommendedRepository) {
+    RecommendedRepository existing =
+        RecommendedRepository.find("repositoryId = ?1", recommendedRepository.getRepositoryId())
+            .firstResult();
+
+    if (existing != null) {
+      throw new ClientErrorException(
+          "Repository with repository ID '"
+              + recommendedRepository.getRepositoryId()
+              + "' already exists",
+          Response.Status.CONFLICT);
+    }
+
+    // validate and only allow via Re3Data IDs
+    if (!isValidRepositoryId(recommendedRepository.getRepositoryId())) {
+      throw new BadRequestException(
+          "Invalid or unknown Re3Data repository ID: " + recommendedRepository.getRepositoryId());
+    }
+
+    recommendedRepository.persist();
+
+    return recommendedRepository;
+  }
+
+  @Transactional
+  public void deleteRecommendedRepository(Long id) {
+    RecommendedRepository repository = RecommendedRepository.findById(id);
+    if (repository == null) {
+      throw new NotFoundException("Repository with database ID '" + id + "' does not exist");
+    }
+
+    repository.delete();
+  }
+}
--- a/src/main/java/org/damap/base/rest/theme/service/ColorThemeService.java
+++ b/src/main/java/org/damap/base/rest/theme/service/ColorThemeService.java
@@ -1,0 +1,24 @@
+package org.damap.base.rest.theme.service;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.transaction.Transactional;
+import jakarta.validation.Valid;
+import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.domain.ColorTheme;
+
+@ApplicationScoped
+@JBossLog
+public class ColorThemeService {
+
+  @Transactional
+  public ColorTheme getTheme() {
+    return ColorTheme.findAll().firstResult();
+  }
+
+  @Transactional
+  public ColorTheme uploadTheme(@Valid ColorTheme colorTheme) {
+    ColorTheme.deleteAll();
+    colorTheme.persist();
+    return colorTheme;
+  }
+}
--- a/src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
+++ b/src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
@@ -1,0 +1,183 @@
+package org.damap.base.rest.theme.service;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.transaction.Transactional;
+import jakarta.ws.rs.BadRequestException;
+import jakarta.ws.rs.ClientErrorException;
+import jakarta.ws.rs.InternalServerErrorException;
+import jakarta.ws.rs.NotFoundException;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.domain.Image;
+import org.hibernate.Hibernate;
+import org.jboss.resteasy.reactive.multipart.FileUpload;
+
+@ApplicationScoped
+@JBossLog
+public class ImageThemeService {
+
+  public enum AllowedImageType {
+    PNG("image/png"),
+    JPEG("image/jpeg"),
+    JPG("image/jpg"),
+    GIF("image/gif"),
+    SVG("image/svg+xml"),
+    WEBP("image/webp");
+
+    private final String mimeType;
+
+    AllowedImageType(String mimeType) {
+      this.mimeType = mimeType;
+    }
+
+    public String getMimeType() {
+      return mimeType;
+    }
+
+    public static boolean isAllowed(String mimeType) {
+      if (mimeType == null) {
+        return false;
+      }
+      for (AllowedImageType type : values()) {
+        if (type.mimeType.equals(mimeType)) {
+          return true;
+        }
+      }
+      return false;
+    }
+  }
+
+  // Maximum file size: 5MB
+  private static final long MAX_FILE_SIZE = 5 * 1024 * 1024;
+
+  @Transactional
+  public Image uploadImage(String imageKey, FileUpload file) {
+    if (imageKey == null || imageKey.isBlank()) {
+      throw new BadRequestException("Image key is required");
+    }
+
+    if (file == null || file.uploadedFile() == null) {
+      throw new BadRequestException("File is required");
+    }
+
+    try {
+      byte[] imageData = Files.readAllBytes(file.uploadedFile());
+
+      if (imageData.length == 0) {
+        throw new BadRequestException("No image data provided");
+      }
+
+      if (imageData.length > MAX_FILE_SIZE) {
+        throw new ClientErrorException(
+            "File size exceeds maximum allowed size of " + (MAX_FILE_SIZE / 1024 / 1024) + "MB",
+            Response.Status.REQUEST_ENTITY_TOO_LARGE);
+      }
+
+      String mimeType = detectMimeType(file);
+
+      if (!isAllowedImageType(mimeType)) {
+        throw new ClientErrorException(
+            "Unsupported image type: " + mimeType + ". Allowed types: " + getAllowedTypesString(),
+            Response.Status.UNSUPPORTED_MEDIA_TYPE);
+      }
+
+      Image existingImage = Image.find("imageKey", imageKey).firstResult();
+
+      Image image;
+      if (existingImage != null) {
+        existingImage.setData(imageData);
+        existingImage.setFilesize((long) imageData.length);
+        existingImage.setMimeType(mimeType);
+        existingImage.persist();
+        log.info("Updated existing image for key: " + imageKey);
+        image = existingImage;
+      } else {
+        image = new Image(imageKey, imageData, (long) imageData.length, mimeType);
+        image.persist();
+        log.info("Created new image for key: " + imageKey);
+      }
+
+      return image;
+
+    } catch (IOException e) {
+      throw new InternalServerErrorException("Failed to upload image", e);
+    }
+  }
+
+  @Transactional
+  public List<Image> getImages() {
+    List<Image> images = Image.findAll().list();
+    images.forEach(img -> Hibernate.initialize(img.getData()));
+    return images;
+  }
+
+  @Transactional
+  public void deleteImage(String imageKey) {
+    if (imageKey == null || imageKey.isBlank()) {
+      throw new BadRequestException("Image key is required");
+    }
+
+    if (Image.delete("imageKey", imageKey) == 0) {
+      throw new NotFoundException("Image not found for key: " + imageKey);
+    }
+  }
+
+  private String detectMimeType(FileUpload file) {
+    // MIME from header
+    String mimeType = file.contentType();
+    if (mimeType != null && !mimeType.isBlank()) {
+      return mimeType;
+    }
+
+    // MIME from filename
+    String filename = file.fileName();
+    if (filename != null && !filename.isBlank()) {
+      mimeType = detectMimeTypeFromExtension(filename);
+      if (mimeType != null) {
+        return mimeType;
+      }
+    }
+
+    // MIME from content
+    try {
+      mimeType = Files.probeContentType(file.uploadedFile());
+      if (mimeType != null && !mimeType.isBlank()) {
+        return mimeType;
+      }
+    } catch (IOException e) {
+      // ignore and return the fallback
+    }
+
+    // fallback
+    return MediaType.APPLICATION_OCTET_STREAM;
+  }
+
+  private String detectMimeTypeFromExtension(String filename) {
+    String extension = filename.substring(filename.lastIndexOf('.') + 1).toLowerCase();
+    return switch (extension) {
+      case "png" -> "image/png";
+      case "jpg", "jpeg" -> "image/jpeg";
+      case "gif" -> "image/gif";
+      case "svg" -> "image/svg+xml";
+      case "webp" -> "image/webp";
+      case "ico" -> "image/x-icon";
+      default -> null;
+    };
+  }
+
+  private String getAllowedTypesString() {
+    StringBuilder sb = new StringBuilder();
+    for (AllowedImageType type : AllowedImageType.values()) {
+      sb.append(type.getMimeType()).append(", ");
+    }
+    return sb.toString().substring(0, sb.length() - 2);
+  }
+
+  private boolean isAllowedImageType(String mimeType) {
+    return AllowedImageType.isAllowed(mimeType);
+  }
+}
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -23,7 +23,7 @@ damap:
     # Role name used to identify DAMAP super admins
     # If DAMAP is locally deployed ADMIN_ROLE constant in the DefaultProfile test profile needs to be
     # the same as here, otherwise the tests fail
-    admin-role-name: DAMAP ADMIN
+    admin-role-name: Damap Admin
   repositories:
     # Zenodo, Open Science Framework, DRYAD, Mendeley Data
     # can also be [] for no repositories recommendation
--- a/src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_1.yaml
+++ b/src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_1.yaml
@@ -1,0 +1,100 @@
+databaseChangeLog:
+  - changeSet:
+      id: 25
+      author: Vlad Dancea
+      comment: Create color-theme, image-theme and recommended-repository tables
+      changes:
+        - createSequence:
+            sequenceName: color_theme_seq
+            startValue: 1
+            incrementBy: 50
+
+        - createTable:
+            tableName: color_theme
+            columns:
+              - column:
+                  name: id
+                  type: BIGINT
+                  constraints:
+                    primaryKey: true
+              - column:
+                  name: exact_colors
+                  type: BOOLEAN
+                  defaultValueBoolean: false
+                  constraints:
+                    nullable: false
+              - column:
+                  name: colors
+                  type: CLOB
+                  constraints:
+                    nullable: false
+
+        - createSequence:
+            sequenceName: image_seq
+            startValue: 1
+            incrementBy: 50
+
+        - createTable:
+            tableName: image
+            columns:
+              - column:
+                  name: id
+                  type: BIGINT
+                  constraints:
+                    primaryKey: true
+              - column:
+                  name: image_key
+                  type: VARCHAR(255)
+                  constraints:
+                    nullable: false
+                    unique: true
+              - column:
+                  name: data
+                  type: BLOB
+                  constraints:
+                    nullable: false
+              - column:
+                  name: filesize
+                  type: BIGINT
+                  constraints:
+                    nullable: false
+              - column:
+                  name: mime_type
+                  type: VARCHAR(255)
+                  constraints:
+                    nullable: false
+
+
+        - createSequence:
+            sequenceName: recommended_repository_seq
+            startValue: 1
+            incrementBy: 50
+
+        - createTable:
+            tableName: recommended_repository
+            columns:
+              - column:
+                  name: id
+                  type: BIGINT
+                  constraints:
+                    primaryKey: true
+              - column:
+                  name: repository_id
+                  type: VARCHAR(255)
+                  constraints:
+                    nullable: false
+                    unique: true
+
+      rollback:
+        - dropTable:
+            tableName: recommended_repository
+        - dropSequence:
+            sequenceName: recommended_repository_seq
+        - dropTable:
+            tableName: image
+        - dropSequence:
+            sequenceName: image_seq
+        - dropTable:
+            tableName: color_theme
+        - dropSequence:
+            sequenceName: color_theme_seq
--- a/docker/example.env
+++ b/docker/example.env
@@ -60,7 +60,7 @@ DAMAP_AUTH_AFFILIATIONS_CLAIM=affiliation
 # Role name used to identify DAMAP super admins
 # If DAMAP is locally deployed and the variable is not set to damap-super-admin, ADMIN_ROLE constant in the
 # DefaultProfile test profile needs to be changed accordingly, otherwise the tests fail
-DAMAP_AUTH_ADMIN_ROLE_NAME='DAMAP ADMIN'
+DAMAP_AUTH_ADMIN_ROLE_NAME='Damap Admin'
 
 ## =====================
 ## Databases
--- a/src/main/java/org/damap/base/domain/ColorTheme.java
+++ b/src/main/java/org/damap/base/domain/ColorTheme.java
@@ -1,0 +1,27 @@
+package org.damap.base.domain;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import jakarta.persistence.Column;
+import jakarta.persistence.Convert;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Table;
+import java.util.HashMap;
+import java.util.Map;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import org.damap.base.domain.converter.MapToJsonConverter;
+import org.hibernate.Length;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@Entity
+@Table(name = "color_theme")
+public class ColorTheme extends PanacheEntity {
+
+  @Column(name = "exact_colors", nullable = false)
+  private boolean exactColors;
+
+  @Convert(converter = MapToJsonConverter.class)
+  @Column(name = "colors", length = Length.LONG32)
+  private Map<String, String> colors = new HashMap<>();
+}
--- a/src/main/java/org/damap/base/domain/converter/MapToJsonConverter.java
+++ b/src/main/java/org/damap/base/domain/converter/MapToJsonConverter.java
@@ -1,0 +1,35 @@
+package org.damap.base.domain.converter;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Map;
+
+@Converter(autoApply = false)
+public class MapToJsonConverter implements AttributeConverter<Map<String, String>, String> {
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  @Override
+  public String convertToDatabaseColumn(Map<String, String> attribute) {
+    if (attribute == null) return null;
+    try {
+      return MAPPER.writeValueAsString(attribute);
+    } catch (JsonProcessingException e) {
+      throw new IllegalArgumentException("Could not serialize map to string", e);
+    }
+  }
+
+  @Override
+  public Map<String, String> convertToEntityAttribute(String dbData) {
+    if (dbData == null || dbData.isEmpty()) return Collections.emptyMap();
+    try {
+      return MAPPER.readValue(
+          dbData, MAPPER.getTypeFactory().constructMapType(Map.class, String.class, String.class));
+    } catch (IOException e) {
+      throw new IllegalArgumentException("Could not deserialize string to map", e);
+    }
+  }
+}
--- a/src/main/java/org/damap/base/rest/ConfigResource.java
+++ b/src/main/java/org/damap/base/rest/ConfigResource.java
@@ -1,15 +1,21 @@
 package org.damap.base.rest;
 
 import jakarta.annotation.security.PermitAll;
+import jakarta.inject.Inject;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.core.MediaType;
 import java.net.URL;
+import java.util.List;
 import java.util.Optional;
 import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.domain.ColorTheme;
+import org.damap.base.domain.Image;
 import org.damap.base.rest.config.domain.ConfigDO;
 import org.damap.base.rest.config.domain.PersonServiceConfigurations;
+import org.damap.base.rest.theme.service.ColorThemeService;
+import org.damap.base.rest.theme.service.ImageThemeService;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 
 /** ConfigResource class. */
@@ -66,6 +72,10 @@ public class ConfigResource {
   @ConfigProperty(name = "damap.title", defaultValue = "DAMAP Tool")
   String appTitle;
 
+  @Inject ColorThemeService colorThemeService;
+
+  @Inject ImageThemeService imageThemeService;
+
   /**
    * config.
    *
@@ -91,6 +101,12 @@ public class ConfigResource {
     configDO.setFitsServiceAvailable(getFitsServiceAvailability());
     configDO.setLivePreviewAvailable(getGotenbergServiceAvailability());
     configDO.setEthicalReportEnabled(ethicalReportEnabled);
+
+    ColorTheme colorTheme = colorThemeService.getTheme();
+    configDO.setColorTheme(colorTheme);
+
+    List<Image> images = imageThemeService.getImages();
+    configDO.setImages(images);
 
     return configDO;
   }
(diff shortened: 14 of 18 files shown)


src/main/resources/application.yaml
```diff
@@ -2,6 +2,7 @@
 # replace these in the config of your custom project or by overriding these variables
 damap:
   title: ${DAMAP_TITLE:DAMAP Tool}
+  name: damap
```
ValentinFutterer: This is not necessary, the instance id in case of multitenancy mode will be taken from the auth token - the switch between dbs will also happen automatically, no need to code any logic regarding multitenancy.

src/main/java/org/damap/base/domain/Image.java
```diff
@@ -0,0 +1,38 @@
...
+  private String name; // Maps to damap.name config property
+
+  @Column(name = "image_key", nullable = false)
+  private String imageKey;
+
+  @Column(name = "filename", nullable = false, length = 500)
+  private String filename;
+
+  @Basic(fetch = FetchType.LAZY)
+  @Column(name = "data", nullable = false, columnDefinition = "BYTEA")
```
ValentinFutterer: `"BYTEA"` is PostgreSQL specific - I dont know how it interacts with Oracle and Liquibase, but just to be safe i would use something neutral here.

src/main/java/org/damap/base/rest/ThemeResource.java
```diff
@@ -0,0 +1,53 @@
...
+@Consumes(MediaType.APPLICATION_JSON)
+@Authenticated
+@JBossLog
+public class ThemeResource {
+
+  @Inject ThemeService themeService;
+
+  /**
+   * Get the theme for the configured instance name. The name is configured in application.yaml
+   * under damap.name. Public endpoint - no authentication required.
```
DanceaVlad: this endpoint is only used when the admin gets the theme in the editor, so it is not a public endpoint

src/main/java/org/damap/base/domain/Image.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.EqualsAndHashCode;
+import org.hibernate.envers.Audited;
+import org.hibernate.envers.NotAudited;
+
+@Data
+@EqualsAndHashCode(
+    callSuper = true,
+    exclude = {"data"})
+@Entity
+@Audited
```
ValentinFutterer: I dont think that we need to audit the admin tables - so we can remove that in my opinion (for all tables)

src/main/java/org/damap/base/domain/Image.java
```diff
@@ -0,0 +1,45 @@
...
+@Table(
+    name = "image",
+    uniqueConstraints = {
+      @UniqueConstraint(
+          name = "idx_image_name_key",
+          columnNames = {"name", "image_key"})
+    })
```
ValentinFutterer: I would just use `@column(unique=true)` instead of doing the more complicated `@uniqueconstraints`

src/main/java/org/damap/base/domain/Image.java
```diff
@@ -0,0 +1,45 @@
...
+public class Image extends PanacheEntity {
+
+  @Column(name = "name", nullable = false)
+  private String name; // Maps to damap.name config property
```
ValentinFutterer: Not necessary, as discussed

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.6.0_1.yaml
```diff
@@ -0,0 +1,261 @@
...
+        - createTable:
+            tableName: image
+            columns:
+              - column:
+                  name: id
+                  type: BIGINT
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
```
ValentinFutterer: Can be removed, this is taken care of by the access token of the user

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.6.0_1.yaml
```diff
@@ -0,0 +1,261 @@
...
+
+        - createTable:
+            tableName: image
+            columns:
+              - column:
+                  name: id
+                  type: BIGINT
+                  constraints:
+                    primaryKey: true
+                    nullable: false
```
ValentinFutterer: `primarykey: true` already implies `nullable: false`

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.6.0_1.yaml
```diff
@@ -0,0 +1,261 @@
...
+              - column:
+                  name: colors
+                  type: CLOB
+                  constraints:
+                    nullable: true
+
+        - createSequence:
+            sequenceName: image_seq
+            startValue: 1
+            incrementBy: 1
```
ValentinFutterer: You should always use `incrementBy: 50`, see hibernate 6

src/main/java/org/damap/base/rest/config/domain/ImageThemeDO.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.List;
+import lombok.Data;
+import org.damap.base.rest.theme.domain.ImageDO;
+
+/** ImageThemeDO class. */
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ImageThemeDO {
+  private List<ImageDO> images;
+}
```
ValentinFutterer: Isnt this class just a wrapper around a list? If yes, then why not just use the list.

src/main/java/org/damap/base/rest/theme/domain/ImageDO.java
```diff
@@ -0,0 +1,14 @@
...
+
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ImageDO {
+  private Long id;
+  private String imageKey;
+  private String filename;
+  private Long filesize;
+  private String mimeType;
+}
```
ValentinFutterer: Like discussed in person, we should remove the DOs, if they contain exactly the same information as the entities.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,128 @@
...
+  @ConfigProperty(name = "damap.name")
+  String configuredName;
+
+  @Transactional
+  public Image getImageByKey(String imageKey) {
+    if (imageKey == null || imageKey.isBlank()) {
+      throw new IllegalArgumentException("Image key is required");
+    }
+
+    Image image = Image.find("name = ?1 and imageKey = ?2", configuredName, imageKey).firstResult();
```
ValentinFutterer: Just fyi, usually we use repo classes for reusable and more "complex" queries. In this case, the static method is fine imo, since i cant imagine  a situation where we will need to reuse this somewhere and its really simple.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,128 @@
...
+    if (image == null) {
+      throw new IllegalStateException("Image not found for key: " + imageKey);
+    }
+
+    byte[] imageData = image.getData();
+    if (imageData == null || imageData.length == 0) {
+      throw new IllegalStateException("Image data not found");
+    }
```
ValentinFutterer: From the IllegalStateException docs: `Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.`

Wrong exception type here, this is clearly a case for `NotFoundException`.

ValentinFutterer: I wouldnt cover this case here, it should never happen. Input should be sanitized before it gets stored in the database. If it would get into the db, this piece of code would break the whole application, since the fe would never correctly load.

Also the column is not nullable.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,128 @@
...
+  @Transactional
+  public ImageDO uploadImage(String imageKey, FileUpload file) {
+
+    if (imageKey == null || imageKey.isBlank()) {
+      throw new IllegalArgumentException("Image key is required");
+    }
+
+    if (file == null || file.uploadedFile() == null) {
+      throw new IllegalArgumentException("File is required");
+    }
```
ValentinFutterer, line 65: I would just throw `BadRequestException`

src/main/java/org/damap/base/domain/Image.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Column(name = "image_key", nullable = false)
+  private String imageKey;
+
+  @Column(name = "filename", nullable = false, length = 500)
+  private String filename;
```
ValentinFutterer: Is basically unused and can be discarded imo.

src/main/java/org/damap/base/rest/ImageThemeResource.java
```diff
@@ -0,0 +1,49 @@
+package org.damap.base.rest;
```
ValentinFutterer: I would just delete this file and put  the endpoint in AdminResource, to keep similar resources together.

src/main/java/org/damap/base/rest/ImageThemeResource.java
```diff
@@ -0,0 +1,49 @@
...
+  @Transactional
+  public Response getImage(@QueryParam("imageKey") String imageKey) {
+    log.info("GET /api/image?imageKey=" + imageKey);
+
+    Image image = this.imageThemeService.getImageByKey(imageKey);
+
+    String contentType = image.getMimeType();
+    if (contentType == null || contentType.isBlank()) {
+      contentType = MediaType.APPLICATION_OCTET_STREAM;
+    }
```
ValentinFutterer: This is already taken care of when uploading files.

src/main/java/org/damap/base/rest/AdminResource.java
```diff
@@ -51,4 +63,82 @@ public void deleteAppBanner() {
...
+    log.info("PUT /admin/color-theme");
+    return colorThemeService.uploadTheme(colorThemeDO);
+  }
+
+  @PUT
+  @Path("/image-theme")
+  @RolesAllowed("Damap Admin")
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
+  public ImageDO uploadImage(
+      @FormParam("imageKey") String imageKey, @FormParam("file") FileUpload file) {
```
ValentinFutterer: I am vey surprised that this works, since `@FormParam` is supposed to be used for [rest clients](https://quarkus.io/guides/rest-client#form-parameters) 😅
Intellij even complains about that:
image

Here are the [docs](https://quarkus.io/guides/rest#multipart) on how to do the server end.

src/main/java/org/damap/base/rest/AdminResource.java
```diff
@@ -51,4 +63,82 @@ public void deleteAppBanner() {
...
+    log.info("PUT /admin/image-theme");
```
ValentinFutterer: These logs do not tell me much. You should at least add the image key thats stored and write full text instead of the path. (same for the whole file)

src/main/java/org/damap/base/rest/ImageThemeResource.java
```diff
@@ -0,0 +1,49 @@
...
+
+    return Response.ok(image.getData())
+        .header("Content-Type", contentType)
+        .header("Content-Length", image.getFilesize())
+        .build();
+  }
```
ValentinFutterer: Coding the response on your own is fine, but the [quarkus-rest library](https://quarkus.io/guides/rest#multipart-output) has methods to handle this, might be worth it to take a look.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,128 @@
...
+    byte[] imageData = image.getData();
+    if (imageData == null || imageData.length == 0) {
+      throw new IllegalStateException("Image data not found");
+    }
+
+    return image;
+  }
+
+  @Transactional
+  public ImageDO uploadImage(String imageKey, FileUpload file) {
```
ValentinFutterer: Needs some sanitation like allowed filetypes and is data sent with the image (data != null or blank)

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,128 @@
...
+      image.setFilename(file.fileName());
+      image.setData(imageData);
+      image.setFilesize((long) imageData.length);
+      image.setMimeType(mimeType);
+      image.persist();
+
+      return ImageDOMapper.mapEntityToDO(image, new ImageDO());
+
+    } catch (IOException e) {
+      throw new IllegalStateException("Failed to upload image", e);
```
ValentinFutterer: You can just throw an internal server error rexception.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,128 @@
...
+      byte[] imageData = stream.toByteArray();
+
+      Image image = new Image();
+      image.setName(configuredName);
+      image.setImageKey(imageKey);
+      image.setFilename(file.fileName());
+      image.setData(imageData);
+      image.setFilesize((long) imageData.length);
+      image.setMimeType(mimeType);
+      image.persist();
```
ValentinFutterer: The underlying endpoint is  a PUT, so you need to overwrite potential previous files under the same key, otherwise you will violate the unique constraint.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,128 @@
...
+
+      ByteArrayOutputStream stream = new ByteArrayOutputStream();
+      byte[] buffer = new byte[1024];
+      int len;
+      InputStream inputStream = new ByteArrayInputStream(fileData);
+      while ((len = inputStream.read(buffer)) > -1) {
+        stream.write(buffer, 0, len);
+      }
+      stream.flush();
+      byte[] imageData = stream.toByteArray();
```
ValentinFutterer: I dont get understand this is necessary, this reads like C code 😢 
Just `byte[] fileData = Files.readAllBytes(file.uploadedFile());` worked fine for me

src/main/java/org/damap/base/domain/Image.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Lob
+  @Basic(fetch = FetchType.LAZY)
+  @Column(name = "data", nullable = false)
+  @NotAudited
+  private byte[] data;
+
+  @Column(name = "filesize", nullable = false)
+  private Long filesize;
```
ValentinFutterer: Minor: There should probably be a comment to denote the unit used here

src/main/java/org/damap/base/domain/RecommendedRepository.java
```diff
@@ -0,0 +1,27 @@
...
+@Table(name = "recommended_repository")
+public class RecommendedRepository extends PanacheEntityBase {
+
+  @Id
+  @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "recommended_repository_seq")
+  @SequenceGenerator(
+      name = "recommended_repository_seq",
+      sequenceName = "recommended_repository_seq",
+      allocationSize = 1)
+  public Long id;
```
ValentinFutterer: If you use `PanacheEntity` instead of `PanacheEntityBase` you dont need an id field, since it will be generated automatically. Also, liquibase takes care of the sequences.

src/main/java/org/damap/base/domain/RecommendedRepository.java
```diff
@@ -0,0 +1,27 @@
...
+import io.quarkus.hibernate.orm.panache.PanacheEntityBase;
+import jakarta.persistence.*;
+import lombok.Data;
+import lombok.ToString;
+import org.hibernate.envers.Audited;
+
+@Data
+@ToString
+@Entity
+@Inheritance(strategy = InheritanceType.JOINED)
```
ValentinFutterer: Why inheritance???

src/main/java/org/damap/base/domain/RecommendedRepository.java
```diff
@@ -0,0 +1,27 @@
+package org.damap.base.domain;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntityBase;
+import jakarta.persistence.*;
+import lombok.Data;
+import lombok.ToString;
+import org.hibernate.envers.Audited;
+
+@Data
+@ToString
```
ValentinFutterer: `@Data` includes `@ToString`

src/main/java/org/damap/base/domain/RecommendedRepository.java
```diff
@@ -0,0 +1,27 @@
+package org.damap.base.domain;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntityBase;
+import jakarta.persistence.*;
+import lombok.Data;
+import lombok.ToString;
+import org.hibernate.envers.Audited;
+
+@Data
```
ValentinFutterer, line 12: This now also needs an `@EqualsAndHashCode(callSuper = true)` annotation, otherwise it wont use the id field from PanacheEntity

src/main/java/org/damap/base/domain/RecommendedRepository.java
```diff
@@ -0,0 +1,27 @@
+package org.damap.base.domain;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntityBase;
+import jakarta.persistence.*;
```
ValentinFutterer: Minor: * imports are considered bad practice.

src/main/java/org/damap/base/rest/admin/domain/RecommendedRepositoryDO.java
```diff
@@ -0,0 +1,17 @@
+package org.damap.base.rest.admin.domain;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.Size;
+import lombok.Data;
+
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class RecommendedRepositoryDO {
```
ValentinFutterer, line 6: Like discussed, we should remove DOs that dont serve a specific purpose and arent different from their entity counterpart.

src/main/java/org/damap/base/r3data/RepositoriesService.java
```diff
@@ -22,8 +23,8 @@ public class RepositoriesService {
 
   @Inject @RestClient RepositoriesRemoteResource repositoriesRemoteResource;
 
-  @ConfigProperty(name = "damap.repositories.recommendation")
-  Optional<String[]> repositoriesRecommendation;
+  @ConfigProperty(name = "damap.name")
+  String configuredName;
```
ValentinFutterer: As said before, this should be removed. We can get the affiliation from the access token.

src/main/java/org/damap/base/r3data/RepositoriesService.java
```diff
@@ -138,4 +150,35 @@ public String getRepositoryURL(String id) {
...
+    try {
+      Re3Data repo = this.getById(repositoryId);
+      return RepositoryMapper.mapToRepositoryDetails(repo, repositoryId);
+    } catch (Exception e) {
+      log.infov(
+          "Failed to retrieve repository details for ID {0}: {1}", repositoryId, e.getMessage());
+      return null;
+    }
+  }
+
```
ValentinFutterer: This is essentially just a wrapper around getbyId with an additional call to the RepositoryDetails mapper. Its only purpose in the code is to check if an id is valid, by doing a null comparison to the result (see RecommendedRepositoriesService). And in this case, why not just use getById?
The whole method should be removed to keep the code cleaner.

src/main/java/org/damap/base/r3data/RepositoriesService.java
```diff
@@ -138,4 +150,35 @@ public String getRepositoryURL(String id) {
...
+  /**
+   * invalidateRecommendedRepositoriesCache.
+   *
+   * <p>Invalidates the cache for recommended repositories to ensure fresh data is loaded.
+   */
+  @CacheInvalidate(cacheName = "recommendedRepositories")
+  public void invalidateRecommendedRepositoriesCache() {
+    log.info("Invalidating recommended repositories cache");
+  }
```
ValentinFutterer: There is no reason to chache the etRecommeneded method. The getByiD it uses is already cached and caching the database access is overkill.

src/main/java/org/damap/base/r3data/RepositoriesService.java
```diff
@@ -43,18 +44,29 @@ public generated.List getAll() {
   @CacheResult(cacheName = "recommendedRepositories")
```
ValentinFutterer: See the other cache comment.

src/main/java/org/damap/base/r3data/RepositoriesService.java
```diff
@@ -43,18 +44,29 @@ public generated.List getAll() {
...
   public List<RepositoryDetails> getRecommended() {
     List<RepositoryDetails> recommendedRepositories = new ArrayList<>();
-    if (repositoriesRecommendation.isEmpty()) {
+
+    List<RecommendedRepository> dbRecommendations =
+        RecommendedRepository.list("name = ?1 ORDER BY id ASC", configuredName);
```
ValentinFutterer: Can be replaced with `RecommendedRepository.listAll(Sort.by(id).ascending())`

src/main/java/org/damap/base/r3data/RepositoriesService.java
```diff
@@ -43,18 +44,29 @@ public generated.List getAll() {
...
+
+    if (dbRecommendations.isEmpty()) {
       return recommendedRepositories;
     }
```
ValentinFutterer: This is completely superflous and should be removed, for loops are not entered if there is nothing to loop over.

src/main/java/org/damap/base/r3data/RepositoriesService.java
ValentinFutterer: Can and should now be replaced with `RecommendedRepository.listAll(Sort.("id).ascending())`

src/main/java/org/damap/base/r3data/RepositoriesService.java
ValentinFutterer: This code is superflous and should be removed, for loops are not entered if there is nothing to loop over.

src/main/java/org/damap/base/r3data/RepositoriesService.java
```diff
@@ -43,18 +44,29 @@ public generated.List getAll() {
...
-          recommendedRepositories.add(RepositoryMapper.mapToRepositoryDetails(repo, id));
-        } catch (Exception e) {
-          log.infov("Failed to retrieve repository for ID {0}, error: {1}", id, e.getMessage());
-        }
+    for (RecommendedRepository dbRepo : dbRecommendations) {
+      String repositoryId = dbRepo.getRepositoryId();
+
+      if (!repositoryId.startsWith("r3d")) {
+        continue;
+      }
```
ValentinFutterer: This assumes we let bad ids into the db in the first place and should be handled when creating new repositories.

src/main/java/org/damap/base/r3data/RepositoriesService.java
```diff
@@ -43,18 +44,29 @@ public generated.List getAll() {
...
+
+      try {
+        Re3Data repo = this.getById(repositoryId);
+        RepositoryDetails details = RepositoryMapper.mapToRepositoryDetails(repo, repositoryId);
+
+        recommendedRepositories.add(details);
+      } catch (Exception e) {
+        log.infov(
+            "Failed to retrieve repository for ID {0}, error: {1}", repositoryId, e.getMessage());
```
ValentinFutterer, line 54: We can leave this in, the exception handling here will be fixed by the exception handling PR

src/main/java/org/damap/base/r3data/RepositoriesService.java
ValentinFutterer: Caching this methods result is not necessary imo, the getbyId is already cached, which is the important part, since it protects the API.

src/main/java/org/damap/base/r3data/RepositoriesService.java
ValentinFutterer: See the previous comment about chaching here.

src/main/java/org/damap/base/r3data/RepositoriesService.java
```diff
@@ -138,4 +150,35 @@ public String getRepositoryURL(String id) {
...
+
+    try {
+      Re3Data repo = this.getById(repositoryId);
+      return RepositoryMapper.mapToRepositoryDetails(repo, repositoryId);
+    } catch (Exception e) {
+      log.infov(
+          "Failed to retrieve repository details for ID {0}: {1}", repositoryId, e.getMessage());
+      return null;
+    }
+  }
```
ValentinFutterer: This is just a wrapper around getbyID and the mapper function - and its only used once, to check if the result is null or not. We can just remove it to streamline the code and replace its usage with getByID.

src/main/java/org/damap/base/rest/admin/mapper/RecommendedRepositoryDOMapper.java
```diff
@@ -0,0 +1,44 @@
+package org.damap.base.rest.admin.mapper;
+
+import lombok.experimental.UtilityClass;
+import org.damap.base.domain.RecommendedRepository;
+import org.damap.base.rest.admin.domain.RecommendedRepositoryDO;
+
+@UtilityClass
+public class RecommendedRepositoryDOMapper {
```
ValentinFutterer: Like discussed before, DOs are not necessary, so we can get rid of the mapper.

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+    return RecommendedRepositoryDOMapper.mapEntityToDO(repository, new RecommendedRepositoryDO());
+  }
+
+  @Transactional
+  public RecommendedRepositoryDO updateRecommendedRepository(
+      Long id, @Valid RecommendedRepositoryDO recommendedRepositoryDO) {
+    RecommendedRepository repository = RecommendedRepository.findById(id);
+    if (repository == null) {
+      throw new IllegalStateException("Repository with ID '" + id + "' does not exist");
+    }
```
ValentinFutterer: This is used in a PUT request, the PUT HTTP specification states that the resource should be created, if it doesnt exist. PUT is idempotent, so every request should leave the server in the same state.

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+
+    RecommendedRepository existing =
+        RecommendedRepository.find(
+                "repositoryId = ?1 AND id != ?2", recommendedRepositoryDO.getRepositoryId(), id)
+            .firstResult();
+
+    if (existing != null) {
+      throw new IllegalStateException(
+          "Repository with ID '" + recommendedRepositoryDO.getRepositoryId() + "' already exists");
+    }
```
ValentinFutterer, line 81: Same thing as above, its a PUT

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+
+    if (!isValidRepositoryId(recommendedRepositoryDO.getRepositoryId())) {
+      throw new IllegalArgumentException(
```
ValentinFutterer: Bad request would fit better for now, until we redo the error handling

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+    RecommendedRepository repository = new RecommendedRepository();
+    repository.setRepositoryId(recommendedRepositoryDO.getRepositoryId());
+    repository.persist();
+    repositoriesService.invalidateRecommendedRepositoriesCache();
+
+    return RecommendedRepositoryDOMapper.mapEntityToDO(repository, new RecommendedRepositoryDO());
+  }
+
+  @Transactional
+  public RecommendedRepositoryDO updateRecommendedRepository(
```
ValentinFutterer: I just realized that the method is compeltely unused since we dont have an update in the frontend (and no need for it). I am going to leave the other comments in for educational purposes but the method should be removed.

src/main/java/org/damap/base/rest/AdminResource.java
```diff
@@ -51,4 +63,82 @@ public void deleteAppBanner() {
...
+  public RecommendedRepositoryDO createRecommendedRepository(
+      @Valid RecommendedRepositoryDO recommendedRepositoryDO) {
+    log.info("POST /admin/recommended-repositories");
+    return this.recommendedRepositoryService.createRecommendedRepository(recommendedRepositoryDO);
+  }
+
+  @PUT
+  @Path("/recommended-repositories/{id}")
+  @RolesAllowed("Damap Admin")
+  public RecommendedRepositoryDO updateRecommendedRepository(
```
ValentinFutterer: Unused and can be removed

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+
+  @Transactional
+  public void deleteRecommendedRepository(Long id) {
+    RecommendedRepository repository = RecommendedRepository.findById(id);
+    if (repository == null) {
+      throw new IllegalStateException("Repository with ID '" + id + "' does not exist");
+    }
+
+    repository.delete();
+    repositoriesService.invalidateRecommendedRepositoriesCache();
```
ValentinFutterer: See previous caching comments.

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+  public RecommendedRepositoryDO createRecommendedRepository(
+      @Valid RecommendedRepositoryDO recommendedRepositoryDO) {
+    RecommendedRepository existing =
+        RecommendedRepository.find("repositoryId = ?1", recommendedRepositoryDO.getRepositoryId())
+            .firstResult();
+
+    if (existing != null) {
+      throw new IllegalStateException(
+          "Repository with ID '" + recommendedRepositoryDO.getRepositoryId() + "' already exists");
+    }
```
ValentinFutterer, line 81: Should throw a 409 Conflict error

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+
+    // validate and only allow via Re3Data IDs
+    if (!isValidRepositoryId(recommendedRepositoryDO.getRepositoryId())) {
+      throw new IllegalArgumentException(
+          "Invalid or unknown Re3Data repository ID: " + recommendedRepositoryDO.getRepositoryId());
+    }
```
ValentinFutterer, line 87: Its a little doublechecking, but it protects against API misuse.
See other comments for the exception type.

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+
+    RecommendedRepository repository = new RecommendedRepository();
+    repository.setRepositoryId(recommendedRepositoryDO.getRepositoryId());
+    repository.persist();
+    repositoriesService.invalidateRecommendedRepositoriesCache();
```
ValentinFutterer: See cache comments

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+        .toList();
+  }
+
+  public RecommendedRepositoryDO getRecommendedRepository(Long id) {
+    RecommendedRepository repository = RecommendedRepository.findById(id);
+    if (repository == null) {
+      return null;
+    }
+    return RecommendedRepositoryDOMapper.mapEntityToDO(repository, new RecommendedRepositoryDO());
+  }
```
ValentinFutterer:
```suggestion
  public RecommendedRepository getRecommendedRepository(Long id) {
    return RecommendedRepository.findById(id);
  }
```

And at this point you could just cut the method, since its just a wrapper around one line.

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+
+  public List<RecommendedRepositoryDO> getActiveRecommendedRepositories() {
+    List<RecommendedRepository> repositories = RecommendedRepository.list("ORDER BY id ASC");
+
+    return repositories.stream()
+        .map(
+            repo ->
+                RecommendedRepositoryDOMapper.mapEntityToDO(repo, new RecommendedRepositoryDO()))
+        .toList();
+  }
```
ValentinFutterer: Unused, gets the axe 🪓

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+
+  public List<RecommendedRepositoryDO> getRecommendedRepositories() {
+    List<RecommendedRepository> repositories = RecommendedRepository.list("ORDER BY id ASC");
+
+    return repositories.stream()
+        .map(
+            repo ->
+                RecommendedRepositoryDOMapper.mapEntityToDO(repo, new RecommendedRepositoryDO()))
+        .toList();
+  }
```
ValentinFutterer:
```suggestion
  public List<RecommendedRepositoryDO> getRecommendedRepositories() {
    return RecommendedRepository.listAll().ascending();
  }
```

And at this point its just a wrapper around one line and we can remove the method.

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,124 @@
...
+@JBossLog
+public class RecommendedRepositoryService {
+
+  @Inject RepositoriesService repositoriesService;
+
+  private boolean isValidRepositoryId(String repositoryId) {
+    if (repositoryId == null) {
+      return false;
+    }
+    if (!repositoryId.matches("^r3d\\d{6,}$")) {
```
ValentinFutterer: I didnt find anythin about re3data id schema, seems like most of them have 9 digits. But minimum 6 is fine, to catch potential outliers. maybe write a comment to say that this is not the official pattern.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.6.0_1.yaml
```diff
@@ -0,0 +1,261 @@
...
+                  name: id
+                  type: BIGINT
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: name
+                  type: VARCHAR(255)
+                  constraints:
+                    nullable: false
```
ValentinFutterer: As discussed, this can be removed. This also implies some work in rewriting the service. urrently, we can only have one color theme stored, so the logic should reflect that in some way (Get with no id, only PUT for modification, no POST)

src/main/java/org/damap/base/rest/theme/service/ColorThemeService.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @Transactional
+  public void deleteTheme() {
+    ColorTheme colorTheme = this.getThemeByName();
+    if (colorTheme == null) {
+      throw new IllegalStateException("Theme with name '" + configuredName + "' does not exist");
+    }
+    colorTheme.delete();
+  }
```
ValentinFutterer: Completely unused and should be removed

ValentinFutterer (commented):
Sorry for the 60 comment bomb 😅 
Most of it should be fast fixes

src/main/java/org/damap/base/domain/Image.java
DanceaVlad: Done

src/main/java/org/damap/base/domain/Image.java
DanceaVlad: Removed since we also no longer need the name property

src/main/java/org/damap/base/domain/Image.java
DanceaVlad: Done

src/main/java/org/damap/base/domain/Image.java
DanceaVlad: Discarded

src/main/java/org/damap/base/domain/Image.java
DanceaVlad: Done

src/main/java/org/damap/base/domain/RecommendedRepository.java
DanceaVlad: Done

src/main/java/org/damap/base/domain/RecommendedRepository.java
DanceaVlad: Done

src/main/java/org/damap/base/domain/RecommendedRepository.java
DanceaVlad: Done

src/main/java/org/damap/base/domain/RecommendedRepository.java
DanceaVlad: used normal PanacheEntity

src/main/java/org/damap/base/domain/RecommendedRepository.java
DanceaVlad: Deleted

src/main/java/org/damap/base/r3data/RepositoriesService.java
DanceaVlad: Removed

src/main/java/org/damap/base/r3data/RepositoriesService.java
DanceaVlad: Done

DanceaVlad: Done

src/main/java/org/damap/base/r3data/RepositoriesService.java
DanceaVlad: Replaced

DanceaVlad: Replaced

src/main/java/org/damap/base/rest/config/domain/ImageThemeDO.java
DanceaVlad: Removed the DO

src/main/java/org/damap/base/rest/theme/domain/ImageDO.java
DanceaVlad: The DO is just the metadata. The entity contains the file aswell. You could argue I could still remove the DO and return the Image entity with a @JsonIgnore annotation for the data field.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Done

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Done

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Done

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Done

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Done

DanceaVlad: Done

src/main/java/org/damap/base/rest/AdminResource.java
DanceaVlad: Done

src/main/java/org/damap/base/rest/ImageThemeResource.java
DanceaVlad: Done

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.6.0_1.yaml
DanceaVlad: Done

src/main/resources/application.yaml
DanceaVlad: Removed

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_1.yaml
```diff
@@ -0,0 +1,100 @@
+databaseChangeLog:
```
ValentinFutterer, line 1: We are on version 4.6.1, so the file should be renamed to keep with conventions

src/main/java/org/damap/base/r3data/RepositoriesService.java
ValentinFutterer: This if check is still here

src/main/java/org/damap/base/rest/AdminResource.java
ValentinFutterer: As discussed, I think we should follow the docs (both work, but its saver to stick to them) 
```suggestion
  @Consumes(MediaType.MULTIPART_FORM_DATA)
  public ImageDO uploadImage(
      @RestForm("imageKey") String imageKey, @RestForm("image") FileUpload file) {
```

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
ValentinFutterer: .

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,140 @@
...
+
+    return image;
+  }
+
+  @Transactional
+  public ImageDO uploadImage(String imageKey, FileUpload file) {
+
+    if (imageKey == null || imageKey.isBlank()) {
+      throw new IllegalArgumentException("Image key is required");
+    }
```
ValentinFutterer, line 61: Should also be bad request

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,140 @@
...
+    if (file == null || file.uploadedFile() == null) {
+      throw new BadRequestException("File is required");
+    }
+
+    try {
+      byte[] fileData = Files.readAllBytes(file.uploadedFile());
+
+      if (fileData == null || fileData.length == 0) {
+        throw new IllegalArgumentException("No image data provided");
+      }
```
ValentinFutterer: Same bad request as above.
Also, fileData can never be null, so this check is not needed.
There should be something that checks file size, otherwise i could just upload a 50Gb file with a png ending

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,140 @@
...
+      if (mimeType == null || mimeType.isBlank()) {
+        mimeType = guessContentType(file.fileName());
+      }
+
+      if (!isAllowedImageType(mimeType)) {
+        throw new IllegalArgumentException("Unsupported image type: " + mimeType +
+          ". Allowed types: PNG, JPG, JPEG, GIF, SVG, WebP, ICO");
+      }
+
+      byte[] imageData = fileData;
```
ValentinFutterer: This is just a reassignemnt without purpose.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,140 @@
...
+
+      String mimeType = file.contentType();
+      if (mimeType == null || mimeType.isBlank()) {
+        mimeType = guessContentType(file.fileName());
+      }
+
+      if (!isAllowedImageType(mimeType)) {
+        throw new IllegalArgumentException("Unsupported image type: " + mimeType +
+          ". Allowed types: PNG, JPG, JPEG, GIF, SVG, WebP, ICO");
+      }
```
ValentinFutterer, line 86: Should be unsupported media type exception

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,140 @@
...
+        image.setData(imageData);
+        image.setFilesize((long) imageData.length);
+        image.setMimeType(mimeType);
+        log.info("Updated existing image for key: " + imageKey);
+      } else {
+        image = new Image();
+        image.setImageKey(imageKey);
+        image.setData(imageData);
+        image.setFilesize((long) imageData.length);
+        image.setMimeType(mimeType);
```
ValentinFutterer: You should use a constructor here

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,140 @@
...
+
+    String name = filename.toLowerCase();
+    if (name.endsWith(".png")) return "image/png";
+    if (name.endsWith(".jpg") || name.endsWith(".jpeg")) return "image/jpeg";
+    if (name.endsWith(".svg")) return "image/svg+xml";
+    if (name.endsWith(".gif")) return "image/gif";
+    if (name.endsWith(".webp")) return "image/webp";
+    if (name.endsWith(".ico")) return "image/x-icon";
+    return MediaType.APPLICATION_OCTET_STREAM;
+  }
```
ValentinFutterer: Check out the FileTypeDetector class, it probably covers this usecase or does so with minimal effort and we can later maybe reuse to replace the fits service

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @Transactional
+  public List<ImageDO> getImages() {
+    List<Image> images = Image.findAll().list();
+    List<ImageDO> imageDOs =
+        images.stream()
+            .map(img -> ImageDOMapper.mapEntityToDO(img, new ImageDO()))
+            .collect(Collectors.toList());
+
+    return imageDOs;
```
ValentinFutterer: Minor:
```suggestion
    return images.stream()
            .map(img -> ImageDOMapper.mapEntityToDO(img, new ImageDO()))
            .collect(Collectors.toList());
```

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  @Transactional
+  public void deleteImage(String imageKey) {
+    if (imageKey == null || imageKey.isBlank()) {
+      throw new IllegalArgumentException("Image key is required");
+    }
```
ValentinFutterer, line 122: I am not gonna write it anymore, but it should be bad request

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
ValentinFutterer: Top one not done

src/main/java/org/damap/base/rest/ImageThemeResource.java
ValentinFutterer: Not done

src/main/java/org/damap/base/rest/ImageThemeResource.java
ValentinFutterer: Did you check this?

src/main/java/org/damap/base/rest/AdminResource.java
```diff
@@ -48,7 +72,112 @@ public BannerDO updateAppBanner(@Valid BannerDO bannerDO) {
...
+
+  @GET
+  @Path("/image")
+  @PermitAll
+  @Produces({
+    "image/png",
+    "image/jpeg",
+    "image/gif",
+    "image/svg+xml",
+    MediaType.APPLICATION_OCTET_STREAM
```
ValentinFutterer: Not the same list as in the service, this should probably be an enum somewhere to keep it consistent.
Or a static class with a constant private list or something similar.

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,102 @@
...
+  public RecommendedRepository createRecommendedRepository(
+      @Valid RecommendedRepository recommendedRepository) {
+    RecommendedRepository existing =
+        RecommendedRepository.find("repositoryId = ?1", recommendedRepository.getRepositoryId())
+            .firstResult();
+
+    if (existing != null) {
+      throw new IllegalStateException(
+          "Repository with ID '" + recommendedRepository.getRepositoryId() + "' already exists");
+    }
```
ValentinFutterer, line 81: Should be a conflict exception

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,102 @@
...
+
+    RecommendedRepository conflictingRepository =
+        RecommendedRepository.find(
+                "repositoryId = ?1 AND id != ?2", recommendedRepository.getRepositoryId(), id)
+            .firstResult();
+
+    if (conflictingRepository != null) {
+      throw new IllegalStateException(
+          "Repository with ID '" + recommendedRepository.getRepositoryId() + "' already exists");
+    }
```
ValentinFutterer: Not db ID, repository id

.java-version
```diff
@@ -0,0 +1 @@
+17
```
ValentinFutterer: ❓

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
ValentinFutterer: This is still the case please fix

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
ValentinFutterer: .

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
ValentinFutterer: Same thing again, please take care to read and respond to all my comments, this method and the accompyaning endpoint are still in code and they are completely unused.

src/main/java/org/damap/base/rest/AdminResource.java
ValentinFutterer: Same again, this can be removed as its completely unused

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
ValentinFutterer: .

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
ValentinFutterer: .

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,102 @@
...
+    try {
+      repositoriesService.getById(repositoryId);
+      return true;
+    } catch (Exception e) {
+      return false;
+    }
+  }
+
+  public List<RecommendedRepository> getRecommendedRepositories() {
+    return RecommendedRepository.list("ORDER BY id ASC");
```
ValentinFutterer:
```suggestion
    return RecommendedRepository.listAll().ascending();
```

src/main/java/org/damap/base/rest/theme/service/ColorThemeService.java
ValentinFutterer: Its still in the code

ValentinFutterer (commented):
PR description is out of date and inaccurate and should be fixed for archival purposes.
Documentation also needs an update (markdown in the repo and guide on website)
Also, most of the illegal state or illegal argument exceptions can be replaced with bad request or not found, i stopped writing comments about that halfway through

In the future, please try to respond to all the comments - its tedious to write the same thing twice.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Done

src/main/java/org/damap/base/rest/ImageThemeResource.java
DanceaVlad: Done

src/main/java/org/damap/base/rest/AdminResource.java
DanceaVlad: Done

src/main/java/org/damap/base/rest/ImageThemeResource.java
DanceaVlad: Removed this endpoint, now the config returns the images on startup

src/main/java/org/damap/base/r3data/RepositoriesService.java
DanceaVlad: Removed

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
DanceaVlad: There is no more PUT endpoint to use this

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
DanceaVlad: Done

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
DanceaVlad: removed the whole PUT flow

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
DanceaVlad: Changed to `BadRequestException`

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
DanceaVlad: Commented

src/main/java/org/damap/base/rest/theme/service/ColorThemeService.java
DanceaVlad: Removed

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_1.yaml
DanceaVlad: Changed to 4.6.1_1

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Removed the if check, changed to RestForm, added the comment and changed to BadRequest

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Added a check for file size, changed to BadRequest or NotFound for the whole file.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Removed completely

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Removed

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Changed

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Removed entirely

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
DanceaVlad: Done

.java-version
DanceaVlad: Removed

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
DanceaVlad: Done

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
DanceaVlad: Done with a `ClientErrorException` and `Response.Status.Conflict`

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+  public List<RecommendedRepository> getRecommendedRepositories() {
+    return RecommendedRepository.listAll(Sort.by("id").ascending());
+  }
+
+  public RecommendedRepository getRecommendedRepository(Long id) {
+    return RecommendedRepository.findById(id);
+  }
```
ValentinFutterer: Unused and should be removed
```suggestion
```

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,184 @@
...
+
+  @Transactional
+  public Image uploadImage(String imageKey, FileUpload file) {
+    if (imageKey == null || imageKey.isBlank()) {
+      throw new BadRequestException("Image key is required");
+    }
+
+    if (file == null || file.uploadedFile() == null) {
+      throw new BadRequestException("File is required");
+    }
```
ValentinFutterer, line 65: There is a 413 exception for this

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,184 @@
...
+
+    // MIME from content
+    try {
+      mimeType = Files.probeContentType(file.uploadedFile());
+      if (mimeType != null && !mimeType.isBlank()) {
+        return mimeType;
+      }
+    } catch (Exception e) {
+    }
```
ValentinFutterer: Code smells - catching a generic exception and then doing nothing with it

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,184 @@
...
+    if (imageKey == null || imageKey.isBlank()) {
+      throw new BadRequestException("Image key is required");
+    }
+
+    if (Image.delete("imageKey", imageKey) == 0) {
+      throw new NotFoundException("Image not found for key: " + imageKey);
+    }
+  }
+
+  private String detectMimeType(FileUpload file, byte[] imageData) {
```
ValentinFutterer: `imageData` is unused

```suggestion
  private String detectMimeType(FileUpload file) {
```

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,184 @@
...
+    // MIME from header
+    String mimeType = file.contentType();
+    if (mimeType != null && !mimeType.isBlank()) {
+      return mimeType;
+    }
+
+    // MIME from content
+    try {
+      mimeType = Files.probeContentType(file.uploadedFile());
```
ValentinFutterer: Since you didnt install a FileTypeDetector, this might not do anthing, dependant on the base implementation. You should check if it can be removed.

ValentinFutterer (commented):
Only a few comments this time, thanks for adressing all of my previous ones.
I will do the integration testing together with the frontend review, in the meantime, could you please attend to the tests that are failing and rebase.

Seeing forward to finally merging it

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Changed

src/main/java/org/damap/base/rest/admin/service/RecommendedRepositoryService.java
DanceaVlad: Removed

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
DanceaVlad: Added a comment about it

src/main/resources/application.yaml
```diff
@@ -28,10 +28,6 @@ damap:
     username: ${DAMAP_DATASOURCE_USERNAME:damap}
     password: ${DAMAP_DATASOURCE_PASSWORD:pw4damap}
     dialect: ${DAMAP_DATASOURCE_DIALECT:org.hibernate.dialect.PostgreSQLDialect}
-  repositories:
-    # Zenodo, Open Science Framework, DRYAD, Mendeley Data
-    # can also be [] for no repositories recommendation
-    recommendation: ['r3d100010468', 'r3d100011137', 'r3d100000044', 'r3d100011868'] # your-recommended-repositories
```
ValentinFutterer, line 30: Test profile still has repositories

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_1.yaml
ValentinFutterer, line 1: Needs to be brought to 4.7.0

ValentinFutterer (commented):
Can you also please rebase while you are at it :)

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,184 @@
...
+      }
+
+      if (imageData.length > MAX_FILE_SIZE) {
+        throw new ClientErrorException(
+            "File size exceeds maximum allowed size of " + (MAX_FILE_SIZE / 1024 / 1024) + "MB",
+            Response.Status.REQUEST_ENTITY_TOO_LARGE);
+      }
+
+      String mimeType = detectMimeType(file);
+
```
BlueTechie, line 81: Instead of the custom logic to detect the Mime type of the file, we can simplify this by importing the Apache Tika library and replacing the line above by:
```java
String mimeType = tika.detect(imageData);
```

This would not only simplify the code (you can then delete the `detectMimeType(...)` and `detectMimeTypeFromExtension(...)` functions), but also make it more secure: 
The custom code trusts that the content-type header, resp. the file extension in the filename, corresponds to the actual file type. However this isn't necessarily the case, since they're both controlled by the user, e.g. I can rename `virus.exe` to `logo.png` or change the content-type to PNG and then upload a malicious file to the server. 

With the Tika alternative, the Mime type is given by the magic bytes of the file, which makes the above impossible.

src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -0,0 +1,184 @@
...
+@ApplicationScoped
+@JBossLog
+public class ImageThemeService {
+
+  public enum AllowedImageType {
+    PNG("image/png"),
+    JPEG("image/jpeg"),
+    JPG("image/jpg"),
+    GIF("image/gif"),
+    SVG("image/svg+xml"),
```
BlueTechie, line 28: SVGs are XML files which can load external code (for instance JavaScript) and this means we have to be careful here. Allowing SVG uploads opens the door to all sorts of shenanigans: XSS attacks where a malicious logo-SVG is loaded for DAMAP users and then steals credentials, [SVGs that can crash the backend server](https://medium.com/@instatunnel/billion-laughs-attack-the-xml-that-brings-servers-to-their-knees-f83ba617caa4) etc...

Here are a few options to fix this or at least reduce the risks:
- Disallow SVGs uploads outright.
- Allow SVG uploads, but then convert the SVG file to a PNG/JPEG and store it that way. This way we don't store user-controlled code on the server nor do we serve it to the frontend.
- Sanitize SVGs before storing them in the database. I tried looking for a Java library for this but it seems that there are no good options. We could in theory do custom sanitization, but this would probably add too much complexity and would be difficult to get right.
- Change the Content Security Policy so that the frontend doesn't load external scripts. This is not too much effort but maybe a bit harder to debug since the dev environment is different than the deployment one (where `src 'self'` covers backend and frontend).

Interesting read on the subject/source: 
https://blobfolio.com/2017/when-a-stranger-calls-sanitizing-svgs/
https://stackoverflow.com/questions/10557137/are-user-uploaded-svgs-an-xss-risk-how-can-you-sanitize-an-svg

I propose discussing this jointly with @ValentinFutterer to see which option works best.

BlueTechie (changes requested):
I focused my review on the image upload functionality, especially with regards to security. 
- Allowing SVGs unfortunately introduces vulnerabilities, which I explain below. 
-  We currently use custom Mime type detection, which can be made simpler and more secure with a library.

Otherwise from what I saw, this PR introduces nice changes which are intuitive to use 👍

PR #393, opened 2025-06-12
Helm chart for Kubernetes/OpenShift deployment

bitmotivator:
This PR adds a [Helm chart](https://helm.sh/docs/topics/charts/) for deploying DAMAP on Kubernetes/OpenShift.

# TODO

- [x] Documentation
- [x] Flag to skip deploying Keycloak and use a preexisting OIDC server instead
- [x] CI pipeline to test Helm chart

 ### Update [2025-11-06T19:25Z]
 Chart needs to be updated due to the following PRs:
 - [x] #446 
 - [x] #424 
 
### Update [2026-02-06T15:08Z]
Chart needs to be updated due to the following PRs:
 - [x] #459 
 
 ### Update [2026-02-10T17:26Z]
 Testing Checklist:
 - [ ] Vanilla k8s local
 - [ ]  Vanilla k8s online
 - [ ] OpenShift online (staging.damap.org)

PR #428, opened 2025-09-17
Helm Chart: Fixes & Stable deployment (Vanilla k8s & Openshift)

SotosTsepe:
## Description
This PR introduces minor fixes to the Helm chart and adds out-of-the-box compatibility for both vanilla Kubernetes and OpenShift.
It is primarily a preparatory step to reduce the workload for the upcoming PR #393.

## Changes (most recent last)
Each number represents a commit which will be squashed:
1. Nest Keycloak PVs and PVCs under a single if block.
Define Keycloak DB volumes conditionally.
Run DB containers as root on vanilla Kubernetes.

2. Fix healthchecks for Keycloak DB and DAMAP DB.
Comment out `securityContext` for Keycloak DB.
Add UID/GID for Gotenberg container (vanilla K8s).
Hardcode passwords for DAMAP & Keycloak DBs in `values.yaml` (with inline reasoning note).

3. Minor formatting fixes (mostly newlines).
Set wrong DAMAP_AUTH_BACKEND_URL temporarily for testing.

4. Run Keycloak & DAMAP DB as non-root on OpenShift.
Improve Helm conditionals readability (indentation).
Set image pull policy for DAMAP frontend & backend if using `next` or `latest` tag.
Revert `DAMAP_AUTH_BACKEND_URL` to the correct value.
Conditionally define volume mount for DAMAP DB container.

5. `create-test-cluster.sh`: Fix host format when writing to `hosts.toml`.

6. Nest DAMAP PVs and PVCs under a single if block (similar to 1).
Conditionally define DAMAP DB Deployment volume (similar to 1, fragmented across 4 & 6).

7. Add environment variables to DAMAP backend for verbose logs when in debug mode.

8. Remove trailing slash from Keycloak Realm Import origins (fixes request failures).

9. Improve spacing, indentation, and hyphen consistency in Helm statements.

10. **IMPORTANT**:
Ensure Dynamic hostname resolving for Keycloak if running on localhost.
See: https://github.com/keycloak/keycloak/issues/17634 and https://github.com/keycloak/keycloak/issues/27660#issuecomment-2117997684.
Remove hostname definition from CLI args (already set via env var).

11. Deploy with PURE mock integration and ORCID enabled (with sample values).

12. Display 2 recommended projects for the default user.

13. Ensure Keycloak hostname always includes the protocol in every mode.

14. Replace randAscii with randAlphaNum for passwords. 
Previously, passwords were generated using randAscii, which occasionally produced characters like backslashes or double quotes. These special characters could break Helm rendering, causing the Helm command to fail
intermittently.
 
#### Database
- [X] Tested with PostgreSQL

#### Deployment
- [X] Tested on vanilla Kubernetes (locally)
- [X] Tested on Openshift (online domain)

#### UI tests
- [X] Tested all UI routes manually

Pending TODOs:
- Fix GH actions Pipeline (will be fixed on #393)
- Add docs (will be addressed on #393)


SotosTsepe:
cc @bitmotivator @ValentinFutterer

PR #431, opened 2025-10-13
Add multitenancy feature

### LINKED ISSUE #415: Native multi-tenancy
In order to make DAMAP cloud possible, we need native multi-tenancy in DAMAP. This means we should only ever need to deploy one DAMAP instance, not multiple.

ValentinFutterer:
## Description

Feature/Config/Documentation

#### What does this PR do?
Introduces the multitenancy feature to DAMAP - how to use it is detailed in the documentation, it invloves setting two environmnt variables and creating an external tenant config file.

#### Code review focus
To test locally, paste this block in your external config file and change the bootstrap url in the application.yaml to `jdbc:postgresql://localhost:8088/damap`
```
  # testing also change bootstrap url to jdbc:postgresql://localhost:8088/damap
  damap:
    title: "DAMAP (Development)"
    origins: http://localhost:4200
    auth:
      backend:
        url: http://localhost:8087/realms/damap
    mock-endpoint-url: http://localhost:8091
    fits-url: http://localhost:8089/fits
    gotenberg-url: http://localhost:3000
    elsevier-pure-endpoint-url: "http://localhost:9999/fake-pure-api"
```

Build it locally using `mvn clean package "-Dquarkus.package.jar.type=mutable-jar" "-Dquarkus.profile=multitenant" -DskipTests`, reaugment with `java "-Dquarkus.profile=multitenant" "-Dquarkus.launch.rebuild=true" -jar target/quarkus-app/quarkus-run.jar` and finally start with `java -jar target/quarkus-app/quarkus-run.jar`.

To test it in docker / kubernetes, follow the documentation provided in the PR..

closes GH-415

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,6 +49,9 @@ ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jb
     LANG="en_US.UTF-8" \
     LANGUAGE="en_US:en"
 
+# to start in multitenant mode, pass -Dquarkus.profile=multitenant
+ENV QUARKUS_PROFILES=""
+
 WORKDIR /quarkus-app
 
 COPY --from=builder /home/app/target/quarkus-app/ ./
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,9 +49,6 @@ ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jb
     LANG="en_US.UTF-8" \
     LANGUAGE="en_US:en"
 
-# to start in multitenant mode, pass -Dquarkus.profile=multitenant
-ENV QUARKUS_PROFILES=""
-
 WORKDIR /quarkus-app
 
 COPY --from=builder /home/app/target/quarkus-app/ ./
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -7,7 +7,7 @@
 set -euo pipefail
 
 echo "$(date +"%Y-%m-%d %H:%M:%S,%3N") INFO  [docker-entrypoint.sh] Running Quarkus re-augmentation."
-java ${JAVA_OPTIONS} -Dquarkus.launch.rebuild=true -jar quarkus-run.jar
+java ${JAVA_OPTIONS} ${QUARKUS_PROFILES} -Dquarkus.launch.rebuild=true -jar quarkus-run.jar
 
 echo "$(date +"%Y-%m-%d %H:%M:%S,%3N") INFO  [docker-entrypoint.sh] Starting application."
 exec java ${JAVA_OPTIONS} -jar quarkus-run.jar
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -7,7 +7,7 @@
 set -euo pipefail
 
 echo "$(date +"%Y-%m-%d %H:%M:%S,%3N") INFO  [docker-entrypoint.sh] Running Quarkus re-augmentation."
-java ${JAVA_OPTIONS} ${QUARKUS_PROFILES} -Dquarkus.launch.rebuild=true -jar quarkus-run.jar
+java ${JAVA_OPTIONS} -Dquarkus.launch.rebuild=true -jar quarkus-run.jar
 
 echo "$(date +"%Y-%m-%d %H:%M:%S,%3N") INFO  [docker-entrypoint.sh] Starting application."
 exec java ${JAVA_OPTIONS} -jar quarkus-run.jar
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -211,4 +211,4 @@ rest:
 
     config:
       # should point to a (out of classpath) config file like this /path/to/config/file.yaml
-      locations: ${DAMAP_MULTITENANT_CONFIG:}
+      locations: /opt/damap/tenants.yaml
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -210,5 +210,5 @@ rest:
       multitenant: DATABASE
 
     config:
-      # should point to a (out of classpath) config file like this file:///path/to/config/file.yaml
+      # should point to a (out of classpath) config file like this /path/to/config/file.yaml
       locations: ${DAMAP_MULTITENANT_CONFIG:}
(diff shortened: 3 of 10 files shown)


ValentinFutterer:
Note: I should also add a config option, where you can block access for non-admins, so admins have time enough to add customizations, before going live.

Dockerfile
```diff
@@ -42,6 +42,9 @@ ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jb
     LANG="en_US.UTF-8" \
     LANGUAGE="en_US:en"
 
+# to start in multitenant mode, pass -Dquarkus.profile=multitenant
+ENV QUARKUS_PROFILES=""
```
SotosTsepe: This is redundant; Quarkus already provides the `QUARKUS_PROFILE` runtime property with the same effect.

docker/docker-entrypoint.sh
```diff
@@ -7,7 +7,7 @@
 set -euo pipefail
 
 echo "$(date +"%Y-%m-%d %H:%M:%S,%3N") INFO  [docker-entrypoint.sh] Running Quarkus re-augmentation."
-java ${JAVA_OPTIONS} -Dquarkus.launch.rebuild=true -jar quarkus-run.jar
+java ${JAVA_OPTIONS} ${QUARKUS_PROFILES} -Dquarkus.launch.rebuild=true -jar quarkus-run.jar
```
SotosTsepe: Also redundant, same reason as above.

docker/docker-entrypoint.sh
```diff
@@ -7,7 +7,7 @@
...
 
 echo "$(date +"%Y-%m-%d %H:%M:%S,%3N") INFO  [docker-entrypoint.sh] Starting application."
-exec java ${JAVA_OPTIONS} -jar quarkus-run.jar
+exec java ${JAVA_OPTIONS} -jar quarkus-run.jar
```
SotosTsepe: Don't forget NLs :smiley:

src/main/resources/application.yaml
```diff
@@ -185,3 +184,31 @@ rest:
...
+    hibernate-orm:
+      datasource: bootstrap # hibernate needs a bootstrap db to build the persistence unit - its later unused
+      sql-load-script: no-file
+      schema-management:
+        strategy: none
+      multitenant: DATABASE
+
+    config:
+      # should point to a (out of classpath) config file like this file:///path/to/config/file.yaml
+      locations: ${DAMAP_MULTITENANT_CONFIG:}
```
SotosTsepe: NL missing here as well

SotosTsepe (commented):
Good effort overall; tested in Docker Compose as-is and works.

consider fixing the below comments and adding to the compose file in the `environment` section:

```bash
DAMAP_MULTITENANT_CONFIG: ${DAMAP_MULTITENANT_CONFIG:-/opt/damap/tenants.yml}
QUARKUS_PROFILE: ${QUARKUS_PROFILE:-}
```
 and ofc a volume definition:
 `- ./tenants.yml:/${DAMAP_MULTITENANT_CONFIG:-/opt/damap/tenants.yml}:ro`
 
 This will make it very dynamic, and each user can apply the appropriate customizations/profiles from the `.env` file instead.
 
 This approach was also tested and produced the same results.
 
Finally, we can automate the creation of the DBs even further by handling them from the PostgreSQL or DAMAP side with a small script. Can assist with that, but it's optional and up to you.

Dockerfile
ValentinFutterer: Good catch 👍

docker/docker-entrypoint.sh
ValentinFutterer: Readded it 😃

ValentinFutterer:
Good points, I will move the env variables from the Dockerfile to the compose files and use a .env file for the variables 👍 
>`- ./tenants.yml:/${DAMAP_MULTITENANT_CONFIG:-/opt/damap/tenants.yml}:ro`

About the volume definition, writing it like this would force everyone to have a tenants.yml file in the docker folder and i dont know if its the right choice. Also, i have no idea how this will later work with kubernetes, maybe you can explain, then I will push the fixes and we can go towards merging.

Thx 👍

SotosTsepe:
Here's the approach I would take:
For Docker Compose / local development, we can hardcode the path of tenants.yml inside the container and let users decide if they want to enable multitenancy (if yes then one simply edits the file contents of the tenants.yml):

example:
```bash
    environment:
      DAMAP_MULTITENANT_CONFIG: /opt/damap/tenants.yml
      QUARKUS_PROFILE: ${QUARKUS_PROFILE:-}
    volumes:
      - ./docker/tenants.yml:/opt/damap/tenants.yml:ro
```

For Kubernetes/Helm, the same principle applies mount the file as a volume, or configmap instead of a local volume:
```bash
apiVersion: v1
kind: ConfigMap
metadata:
  name: damap-tenants
  namespace: {{ .Release.Namespace }}
data:
  tenants.yml: |-
    {}  # empty by default; can be customized per deployment
```

and mount it in the DAMAP backend container:
```bash
volumeMounts:
  - name: damap-tenants
    mountPath: /opt/damap/tenants.yml
    subPath: tenants.yml
    readOnly: true

volumes:
  - name: damap-tenants
    configMap:
      name: damap-tenants

environment:
  - name: DAMAP_MULTITENANT_CONFIG
    value: /opt/damap/tenants.yml      
```

**_OR_**

We could also keep things fully minimal and let users create their own tenants.yml manually in Docker Compose or OpenShift.
Pros: zero defaults, fully user-controlled.
Cons: container fails if the file is missing; less friendly for local dev or new deployments.

ValentinFutterer:
I went with your first option and tested it in docker, works perfectly 👍 
The only annoying thing is, that the `tenants.yaml` gets mounted into the container irregardless of if we have multitenant active or not.
Can you please do a quick check and then we can merge.

SotosTsepe:
Feedback after our discussion:
- Delete the `DAMAP_MULTITENANT_CONFIG` environment variable for now and keep a fixed path in the container as well as in the host.
    Docs:
    * Default behavior is start without the profile.
    * If one enables the profile, the service won't start by default. A template in the `tenants.yaml`, will show the structure one needs to follow.
    * For the service to start successfully, the operator has to create the database names in the db container and supply the EXACT same names in the `tenants.yaml`
- Rename `.env` to `example.env` in the `docker` directory.

SotosTsepe:
Testing checklist:

- [x] CI Pipeline succeeds
- [x] Test locally on Docker
     - [x] PostgreSQL
     - [ ] Oracle
- [ ] Test on OpenShift

SotosTsepe (approved):
Approving. Will test the remaining items at a later point and apply a patch if necessary, good job :1st_place_medal:

ValentinFutterer:
> Approving. Will test the remaining items at a later point and apply a patch if necessary, good job 🥇

Thanks for your pointers and the reviews 👍

PR #432, opened 2025-10-14
Preview: add license to dataset table

DanceaVlad:
## Description

License for reused datasets was not exported into PDF previously, this PR fixes that problem. Due to a  bug in the export, a workaround was used - this workaround will be obsolote with the new templating engine upgrade, so its fine to leave it in for now.

### DIFF
--- a/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -1227,6 +1227,14 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
         docVar.add(Optional.ofNullable(dataset.getDescription()).orElse(""));
 
         insertTableCells(xwpfTable, newRow, docVar);
+        if (dataset.getLicense() != null) {
+          ELicense license = reusedDatasets.get(i).getLicense();
+          XWPFParagraph paragraph = newRow.getCell(3).getParagraphs().get(0);
+          if (license != ELicense.NOLICENSE && license != ELicense.CUSTOM) {
+            turnRunIntoHyperlinkRun(paragraph.getRuns().get(0), license.getUrl());
+          }
+          commitTableRows(xwpfTable);
+        }
       }
       xwpfTable.removeRow(xwpfTable.getRows().size() - 1);
       xwpfTable.removeRow(1);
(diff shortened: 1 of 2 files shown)


BlueTechie:
Nice work! To keep it consistent with produced datasets, I suggest we also format the default CC-BY-4.0 entry as a hyperlink pointing to the license URL. I have expanded on your commit with my own commit to achieve this, although in the process I noticed a bug that should be fixed in the future. Maybe @ValentinFutterer can provide external feedback?

ValentinFutterer:
> Nice work! To keep it consistent with produced datasets, I suggest we also format the default CC-BY-4.0 entry as a hyperlink pointing to the license URL. I have expanded on your commit with my own commit to achieve this, although in the process I noticed a bug that should be fixed in the future. Maybe @ValentinFutterer can provide external feedback?

This should be fixed incidentally by the template upgrade, so the workaround is fine.

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -1225,7 +1224,22 @@ public void composeTableReusedDatasets(XWPFDocument document, XWPFTable xwpfTabl
...
         insertTableCells(xwpfTable, newRow, docVar);
+        if (dataset.getLicense() != null) {
+          ELicense license = reusedDatasets.get(i).getLicense();
+          XWPFParagraph paragraph = newRow.getCell(3).getParagraphs().get(0);
+          if (license != ELicense.NOLICENSE && license != ELicense.CUSTOM) {
+            System.out.println(
+                "REUSED paragraph hash: "
+                    + paragraph.hashCode()
+                    + " row hash: "
+                    + newRow.hashCode());
```
ValentinFutterer: This seems like leftover from development. In the future, please clean up your code before you submit it :/

src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -1368,6 +1382,8 @@ public void composeTableDatasetPublication(XWPFTable xwpfTable) {
           ELicense license = newDatasets.get(i).getLicense();
           XWPFParagraph paragraph = newRow.getCell(5).getParagraphs().get(0);
           if (license != ELicense.NOLICENSE && license != ELicense.CUSTOM) {
+            System.out.println(
+                "paragraph hash: " + paragraph.hashCode() + " row hash: " + newRow.hashCode());
```
ValentinFutterer: Same as above

ValentinFutterer (approved):
👍

PR #439, opened 2025-10-27
Config: Add default pure url

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?
Add a default pure url to the application yaml instead of empty "". This prevents the project service from failing when the default project service was chosen. Tests should still run as normal.

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -59,7 +59,7 @@ damap:
   elsevier-pure-project-lead-role-classification: /dk/atira/pure/projectlead
   ## HTTP backend:
   elsevier-pure-backend: http
-  elsevier-pure-endpoint-url: ""
+  elsevier-pure-endpoint-url: "https://your-pure-instance.elsevierpure.com/ws/api"
   elsevier-pure-api-key: "your API key here"
   ## Alternatively, the file-based backend:
   # elsevier-pure-backend: file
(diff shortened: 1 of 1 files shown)


lpandath:
@ValentinFutterer  since Pure is disabled by default (`projects-service: default`), I'm wondering if the empty string `""` was actually clearer? It makes it obvious that nothing is configured yet.

Could be that someone might enable Pure integration and forget to replace the dummy URL, leading to confusing connection errors.

Alternative: We could keep it as `""` and just improve the documentation with a comment like:
```yaml
# Replace with your Pure instance URL (e.g., https://your-pure-instance.elsevierpure.com/ws/api)
elsevier-pure-endpoint-url: ""
```

ValentinFutterer:
@lpandath 
> Add a default pure url to the application yaml instead of empty "". This prevents the project service from failing when the default project service was chosen. Tests should still run as normal.

It needs to be prefilled with a syntactically valid url.

> @ValentinFutterer since Pure is disabled by default (projects-service: default), I'm wondering if the empty string "" was actually clearer? It makes it obvious that nothing is configured yet.

origins, backend url, datasource etc are also all prefilled and need to be changed by universities who want their own instance. Imo it would be even more in line with convention if its prefilled with a dummy url. (Or look at the Pure file paths and API key)

PR #449, opened 2025-12-09
Dockerfile enhancements

SotosTsepe:
### What does this PR do?
This PR updates the Docker build process for the DAMAP backend, focusing on using more secure and actively maintained base images, modernizing metadata, and simplifying the build chain.

Changes include:
- Builder image updates:
      `maven:3.9.5-eclipse-temurin-17-alpine` 🠮 `maven:3.9-eclipse-temurin-17-noble`
      The new image is actively maintained, with fewer security vulnerabilities.

- Runner image updates:
       `rockylinux:8.5` 🠮 `eclipse-temurin:17-jre-noble`
       Same as above, the new image base receives ongoing updates and reduces vulnerability surface.

- Replaces deprecated label-schema labels with up-to-date OCI-compliant metadata.

- Packages: since Alpine-based images are no longer used, no extra packages need to be manually installed.

- The final image increased by ~10 MB (~539 MB total), which is an acceptable trade-off for improved security and maintainability.

#### Files Changed
Dockerfile

### Checks
- [X] Tested with Oracle/PostgreSQL on Docker

### Follow-up
Update the current Java version (17 LTS - reaches EOL towards the end of 2026) to 21 LTS in a subsequent PR.


Dockerfile
```diff
@@ -32,13 +36,11 @@ RUN mvn \
     -Dquarkus.package.jar.type=mutable-jar
```
BlueTechie, line 41: One thing I noticed (which was already here before your changes): currently our `Dockerfile` looks like this:
```Dockerfile
COPY src ./src
COPY ./pom.xml .

RUN mvn \
    -Duser.home="${BUILD_HOME}" \
    -B package \
    -DskipTests \
    -Dquarkus.package.jar.type=mutable-jar
```
The issue is that every time the application changes, the Docker cache is invalidated and the layers below are rerun. In our case, each time a file inside `src` changes, the `mvn package` command is run again, which also unnecessarily downloads all dependencies. To optimize this and speed up build times, we can first download all dependencies specified in the `pom.xml`, and then copy the source code and run `mvn package`. Something like:
```Dockerfile
COPY ./pom.xml ./pom.xml
RUN mvn dependency:go-offline -B
COPY ./src ./src
RUN mvn package
```
see https://whitfin.io/blog/speeding-up-maven-docker-builds/

BlueTechie (commented):
Very nice changes, I could not spot any flaws in your code 👍 The only security-related thing to note is that the application must be writable in phase 1 because of quarkus reaugmentation. This means that if something goes very wrong and there is a RCE vulnerability somewhere, the `.jar` and with it the source code files can be rewritten/backdoored. However, since we need reaugmentation to support Postgres and Oracle Databases, it's a trade-off we have to make.

Dockerfile
SotosTsepe: good catch. This PR focused on improving the security of the base images, but given that this is a minor change, I can apply it and resubmit. :+1:

Dockerfile
```diff
@@ -22,23 +26,26 @@ RUN mkdir -p "${BUILD_HOME}/.m2/repository" && chown -R 1000:0 "${BUILD_HOME}"
...
     -DskipTests \
     -Dquarkus.package.jar.type=mutable-jar
 
 # ===== Stage 2: Runtime ===== #
-# This stage will only contain the runtime binaries without build dependencies.
-FROM rockylinux:8.5 AS runner
-
-ARG JAVA_PACKAGE=java-17-openjdk-headless
+# Minimal runtime container using Eclipse Temurin JRE for Quarkus (Ubuntu Noble 24.04 LTS).
+FROM eclipse-temurin:17-jre-noble AS runner
```
BlueTechie, line 45: Good choice 👍 That image only has a few medium severity vulns, which should be fine for a full OS image:
image

(The damap image has a few high severity issues because of keycloak, quarkus, and netty, which should be fixed in a future PR).

Dockerfile
```diff
@@ -47,17 +54,13 @@ WORKDIR /quarkus-app
...
-# Give write permissions to the /quarkus-app/quarkus directory for group 0 (root).
-# The /quarkus-app/quarkus directory contains generated bytecode and metadata that Quarkus updates at runtime.
-# This ensures compatibility with OpenShift and other Kubernetes distributions that use random UIDs for containers.
-RUN dnf install -y openssl tzdata-java curl ca-certificates ${JAVA_PACKAGE} \
-    && dnf clean all -y \
-    && chown 1001:0 -R ./ \
-    && chmod g+rwX ./quarkus
+# Ensure Quarkus can write generated metadata when running as a random UID.
+RUN chown 1001:0 -R ./ && \
+    chmod g+rwX ./quarkus
```
BlueTechie, line 59: I'm curious: is there a specific reason why the `chmod` command does not have a recursive `-R` like the `chown` above? I tested it with and without the flag, and both variations work locally, but in theory shouldn't you also change the permissions for all subdirectories and files? Otherwise I don't see how it's possible that the application rewrites files inside the `/quarkus` folder without crashing.

BlueTechie (commented):
Nice changes 👍 

While digging deeper into our Docker setup, I found something concerning that relates to the Docker ignore file: it was renamed from `.dockerignore` to `.dockerignore.tmp` 4 years ago for an unknown reason. As far as I know, Docker won't recognize this as a Docker ignore file. When this file is missing, Docker sends the whole directory to the Docker daemon/server at the start of every build as a build context. This includes heavy files like the git history and possibly even secrets like environment variables among other things, which is bad for build speed & security. (https://codefresh.io/blog/not-ignore-dockerignore-2/)
Renaming it won't be enough, we will need to include the `/docker` directory as an exception, otherwise the container won't find the `docker-entrypoint.sh` file specified in the Dockerfile.

Since this is outside the scope of this PR, I will create an issue so we can deal with that issue separately. The only remaining question I have is about the possibly missing `-R` flag, otherwise I think we can merge this PR.

Dockerfile
SotosTsepe: This is the best we can do so far, new patches or major versions should decrease the remaining vulnerabilities.

Dockerfile
SotosTsepe: The lack of -R here is intentional.
At runtime, Quarkus does not rewrite existing files under quarkus-app; during re-augmentation it only creates new files under `quarkus/` (generated bytecode and metadata). For that use case, having group write permission on the directory itself is sufficient, since directory write permission controls the ability to create/remove entries.
The existing files and subdirectories already have the correct permissions from build time, and the runtime UID only needs to add new content, not modify existing files. Therefore, a recursive chmod isn’t strictly required. Using g+rwX (without -R) also avoids unnecessarily changing permissions on all files, which keeps the image closer to what was produced during the build stage.
That said, adding -R would also work, but it’s broader than needed for the actual runtime behavior. The difference is hard to spot in a local/docker/vanilla k8s environment; it shows up specifically on OpenShift, where containers run with a random UID but are assigned to GID 0 (root group).

Snippet of the structure:
```bash
└─$ docker exec -u root -it docker-damap-be-1 bash -c "ls -l /quarkus-app"                                                                                                                                                                                                       [16:56:39]
total 36
drwxr-xr-x 1 1001 root  4096 Jan 13 14:43 app
drwxr-xr-x 1 1001 root  4096 Jan 13 14:43 lib
drwxrwxr-x 1 1001 root  4096 Jan 20 10:53 quarkus
-rw-r--r-- 1 1001 root 10876 Jan 13 14:43 quarkus-app-dependencies.txt
-rw-r--r-- 1 1001 root   688 Jan 13 14:43 quarkus-run.jar
```

BlueTechie: This is very valid and clarifies things, thanks for the explanation!

BlueTechie (approved):

ValentinFutterer (approved):
Good work on the upgrade @SotosTsepe and thanks to @BlueTechie for the good review - caching dependecies lowered the builkd time from ~150 to ~40 sec !

PR #450, opened 2025-12-10
Pure: Avoid ClassCastException when project contains external participants

BlueTechie:
## Description

Bugfix

#### What does this PR do?

Testing the PURE integration with a project file containing external participants, such as `damap-backend/src/test/resources/org/damap/base/integration/pure/projects.json` leads to a `ClassCastException`: 

Untitled

The reason for this is the method to get the recommended projects in the first step, which casts every participant to the class `PureAPIInternalParticipantAssociation`, in particular even external participants. This PR fixes the issue by pre-filtering by internal participants only. Since external participants are not able to have a DAMAP account, we do not need to recommend projects where they are a participant.


ValentinFutterer (approved):
👍

PR #452, opened 2025-12-23
feature: translation management

### LINKED ISSUE #445: Add translation management
**Description**
For more context see https://github.com/damap-org/damap-frontend/issues/497, https://github.com/damap-org/damap-frontend/issues/498, https://github.com/damap-org/damap-frontend/issues/499, https://github.com/damap-org/damap-frontend/issues/500

We need the backend to store and manage translations now, since the json files in the frontend code are not very userfriendly to change / impossible in the cloud version. Storing whole  json files is too unwieldy and would make migrations very difficult to manage, so we decided to store the translation in the db like this:
key | default | custom | language | active

**Proposed solution**
- [ ] Create a new table with the above columns
- [ ] Prefill it with the new keys and old default values from https://github.com/damap-org/damap-frontend/issues/497
- [ ] Add GET endpoint to get all active translations 
- [ ] Add POST endpoint to create a new language (Admin only, should basically copy the english rows, replace the language column and keep the custom empty,  default stays english)
- [ ] Add PATCH endpoint to change custom values and active boolean (Admin only, default should never change)
- [ ] Add DELETE endpoint to completely remove one language (except english, Admin only)

**Alternative solution**
Store translations as a json blob

DanceaVlad:
## Description
Feature

#### What does this PR do?
Implements a way for admins to create/delete/modify languages, along with their translation keys and values

#### Code review focus
- Service error handling and permission, such as not allowing the english language to be deleted, not allowing the default translations to be modified...
- Liquibase changelog id and versions

#### Dependencies
This PR requires a frontend UI implementation.

### Checks
- [ ] remove the Python script for generating the Liquibase changelog and its associated README.md and JSON file

closes GH-445


lpandath:
@DanceaVlad
while integrating the translation management frontend with your backend, I encountered a couple of issues:

POST `${this.backendUrl}translations/language/${language}` - when trying to create a language that already exists, it throws an EntityExistsException. It would be better if this endpoint was idempotent (return success if the language already exists) or returned a clearer error message.

DELETE - same, some deletion scenarios need better error handling.
to be able to work during development, I added a tmp mock service TranslationMockService in the frontend that tracks languages when backend operations fail, until this is resolved.

scripts/README.md
```diff
@@ -0,0 +1,82 @@
+# Translation Management Scripts
+
+## Overview
+
+This directory contains scripts for managing backend translations in the DAMAP system.
+
```
GeoffreyKarnbach: Talked with Valentin, maybe rather include this README content in the PR description, as we do not strictly need it here.

src/main/java/org/damap/base/domain/Translation.java
```diff
@@ -0,0 +1,57 @@
...
+  @Size(min = 1, max = 255)
+  private String key;
+
+  @Column(name = "language", nullable = false, length = 2)
+  @NonNull @NotEmpty
+  @Size(min = 2, max = 2)
+  private String language;
+
+  @Column(name = "default_value", nullable = false, length = Length.LONG32)
+  @NonNull @NotEmpty
```
GeoffreyKarnbach: @ NotEmpty already checks for @ NonNull and @ Size (min = 1), you can edit that

src/main/java/org/damap/base/repo/TranslationRepo.java
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * deleteByLanguage.
+   *
+   * @param language a {@link java.lang.String} object
+   * @return number of deleted entities
+   */
+  @Transactional
+  public long deleteByLanguage(String language) {
+    return delete("language", language);
+  }
```
GeoffreyKarnbach, line 74: Could be worth for this very invasive operation to improve a bit the javadoc and warn about its destructive nature.

src/main/java/org/damap/base/rest/translation/service/TranslationService.java
```diff
@@ -0,0 +1,145 @@
...
+    existing.setActive(request.active());
+
+    return existing;
+  }
+
+  /**
+   * deleteLanguage.
+   *
+   * @param language a {@link java.lang.String} language code
+   */
```
GeoffreyKarnbach, line 141: Improve the JavaDoc

src/main/java/org/damap/base/rest/TranslationResource.java
```diff
@@ -0,0 +1,68 @@
...
+import jakarta.ws.rs.*;
+import java.util.List;
+import lombok.extern.jbosslog.JBossLog;
+import org.damap.base.domain.Translation;
+import org.damap.base.rest.translation.service.TranslationService;
+import org.eclipse.microprofile.openapi.annotations.parameters.RequestBody;
+
+/** TranslationResource class. */
+@Path("/api/languages")
+@JBossLog
```
GeoffreyKarnbach, line 18: Add as class level annotation

> @Produces(MediaType.APPLICATION_JSON)

src/main/java/org/damap/base/rest/TranslationResource.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  @GET
+  public List<String> getAllLanguages() {
+    return translationService.getAllLanguages();
+  }
+
+  @GET
+  @Path("/{language}")
+  public List<Translation> getTranslationsForLanguage(
```
GeoffreyKarnbach, line 36: Do we want to return the entity or should we also maybe create a DTO?

src/main/resources/org/damap/base/db/changeLog-5.x/changeLog-5.0.0_1.yaml
```diff
@@ -0,0 +1,107 @@
+databaseChangeLog:
+  - changeSet:
+      id: 25
```
GeoffreyKarnbach: ID 27, 25 and 26 already exist

src/main/resources/org/damap/base/db/changeLog-5.x/changeLog-5.0.0_2.yaml
GeoffreyKarnbach, line 1: ID 28

GeoffreyKarnbach (changes requested):
Great work! Please just change the few minor requests!

GeoffreyKarnbach:
Also rebase the branch please, there were no conflicts for me when I did it locally!

src/main/java/org/damap/base/rest/translation/service/TranslationService.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+    return translationRepo.findByLanguage(newLanguage);
+  }
+
+  /**
+   * getAllLanguages.
+   *
+   * @return a {@link java.util.List} of language codes
+   */
+  public List<String> getAllLanguages() {
```
lpandath, line 62: This is very inefficient because it loads every single translation row from the database into memory just to get the language codes...would instead be better to handle this at the database level instead?

ValentinFutterer:
Finished by other PRs

PR #453, opened 2025-12-30
EduId integration

### LINKED ISSUE #411: eduID integration
In order to make DAMAP Cloud possible, we must integrate [eduID](https://eduid.at), the Austrian federated login. For the first version, everyone should only be able to log into their own educational institution's instance. If no instance exists, the login should be rejected.

ValentinFutterer:
## Description
Feature

#### What does this PR do?

Adds support for an integration with EduId using an intermediary SAML translation setup with a keycloak interface (not included in the project). 
To support integrating with any keycloak, claims were made configurable and config options in `application.yaml` were renamed to be clearer and given corresponding environment variables to ease deployment.
For multitenant deployments, the affiliation is extracted from the claims, in the case of double affiliation, we return a 401.
Roles claims are now also configuable using a path, e.g. "realm_access/roles" in case of the test keycloak

Its a draft, so no documentation yet. I will add it after everything looks good and is reviewed.

#### Breaking changes
Incompatible with older frontends and older setups will need to change their configs.

#### Code review focus
Please check naming of variables - I hope everything is clearly named but a second opinion is wanted :)
I had to do a weird workaround with the tests regarding the admin-role-name property - I couldnt find a better way, maybe you have an idea.

#### Dependencies
https://github.com/damap-org/damap-frontend/pull/511

closes GH-411

### DIFF
--- a/src/main/java/org/damap/base/security/SecurityService.java
+++ b/src/main/java/org/damap/base/security/SecurityService.java
@@ -9,8 +9,12 @@ import io.smallrye.jwt.auth.principal.JWTParser;
 import io.smallrye.jwt.auth.principal.ParseException;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
+import jakarta.json.JsonArray;
+import jakarta.json.JsonString;
 import jakarta.ws.rs.core.HttpHeaders;
 import java.security.Principal;
+import java.util.List;
+import java.util.Optional;
 import lombok.extern.jbosslog.JBossLog;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.jwt.JsonWebToken;
@@ -24,8 +28,17 @@ import org.eclipse.microprofile.jwt.JsonWebToken;
 public class SecurityService {
   @Inject SecurityIdentity securityIdentity;
 
-  @ConfigProperty(name = "damap.auth.user")
-  String authUser;
+  @ConfigProperty(name = "damap.auth.user-id-claim")
+  String userIdClaim;
+
+  @ConfigProperty(name = "damap.auth.admin-role-name")
+  String adminRoleName;
+
+  @ConfigProperty(name = "damap.auth.affiliations-claim")
+  String affiliationsClaim;
+
+  @ConfigProperty(name = "tenants", defaultValue = "")
+  Optional<List<String>> tenants;
 
   @ConfigProperty(name = "invenio.shared-secret")
   String sharedSecret;
@@ -41,7 +54,7 @@ public class SecurityService {
     final Principal principal = securityIdentity.getPrincipal();
     if (!(principal instanceof OidcJwtCallerPrincipal)) return null;
 
-    return ((OidcJwtCallerPrincipal) principal).getClaims().getClaimValue(authUser).toString();
+    return ((OidcJwtCallerPrincipal) principal).getClaims().getClaimValue(userIdClaim).toString();
   }
 
   /**
@@ -52,16 +65,14 @@ public class SecurityService {
   public String getUserName() {
     final Principal principal = securityIdentity.getPrincipal();
     if (!(principal instanceof OidcJwtCallerPrincipal)) return null;
-    return ((OidcJwtCallerPrincipal) principal).getName();
+    return principal.getName();
   }
 
   public String getDisplayName() {
     final Principal principal = securityIdentity.getPrincipal();
-    if (!(principal instanceof OidcJwtCallerPrincipal)) {
+    if (!(principal instanceof OidcJwtCallerPrincipal oidcPrincipal)) {
       return null;
     }
-
-    OidcJwtCallerPrincipal oidcPrincipal = (OidcJwtCallerPrincipal) principal;
 
     String name = getClaimValueAsString(oidcPrincipal, "name");
     if (name != null) {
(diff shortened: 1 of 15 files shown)


src/main/java/org/damap/base/security/SecurityService.java
```diff
@@ -88,7 +99,35 @@ private String getClaimValueAsString(OidcJwtCallerPrincipal oidcPrincipal, Strin
...
+    }
+
+    List<String> affiliations =
+        ((JsonArray) oidcPrincipal.getClaims().getClaimValue(affiliationsClaim))
+            .stream()
+                .map(JsonString.class::cast)
+                .map(JsonString::getString)
+                .map(aff -> aff.split("@")[1])
+                .distinct()
+                .toList();
```
bitmotivator, line 131: This could do with a little comment. People who do not know what eduID is will be very confused by this.

src/main/java/org/damap/base/security/SecurityService.java
```diff
@@ -88,7 +99,35 @@ private String getClaimValueAsString(OidcJwtCallerPrincipal oidcPrincipal, Strin
...
+
+    List<String> tenantList = tenants.orElse(List.of());
+
+    // check if affiliations are actual tenants
+    // TODO: Move this code into the TenantResolver after merging
+    List<String> validAffiliations = affiliations.stream().filter(tenantList::contains).toList();
+    System.out.println(validAffiliations);
```
bitmotivator: This seems to be debug code?

src/main/java/org/damap/base/security/SecurityService.java
```diff
@@ -88,7 +99,35 @@ private String getClaimValueAsString(OidcJwtCallerPrincipal oidcPrincipal, Strin
...
+    if (!(principal instanceof OidcJwtCallerPrincipal oidcPrincipal)) {
+      return null;
+    }
+
+    List<String> affiliations =
+        ((JsonArray) oidcPrincipal.getClaims().getClaimValue(affiliationsClaim))
+            .stream()
+                .map(JsonString.class::cast)
+                .map(JsonString::getString)
+                .map(aff -> aff.split("@")[1])
```
bitmotivator: This is assuming that the affiliation data contains an `@`. This may not work for eduGAIN, but it works for eduID. The exception this throws will be extremely unhelpful.

src/main/resources/application.yaml
```diff
@@ -5,21 +5,33 @@ damap:
...
-      # Client ID as configured in the OIDC Server:
-      client: 'damap-be'
-    frontend:
-      # Configures the OIDC URL the frontend should send users to in order to log in. This is typically the same
-      # as the backend URL, but may differ if the backend can access the OIDC server over a non-public network.
-      url: http://localhost:8087/realms/damap
-      # Client ID as configured in the OIDC Server:
-      client: 'damap'
+    # Configures the OIDC realm the DAMAP backend should query to verify authentication:
+    server-url: ${DAMAP_SERVER_URL:http://localhost:8087/realms/damap}
```
bitmotivator: This change breaks backwards compatibility. This should be thoroughly documented. Also, please check with @ZenoLC if there's a reason to use a separate OIDC config for the frontend or if this only exists for historical reasons.

src/main/resources/application.yaml
```diff
@@ -5,21 +5,33 @@ damap:
...
+    # Configures the OIDC URL the frontend should send users to in order to log in. This is typically the same
+    # as the backend URL, but may differ if the backend can access the OIDC server over a non-public network.
+    issuer: ${DAMAP_ISSUER:http://localhost:8087/realms/damap}
```
bitmotivator: Is this supposed to read environment variables? If so, [Quarkus does that automatically](https://quarkus.io/guides/config-reference).

src/main/resources/application.yaml
```diff
@@ -5,21 +5,33 @@ damap:
...
+    # Client ID as configured in the OIDC Server:
+    clientID: ${DAMAP_CLIENTID:damap}
     # Scopes to request during login:
-    scope: 'openid profile email offline_access microprofile-jwt roles personID' # your-backend-authentication-scope, including your variable for your user IDs
+    scope: ${DAMAP_SCOPE:openid profile affiliation}
```
bitmotivator: This changes the default scope. Are you sure you want to do that?

bitmotivator (commented):
Looks plausible. My biggest concern here would be that these changes seem to be backwards incompatible. Together with possible DB changes I would consider releasing a next major version. I would also implore you to create a third PR for the reference manual (website) and document these changes, otherwise the docs will get outdated pretty fast.

src/main/resources/application.yaml
ValentinFutterer: I know, but we are planning to release a major version anyway, so it seems like a good point to introduce these changes.

The reason is written in this comment:
```
# Configures the OIDC URL the frontend should send users to in order to log in. This is typically the same
    # as the backend URL, but may differ if the backend can access the OIDC server over a non-public network.
```

I domt know if its normal to have this option (judging by your comment its not). I will check with Zeno if its still needed and why they included it.

src/main/java/org/damap/base/security/SecurityService.java
ValentinFutterer: True, for now I think I will just throaw a meaningful exception and leave it to be expanded later

src/main/java/org/damap/base/security/SecurityService.java
ValentinFutterer: Fair point - will do

src/main/resources/application.yaml
ValentinFutterer: Oh so DAMAP_AUTH_ISSUER will automatically be mapped to the issuer field in the config? Thats good to know :)  I will also adapt the rest of the application.yaml 👍

src/main/resources/application.yaml
ValentinFutterer: The default scope includes personID, thats why I removed it and set it to somehting more neutral. The old default scope is now in the dev profile. Whats the reason for keeping the old scopes as default? Backwards compatibility with old deployments?

src/main/java/org/damap/base/security/SecurityService.java
ValentinFutterer: The print yes. The other code should be moved later.

PR #454, opened 2026-01-19
Gk/user management eduid

GeoffreyKarnbach:
## Description
This PR decouples access management from project contributors by introducing a local user persistence layer. It ensures user data used for access control is stored consistently in the database rather than relying on external services or transient metadata.

#### What does this PR do?
* **User Persistence:** Adds `damap_user` table and repository to store user identity and metadata.
* **Auth Sync:** Implements `DamapSecurityAugmentor` to synchronize user data from OIDC tokens to the database upon authentication. Includes caching to minimize database writes.
* **Access Refactoring:**
    * Decouples `AccessDO` from `ContributorDO` (removes inheritance).
    * Updates `AccessService` to return only explicit access entries; contributors are no longer merged into the response.
    * Renames `universityId` to `identifier` and `access` to `role` in `AccessDO`.
* **Search:** Adds `GET /api/access/user-search` to search registered users by name or email.

#### Breaking changes
* `AccessDO` structure updated (fields renamed, inheritance removed).
* `GET /api/access/dmps/{id}` now returns `List<AccessDO>` instead of `List<ContributorDO>`.
* Contributors listed in a DMP do not implicitly receive access rights anymore; access must be explicitly granted.

#### Code review focus
* `UserSyncService` and `DamapSecurityAugmentor` for synchronization logic.
* `AccessService.getByDmpId` for the removal of the contributor merge logic.
* `AccessMapper` for the updated field mapping.

#### Dependencies
Frontend PR: https://github.com/damap-org/damap-frontend/pull/514

### Checks
- [ ] Export updated
- [x] Documentation added
- [x] Tests added
- [ ] E2e tests created

closes GH-

### DIFF
--- a/src/main/java/org/damap/base/domain/User.java
+++ b/src/main/java/org/damap/base/domain/User.java
@@ -1,0 +1,30 @@
+package org.damap.base.domain;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import jakarta.persistence.*;
+import lombok.*;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@Entity(name = "damap_user")
+@Table(name = "damap_user")
+@EqualsAndHashCode(callSuper = true)
+public class User extends PanacheEntity {
+
+  @Column(name = "identifier", unique = true, nullable = false)
+  private String identifier;
+
+  @Column(name = "name")
+  private String name;
+
+  @Column(name = "first_name")
+  private String firstName;
+
+  @Column(name = "last_name")
+  private String lastName;
+
+  @Column(name = "email")
+  private String email;
+}
--- a/src/main/java/org/damap/base/repo/DamapUserRepo.java
+++ b/src/main/java/org/damap/base/repo/DamapUserRepo.java
@@ -1,0 +1,30 @@
+package org.damap.base.repo;
+
+import io.quarkus.hibernate.orm.panache.PanacheRepository;
+import io.quarkus.panache.common.Page;
+import io.quarkus.panache.common.Parameters;
+import jakarta.enterprise.context.ApplicationScoped;
+import java.util.List;
+import org.damap.base.domain.User;
+
+@ApplicationScoped
+public class DamapUserRepo implements PanacheRepository<User> {
+
+  /** Find user list by match in name or email. */
+  public List<User> findUsersByNameOrEmailMatch(String match) {
+    if (match == null || match.isBlank()) {
+      return List.of();
+    }
+    String pattern = "%" + match.trim().toLowerCase() + "%";
+
+    return find(
+            "lower(name) like :pattern or lower(email) like :pattern",
+            Parameters.with("pattern", pattern))
+        .page(Page.ofSize(20))
+        .list();
+  }
+
+  public User findUserByIdentifier(String identifier) {
+    return find("identifier", identifier).firstResult();
+  }
+}
--- a/src/main/java/org/damap/base/rest/access/domain/AccessDO.java
+++ b/src/main/java/org/damap/base/rest/access/domain/AccessDO.java
@@ -6,23 +6,30 @@ import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Positive;
 import java.util.Date;
 import lombok.Data;
-import lombok.EqualsAndHashCode;
 import org.damap.base.enums.EFunctionRole;
-import org.damap.base.rest.dmp.domain.ContributorDO;
 
 /** AccessDO class. */
 @Data
-@EqualsAndHashCode(callSuper = true)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class AccessDO extends ContributorDO {
+public class AccessDO {
+
+  private Long id;
 
   @NotNull @Positive private long dmpId;
 
-  @NotNull private EFunctionRole access;
+  @NotNull private EFunctionRole role;
 
   @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
   private Date start;
 
   @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
   private Date until;
+
+  private String identifier;
+
+  private String firstName;
+
+  private String lastName;
+
+  private String mbox;
 }
--- a/src/main/java/org/damap/base/rest/access/mapper/AccessMapper.java
+++ b/src/main/java/org/damap/base/rest/access/mapper/AccessMapper.java
@@ -1,14 +1,10 @@
 package org.damap.base.rest.access.mapper;
 
-import java.util.Optional;
 import lombok.experimental.UtilityClass;
 import org.damap.base.domain.Access;
-import org.damap.base.domain.Contributor;
 import org.damap.base.domain.Dmp;
-import org.damap.base.domain.Identifier;
+import org.damap.base.domain.User;
 import org.damap.base.rest.access.domain.AccessDO;
-import org.damap.base.rest.dmp.domain.IdentifierDO;
-import org.damap.base.rest.dmp.mapper.IdentifierDOMapper;
 import org.damap.base.rest.dmp.mapper.MapperService;
 
 /** AccessMapper class. */
@@ -22,30 +18,18 @@ public class AccessMapper {
    * @param accessDO a {@link org.damap.base.rest.access.domain.AccessDO} object
    * @return a {@link org.damap.base.rest.access.domain.AccessDO} object
    */
-  public AccessDO mapEntityToDO(Access access, AccessDO accessDO) {
+  public AccessDO mapEntityToDO(Access access, User user, AccessDO accessDO) {
     accessDO.setId(access.id);
     accessDO.setDmpId(access.getDmp().id);
-    accessDO.setUniversityId(access.getUniversityId());
-    // Get name and mail for accesses from contributor list
-    Optional<Contributor> contributor =
-        access.getDmp().getContributorList().stream()
-            .filter(
-                c ->
-                    c.getUniversityId() != null
-                        && c.getUniversityId().equals(access.getUniversityId()))
-            .findFirst();
-    if (contributor.isPresent()) {
-      Contributor c = contributor.get();
-      accessDO.setFirstName(c.getFirstName());
-      accessDO.setLastName(c.getLastName());
-      accessDO.setMbox(c.getMbox());
+    accessDO.setIdentifier(access.getUniversityId());
+
+    if (user != null) {
+      accessDO.setMbox(user.getEmail());
+      accessDO.setFirstName(user.getFirstName());
+      accessDO.setLastName(user.getLastName());
     }
-    if (access.getPersonIdentifier() != null) {
-      IdentifierDO identifierDO = new IdentifierDO();
-      IdentifierDOMapper.mapEntityToDO(access.getPersonIdentifier(), identifierDO);
-      accessDO.setPersonId(identifierDO);
-    }
-    accessDO.setAccess(access.getRole());
+
+    accessDO.setRole(access.getRole());
     accessDO.setStart(access.getStart());
     accessDO.setUntil(access.getUntil());
     return accessDO;
@@ -62,12 +46,9 @@ public class AccessMapper {
   public Access mapDOtoEntity(AccessDO accessDO, Access access, MapperService mapperService) {
     Dmp dmp = mapperService.getDmpById(accessDO.getDmpId());
     access.setDmp(dmp);
-    access.setRole(accessDO.getAccess());
-    access.setUniversityId(accessDO.getUniversityId());
-    if (accessDO.getPersonId() != null) {
-      access.setPersonIdentifier(
-          IdentifierDOMapper.mapDOtoEntity(accessDO.getPersonId(), new Identifier()));
-    }
+    access.setRole(accessDO.getRole());
+    access.setUniversityId(accessDO.getIdentifier());
+
     access.setStart(accessDO.getStart());
     access.setUntil(accessDO.getUntil());
     return access;
--- a/src/main/java/org/damap/base/rest/access/service/AccessService.java
+++ b/src/main/java/org/damap/base/rest/access/service/AccessService.java
@@ -1,7 +1,5 @@
 package org.damap.base.rest.access.service;
 
-import static org.damap.base.utils.EqualityUtils.nullExclusiveEquals;
-
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.transaction.Transactional;
@@ -11,14 +9,13 @@ import java.util.List;
 import lombok.extern.jbosslog.JBossLog;
 import org.damap.base.domain.Access;
 import org.damap.base.domain.Dmp;
-import org.damap.base.integration.PersonService;
+import org.damap.base.domain.User;
 import org.damap.base.repo.AccessRepo;
+import org.damap.base.repo.DamapUserRepo;
 import org.damap.base.repo.DmpRepo;
-import org.damap.base.rest.PersonServiceBroker;
 import org.damap.base.rest.access.domain.AccessDO;
+import org.damap.base.rest.access.domain.UserDO;
 import org.damap.base.rest.access.mapper.AccessMapper;
-import org.damap.base.rest.dmp.domain.ContributorDO;
-import org.damap.base.rest.dmp.mapper.ContributorDOMapper;
 import org.damap.base.rest.dmp.mapper.MapperService;
 
 /** AccessService class. */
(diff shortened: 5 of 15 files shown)


src/main/java/org/damap/base/domain/User.java
```diff
@@ -0,0 +1,23 @@
+package org.damap.base.domain;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import jakarta.persistence.*;
+import lombok.*;
+
+@Data
```
ValentinFutterer, line 7: Missing an `@EqualsAndHashCode(callSuper = true)` annotation

src/main/java/org/damap/base/domain/User.java
```diff
@@ -0,0 +1,23 @@
...
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@Entity(name = "damap_user")
+@Table(name = "damap_user")
+public class User extends PanacheEntity {
+
+  @Column(name = "subject_id")
+  private String subjectId;
```
ValentinFutterer: SubjectId is bad naming for this field - this field could have many different names, since its freely configurable. Its purpose is to be a unique identifier for the user, naming should reflect this.

Should also be `@Unique` and `@NonNull`

src/main/java/org/damap/base/domain/User.java
```diff
@@ -0,0 +1,23 @@
...
+
+  @Column(name = "nickname")
+  private String nickname;
```
ValentinFutterer: In a similar vein, this is not a nickname, like vfuttere, but the full name of the person. 

Also, in `AccessService` you are splitting the name into first and last name. Why not store `name`, `firstName` and `lastName`, instead of doing the error prone conversion?

src/main/java/org/damap/base/repo/DamapUserRepo.java
```diff
@@ -0,0 +1,30 @@
...
+import io.quarkus.hibernate.orm.panache.PanacheRepository;
+import io.quarkus.panache.common.Page;
+import io.quarkus.panache.common.Parameters;
+import jakarta.enterprise.context.ApplicationScoped;
+import java.util.List;
+import org.damap.base.domain.User;
+
+@ApplicationScoped
+public class DamapUserRepo implements PanacheRepository<User> {
+
```
ValentinFutterer, line 12: OK why is there a `UserRepo`

src/main/java/org/damap/base/repo/UserRepo.java
```diff
@@ -0,0 +1,23 @@
+package org.damap.base.repo;
+
+import io.quarkus.hibernate.orm.panache.PanacheRepository;
+import io.quarkus.panache.common.Parameters;
+import jakarta.enterprise.context.ApplicationScoped;
+import org.damap.base.domain.User;
+
+@ApplicationScoped
+public class UserRepo implements PanacheRepository<User> {
+
```
ValentinFutterer: `DamapUserRepo` exists

src/main/java/org/damap/base/rest/access/domain/AccessUserDO.java
```diff
@@ -0,0 +1,15 @@
...
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class AccessUserDO {
+
+  private String subjectId;
+
+  private String nickname;
+
+  private String email;
+}
```
ValentinFutterer: Isnt that just the UserDO?

src/main/java/org/damap/base/rest/access/mapper/AccessMapper.java
```diff
@@ -43,7 +42,6 @@ public AccessDO mapEntityToDO(Access access, AccessDO accessDO) {
     if (access.getPersonIdentifier() != null) {
       IdentifierDO identifierDO = new IdentifierDO();
       IdentifierDOMapper.mapEntityToDO(access.getPersonIdentifier(), identifierDO);
```
ValentinFutterer: Why is this part still here, to my knowledge, this is not needed anymore. You should take a deeper look at the Access entitty and the existing code and do the refactoring properly.

src/main/java/org/damap/base/rest/access/service/AccessService.java
```diff
@@ -37,55 +35,52 @@ public class AccessService {
 
   @Inject PersonServiceBroker personServiceBroker;
 
+  @Inject DamapUserRepo damapUserRepo;
+
   /**
-   * getByDmpId.
+   * getByDmpId. Returns ONLY users explicitly granted access via the Access table. Contributors
+   * listed in the DMP are ignored.
```
ValentinFutterer: This is written in a way where you need knowledge about history to understand it. Always write comments futureproof.

src/main/java/org/damap/base/rest/access/service/AccessService.java
```diff
@@ -37,55 +35,52 @@ public class AccessService {
 
   @Inject PersonServiceBroker personServiceBroker;
```
ValentinFutterer: `ENABLED_PERSON_SERVICE` and `personServiceBroker` fields in this file are not needed anymore and should be removed.

src/main/java/org/damap/base/rest/access/domain/AccessDO.java
```diff
@@ -25,4 +24,12 @@ public class AccessDO extends ContributorDO {
 
   @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
   private Date until;
+
+  private String universityId;
```
ValentinFutterer: When you rename the subject field, you should also rename this one accordingly (in the whole project)

src/main/java/org/damap/base/rest/access/service/AccessService.java
```diff
@@ -37,55 +35,52 @@ public class AccessService {
...
+                      accessDO.setLastName(nickname.substring(splitIndex + 1));
+                    } else {
+                      accessDO.setFirstName(nickname);
+                      accessDO.setLastName("");
+                    }
+                  }
+                } else {
+                  log.warn(
+                      "User not found in internal DB for Subject ID: " + access.getUniversityId());
+                }
```
ValentinFutterer: Is there a reality in where we create an access without an mbox?
If not, then this whole piece of code isnt needed anymore.
Also, fetching the user and populating the DO with name, mail, etc, should probbaly be done by the mapper.
Sidenote: Mention of Subject ID in an error message again.

src/main/java/org/damap/base/rest/auth/UserSyncService.java
```diff
@@ -0,0 +1,34 @@
...
+      // Create new
+      user = User.builder().subjectId(sub).email(email).nickname(nickname).build();
+      userRepo.persist(user);
+    } else {
+      user.setEmail(email);
+      user.setNickname(nickname);
+      userRepo.persist(user);
+    }
+
+    System.out.println("User sync executed for: " + sub);
```
ValentinFutterer: Leftover from development

src/main/java/org/damap/base/rest/access/domain/AccessDO.java
ValentinFutterer: Why is this called access? In the `Access` entity its named role. Role is a better fit here, imo.

src/main/java/org/damap/base/rest/access/mapper/AccessMapper.java
ValentinFutterer: There is still a contributor in use here

src/main/java/org/damap/base/rest/access/domain/AccessDO.java
```diff
@@ -25,4 +24,12 @@ public class AccessDO extends ContributorDO {
...
+
+  private String firstName;
+
+  private String lastName;
+
+  private String mbox;
```
ValentinFutterer, line 34: Now that i think about it, this is essentially just the user object. Maybe there is world where its useful to replace the fields with one user field, but i havent thought that through. You decide if its worth.

src/main/resources/application.yaml
```diff
@@ -173,7 +177,7 @@ rest:
       issuer: http://localhost:8087/realms/damap
       clientID: damap
       scope: 'openid profile email offline_access microprofile-jwt roles personID'
-      user-id-claim: personID
+      user-id-claim: sub
```
ValentinFutterer: Leftover from development

src/main/java/org/damap/base/rest/access/service/AccessService.java
```diff
@@ -122,4 +117,19 @@ public AccessDO create(AccessDO accessDO) {
   public void delete(long id) {
     accessRepo.deleteById(id);
   }
+
+  public List<AccessUserDO> getAllMatchingUsers(String match) {
```
ValentinFutterer: Isnt that just a search function? I would rename to avoid confusion, at first i thought it was a random helper function.

src/main/java/org/damap/base/security/DamapSecurityAugmentor.java
```diff
@@ -0,0 +1,39 @@
...
+  @Inject UserSyncService userSyncService;
+
+  @Override
+  public Uni<SecurityIdentity> augment(
+      SecurityIdentity identity, AuthenticationRequestContext context) {
+    if (identity.isAnonymous()) {
+      return Uni.createFrom().item(identity);
+    }
+
+    if (identity.getPrincipal() instanceof JsonWebToken jwt) {
```
ValentinFutterer: Should be `OidcJwtCallerPrincipal`, see SecurityService

src/main/java/org/damap/base/security/DamapSecurityAugmentor.java
```diff
@@ -0,0 +1,39 @@
...
+
+      String sub = jwt.getSubject();
+      String email = jwt.getClaim("email");
+      String name = jwt.getClaim("name");
```
ValentinFutterer: Claim names are configurable now, please check the application yaml for more information on that.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_1.yaml
```diff
@@ -0,0 +1,33 @@
+databaseChangeLog:
```
ValentinFutterer, line 1: We are in version 4.7.0

src/main/resources/application.yaml
```diff
@@ -135,6 +135,10 @@ quarkus:
         expire-after-write: P1D
       "repository":
         expire-after-write: P1D
+      "user-sync-cache":
+        initial-capacity: 100
+        maximum-size: 1000
+        expire-after-write: 10m
```
ValentinFutterer: I would maybe up the expiration time to 1h or 2h. People easily spend this much time while writing DMPs and the name of people rarely ever changes. Maybe even go up to 6 or 12 hours. I think the memory overhead is manageable with the amount of users we have.

ValentinFutterer (commented):
Main functionality is pretty solid, but the refactoring of the old code was a little sloppy at places :/
If you clean up the code it looks good to go. (Integration testing with the frontend still pending)

ValentinFutterer:
Ah and i forgot to emntion - you need to write documentation. Every PR needs a description, especially more omportant ones like this with context. You should also write comments on the DAMAPSecurityAugmentor and UserSyncService, toi shortly explain why they are here and what they do.

src/main/java/org/damap/base/rest/access/domain/AccessDO.java
GeoffreyKarnbach: I think it is cleaner to not return the entities and have a separate DO for it.

src/main/java/org/damap/base/rest/access/domain/AccessUserDO.java
GeoffreyKarnbach: It is the user entity, but the entity also contains the internal ID, probably better to not expose it and keep a clean separation between DO and Entities.

src/main/java/org/damap/base/rest/access/service/AccessService.java
GeoffreyKarnbach: Fetching the user cannot be done from the mapper, as it is static and i cannot inject the DamapUserRepo, but i pass it now to the mapper as well. Fixed the rest as well.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_1.yaml
```diff
@@ -0,0 +1,39 @@
...
+            tableName: damap_user
+            columns:
+              - column:
+                  name: id
+                  type: BIGINT
+                  constraints:
+                    primaryKey: true
+              - column:
+                  name: identifier
+                  type: VARCHAR(255)
```
ValentinFutterer, line 47: Liquibase has to reflect the java code exactly, so you always need to keep it in sync. Missing nullable false and unique constraints.

src/main/java/org/damap/base/repo/DamapUserRepo.java
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * @param identifier the subject identifier
+   * @return
+   */
+  public User findUserBySubject(String identifier) {
+    return find(
+            "select user_ from damap_user user_ where user_.identifier = :identifier",
+            Parameters.with("identifier", identifier))
+        .firstResult();
+  }
```
ValentinFutterer: Subject is still mentioned and those two methods are very similar / essentially duplicated code.

src/main/java/org/damap/base/rest/access/service/AccessService.java
ValentinFutterer: Yeah, I am not happy with every mapper being static (for the above reason), but we cant breakt hat convention right now. So this is fine for now.

src/main/java/org/damap/base/rest/access/service/AccessService.java
```diff
@@ -122,4 +89,19 @@ public AccessDO create(AccessDO accessDO) {
   public void delete(long id) {
     accessRepo.deleteById(id);
   }
+
+  public List<AccessUserDO> search(String match) {
```
ValentinFutterer: You would assume a function named search in `AccessService` searches Access.
Rename to seachAccessUserDO or something similar. Might also be an idea to put the function in its own service? Rigth now its fine to have it togetehr with the AccessService, but in the future, other services might also want to fetch users.

src/main/java/org/damap/base/security/DamapSecurityAugmentor.java
```diff
@@ -0,0 +1,57 @@
...
+  @Override
+  public Uni<SecurityIdentity> augment(
+      SecurityIdentity identity, AuthenticationRequestContext context) {
+    if (identity.isAnonymous()) {
+      return Uni.createFrom().item(identity);
+    }
+
+    if (identity.getPrincipal() instanceof OidcJwtCallerPrincipal jwt) {
+
+      String sub = jwt.getClaim(userIdClaim);
```
ValentinFutterer: `sub` is a bad name here, why not UserId.

src/main/resources/application.yaml
ValentinFutterer: The comment was ignored

src/main/resources/application.yaml
ValentinFutterer: .

src/main/java/org/damap/base/rest/access/domain/AccessUserDO.java
ValentinFutterer: Ok I know again what I meant to say with the comment 😅 
`AccessUserDO` is bad naming, since it conceptually links the DO always to Access management and doesnt allow for easy extension. The DO also doesnt do anything special that would inherently link it to Access management.

src/main/java/org/damap/base/rest/access/domain/AccessUserDO.java
```diff
@@ -0,0 +1,15 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Data;
+
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class AccessUserDO {
+
+  private String identifier;
+
+  private String nickname;
```
ValentinFutterer: The DO still has the nickname naming.

ValentinFutterer (commented):
https://github.com/damap-org/damap-backend/pull/454#issuecomment-3835214546

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_1.yaml
BlueTechie: The changelog still has the nullable field set to `true`, but it should be `false` since the Java code for this column is:
```java
@Column(name = "identifier", unique = true, nullable = false)
```

src/main/resources/application.yaml
BlueTechie: Just making sure: this change has been rolled back, `user-id-claim` is back to being `personID`. Not sure if this is what you meant.

ValentinFutterer (approved):
Ready to merge 👍

PR #457, opened 2026-01-20
Repair Docker ignore file for build time and security

### LINKED ISSUE #456: Docker ignore file gets ignored (security relevant)
**Context:** 
When building a Docker image, the whole project directory is sent to the Docker daemon (which acts as a server) to act as a build context. This is unless we specify a different behavior in a `.dockerignore` file. This default behaviour can be bad both for build times and security, since we send over heavy files like the entire git history and in some cases sensitive files like environment variables to another process unnecessarily. 
For more context: https://codefresh.io/blog/not-ignore-dockerignore-2/

**Bug description:**
A little over 4 years ago, `.dockerignore` got renamed to `.dockerignore.tmp`. Docker will therefore ignore this file, which puts us in the situation above.

**How to fix:**
We need to do the following:
1) rename  `.dockerignore.tmp` to `.dockerignore`
2) add the whitelist line `!docker/docker-entrypoint.sh` inside the Docker ignore file, since this file is needed as specified in the Dockerfile.

**Is this a regression?**
Yes

BlueTechie:
## Description

Bugfix

**Context:**
When building a Docker image, the entire project directory is sent to the Docker daemon (which acts as a server) as the build context. This is unless we specify a different behavior in a `.dockerignore` file. The default behaviour without this file can be bad both for build times and security, since we send over heavy files like the entire git history and in some cases sensitive files like environment variables to another process unnecessarily. 
For more context: https://codefresh.io/blog/not-ignore-dockerignore-2/

A little over 4 years ago, `.dockerignore` got renamed to `.dockerignore.tmp`. Docker will therefore ignore this file, which puts us in the situation above.

#### What does this PR do?

- renamed `.dockerfile.tmp` to `.dockerfile`
- added the whitelist rule `!docker/docker-entrypoint.sh`. This is because the entrypoint file is needed as specified in the Dockerfile.

#### Breaking changes

closes GH-456

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,4 @@
 !target/*-runner
 !target/*-runner.jar
 !target/lib/*
+!docker/docker-entrypoint.sh


ValentinFutterer (approved):
Well spotted, good work 👍

PR #458, opened 2026-01-23
Pure: Fix endless loop in listAll

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?
Fixes a Bug where listallPersons/Projects enters an endless loop, due to the offset not being upated correctly. I returned the fucntions to their original implementations, seems like they got changed during the long development process.


lpandath (approved):
Approved, fixes the endless loop in Pure pagination. 

_Note_: this was already resolved, but it seems to have gotten lost in a larger PR. Good to have it properly fixed now.

PR #461, opened 2026-02-05
pure: allow instance use of Pure API and add improvements

lpandath:
## Description
I made the Pure and related classes that the instance uses public. With that, the TUG instance can use
Pure from core and extend it e.g. my own project service that builds on the core one.

#### Code review focus
`ProjectDO` has two different fields: `id` - database primary key and `universityId` (e.g. Pure UUID). When mapping from Pure we must only set `universityId` from Pure’s UUID. We must not set `id` from Pure’s ID: that would make Hibernate treat the entity as an existing one and trigger `EntityExistsException` when saving a DMP that links to this project. The rest of the codebase already uses `universityId` for lookups (e.g. `DmpService`, project read by `universityId`), so this is the correct field for the Pure project reference.

Also handled the pagination - instead of loading all projects in one go; add null checks and error handling.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-

### DIFF
--- a/src/main/java/org/damap/base/integration/pure/FileBasedPureAPI.java
+++ b/src/main/java/org/damap/base/integration/pure/FileBasedPureAPI.java
@@ -6,6 +6,8 @@ import jakarta.enterprise.inject.Typed;
 import jakarta.inject.Inject;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
 import lombok.extern.jbosslog.JBossLog;
 import org.damap.base.rest.config.domain.TenantConfigResolver;
 import org.eclipse.microprofile.rest.client.annotation.RegisterClientHeaders;
@@ -20,17 +22,22 @@ import org.eclipse.microprofile.rest.client.annotation.RegisterClientHeaders;
 @RegisterClientHeaders(PureAuthenticationHeaderFactory.class)
 class FileBasedPureAPI implements PureAPI {
 
+  private final ObjectMapper objectMapper = new ObjectMapper();
+
   @Inject TenantConfigResolver tenantConfigResolver;
 
   @Override
   public PureAPIPaginatedProjectsResponse listAllProjects(Long size, Long offset) {
-    ObjectMapper mapper = new ObjectMapper();
-    try (InputStream in =
-        tenantConfigResolver.getTenantAwareConfig().elsevierPureProjectsFile().openStream()) {
-      return mapper.readValue(in, PureAPIPaginatedProjectsResponse.class);
-    } catch (IOException e) {
-      throw new RuntimeException(e);
+    return singlePage(readAllProjectsFromFile());
+  }
+
+  @Override
+  public PureAPIPaginatedProjectsResponse searchProjects(String q, Long size, Long offset) {
+    List<PureAPIProject> all = readAllProjectsFromFile();
+    if (q != null && !q.isBlank()) {
+      all = all.stream().filter(p -> p != null && p.titleContains(q)).toList();
     }
+    return singlePage(all);
   }
 
   @Override
@@ -43,10 +50,9 @@ class FileBasedPureAPI implements PureAPI {
 
   @Override
   public PureAPIPaginatedPersonsResponse listAllPersons(Long size, Long offset) {
-    ObjectMapper mapper = new ObjectMapper();
     try (InputStream in =
         tenantConfigResolver.getTenantAwareConfig().elsevierPurePersonsFile().openStream()) {
-      return mapper.readValue(in, PureAPIPaginatedPersonsResponse.class);
+      return objectMapper.readValue(in, PureAPIPaginatedPersonsResponse.class);
     } catch (IOException e) {
       throw new RuntimeException(e);
     }
@@ -55,5 +61,32 @@ class FileBasedPureAPI implements PureAPI {
   @Override
   public PureAPIPerson getPerson(String uuid) {
     return listAllPersons().stream().filter(p -> p.getUuid().equals(uuid)).findFirst().orElse(null);
+  }
+
+  private List<PureAPIProject> readAllProjectsFromFile() {
+    try (InputStream in =
+        tenantConfigResolver.getTenantAwareConfig().elsevierPureProjectsFile().openStream()) {
+      PureAPIPaginatedProjectsResponse resp =
+          objectMapper.readValue(in, PureAPIPaginatedProjectsResponse.class);
+      if (resp == null || resp.getItems() == null) {
+        return List.of();
+      }
+      return resp.getItems();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private PureAPIPaginatedProjectsResponse singlePage(List<PureAPIProject> all) {
+    PureAPIPaginatedProjectsResponse page = new PureAPIPaginatedProjectsResponse();
+    page.setCount(all.size());
+
+    PureAPIPageInformation info = new PureAPIPageInformation();
+    info.setOffset(0);
+    info.setSize(all.size());
+    page.setPageInformation(info);
+
+    page.setItems(new ArrayList<>(all));
+    return page;
   }
 }
--- a/src/main/java/org/damap/base/integration/pure/HTTPBasedPureAPI.java
+++ b/src/main/java/org/damap/base/integration/pure/HTTPBasedPureAPI.java
@@ -39,6 +39,36 @@ interface HTTPBasedPureAPI extends PureAPI {
       @QueryParam("size") Long size, @QueryParam("offset") Long offset);
 
   /**
+   * Search projects via {@code POST /projects/search}.
+   *
+   * @param query the search query body.
+   * @return the response containing the matching projects on that page.
+   */
+  @POST
+  @Path("/projects/search")
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Fallback(fallbackMethod = "searchProjectsPostFallback", skipOn = DamapApiException.class)
+  PureAPIPaginatedProjectsResponse searchProjectsPost(PureAPIProjectsQuery query);
+
+  /**
+   * Search projects using pagination. Delegates to {@link #searchProjectsPost} with a {@link
+   * PureAPIProjectsQuery} body.
+   *
+   * @param q query string, passed to Pure API {@code searchString} parameter
+   * @param size the size of the page.
+   * @param offset the offset to start at.
+   * @return the response containing the projects on that page.
+   */
+  @Override
+  default PureAPIPaginatedProjectsResponse searchProjects(String q, Long size, Long offset) {
+    PureAPIProjectsQuery body = new PureAPIProjectsQuery();
+    body.setSearchString(q);
+    if (size != null) body.setSize(size.intValue());
+    if (offset != null) body.setOffset(offset.intValue());
+    return searchProjectsPost(body);
+  }
+
+  /**
    * Retrieve a project with a specific ID.
    *
    * @param uuid the ID of the project.
@@ -104,6 +134,10 @@ interface HTTPBasedPureAPI extends PureAPI {
   }
 
   default PureAPIPaginatedProjectsResponse listAllProjectsFallback(Long size, Long offset) {
+    throw fallback();
+  }
+
+  default PureAPIPaginatedProjectsResponse searchProjectsPostFallback(PureAPIProjectsQuery query) {
     throw fallback();
   }
 
--- a/src/main/java/org/damap/base/integration/pure/PureAPI.java
+++ b/src/main/java/org/damap/base/integration/pure/PureAPI.java
@@ -44,6 +44,16 @@ interface PureAPI {
   PureAPIPerson getPerson(String uuid);
 
   /**
+   * Search projects using pagination.
+   *
+   * @param q query string, passed to Pure API {@code q} parameter
+   * @param size the size of the page.
+   * @param offset the offset to start at.
+   * @return the response containing the projects on that page.
+   */
+  PureAPIPaginatedProjectsResponse searchProjects(String q, Long size, Long offset);
+
+  /**
    * Retrieve all projects, sending multiple queries to the Pure API.
    *
    * @return a list of all projects in the Pure database.
@@ -51,10 +61,38 @@ interface PureAPI {
   default List<PureAPIProject> listAllProjects() {
     List<PureAPIProject> items = new ArrayList<>();
     long offset = 0;
-    long pageSize = 10;
+    long pageSize = 100;
     boolean finished = false;
     while (!finished) {
       PureAPIPaginatedProjectsResponse projects = listAllProjects(pageSize, offset);
+      if (projects == null || projects.items == null || projects.pageInformation == null) {
+        break;
+      }
+      items.addAll(projects.items);
+      offset = projects.pageInformation.offset + projects.pageInformation.size;
+      if (projects.count <= offset) {
+        finished = true;
+      }
+    }
+    return items;
+  }
+
+  /**
+   * Search projects, sending multiple queries to the Pure API.
+   *
+   * @param q query string, passed to Pure API {@code q} parameter
+   * @return a list of matching projects
+   */
+  default List<PureAPIProject> searchAllProjects(String q) {
+    List<PureAPIProject> items = new ArrayList<>();
+    long offset = 0;
+    long pageSize = 100;
+    boolean finished = false;
+    while (!finished) {
+      PureAPIPaginatedProjectsResponse projects = searchProjects(q, pageSize, offset);
+      if (projects == null || projects.items == null || projects.pageInformation == null) {
+        break;
+      }
       items.addAll(projects.items);
       offset = projects.pageInformation.offset + projects.pageInformation.size;
       if (projects.count <= offset) {
@@ -72,10 +110,13 @@ interface PureAPI {
   default List<PureAPIPerson> listAllPersons() {
     List<PureAPIPerson> items = new ArrayList<>();
     long offset = 0;
-    long pageSize = 10;
+    long pageSize = 100;
     boolean finished = false;
     while (!finished) {
       PureAPIPaginatedPersonsResponse persons = listAllPersons(pageSize, offset);
+      if (persons == null || persons.items == null || persons.pageInformation == null) {
+        break;
+      }
       items.addAll(persons.items);
       offset = persons.pageInformation.offset + persons.pageInformation.size;
       if (persons.count <= offset) {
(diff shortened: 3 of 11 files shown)


bitmotivator (commented):
I'm no longer part of the project, but this is almost certainly a bad idea. These classes were intentionally not made public. You are creating an API out of something that is intended purely for use as a mock. This _will_ break when DAMAP changes. The whole idea of introducing a standardized API was to get away from creating implicit APIs in random parts of the application. If DAMAP wants to maintain code quality, it needs to get away from the "everything is public" mentality.

Now, that being said, that is just an opinion, but maybe it helps.

src/main/java/org/damap/base/integration/pure/PureAPI.java
```diff
@@ -51,15 +61,77 @@ interface PureAPI {
...
+        throw new IllegalStateException(
+            "Pure pagination did not advance (projects). offset="
+                + offset
+                + ", returnedOffset="
+                + returnedOffset
+                + ", returnedSize="
+                + returnedSize
+                + ", count="
+                + projects.count);
+      }
```
ValentinFutterer: I think we can trust PURE API to be implemented correctly, or is there a specific reason to include this code?

src/main/java/org/damap/base/integration/pure/PureAPIContentRef.java
```diff
@@ -26,5 +26,5 @@
 })
 abstract class PureAPIContentRef {
   @JsonProperty(required = true)
-  @NotNull String uuid;
+  @NotNull public String uuid;
```
ValentinFutterer: Why is this field set to public?

src/main/java/org/damap/base/integration/pure/PureAPIInternalParticipantAssociation.java
```diff
@@ -17,7 +17,7 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 class PureAPIInternalParticipantAssociation extends PureAPIParticipantAssociation {
   @JsonProperty(required = true)
-  @NotNull PureAPIPersonRef person;
+  @NotNull public PureAPIPersonRef person;
```
ValentinFutterer: Same as above

src/main/java/org/damap/base/integration/pure/PureProjectService.java
```diff
@@ -115,22 +117,79 @@ public ProjectSupplementDO getProjectSupplement(String projectId) {
...
+    long pageSize = 100;
+    long offset = 0;
+    int maxPagesToCheck = 50;
+    int maxResults = 100;
+    int pagesChecked = 0;
+    boolean finished = false;
+
+    try {
+      while (!finished && pagesChecked < maxPagesToCheck && matchingProjects.size() < maxResults) {
+        PureAPIPaginatedProjectsResponse pageResponse = pureAPI.listAllProjects(pageSize, offset);
```
ValentinFutterer: I dont get why you need all of these extra check `listAllprojects` is already paginated, isnt it?

src/main/java/org/damap/base/integration/pure/PureProjectService.java
```diff
@@ -115,22 +117,79 @@ public ProjectSupplementDO getProjectSupplement(String projectId) {
...
+
+      res.setItems(matchingProjects);
+
+    } catch (jakarta.ws.rs.ProcessingException e) {
+      log.error("Error fetching projects from Pure API", e);
+      res.setItems(new ArrayList<>());
+    } catch (Exception e) {
+      log.error("Error fetching projects from Pure API", e);
+      res.setItems(new ArrayList<>());
+    }
```
ValentinFutterer: This is bad error handling, since it just hides the exception. The user wont see any recommended projects and wont be abler to tell if the API is down or not. You can just remove it, There is a open PR for error handling from external APIs.

src/main/java/org/damap/base/integration/pure/FileBasedPureAPI.java
```diff
@@ -57,4 +63,38 @@ public PureAPIPaginatedPersonsResponse listAllPersons(Long size, Long offset) {
...
+      if (resp == null || resp.getItems() == null) {
+        return List.of();
+      }
+      return resp.getItems();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private PureAPIPaginatedProjectsResponse paginateProjects(
```
ValentinFutterer: There really isnt a reason to add pagination to the file based PURE - ts for testing only😅

ValentinFutterer (commented):
Review is just really quick and dirty what i noticed at first glance.

Also you should add caching - this would solve a lot of performance issues.

src/main/java/org/damap/base/integration/pure/PureProjectService.java
```diff
@@ -115,22 +117,79 @@ public ProjectSupplementDO getProjectSupplement(String projectId) {
...
+
+        matchingProjects.addAll(pageMatches);
+
+        long returnedOffset = pageResponse.getPageInformation().getOffset();
+        long returnedSize = pageResponse.getPageInformation().getSize();
+        long nextOffset = returnedOffset + Math.max(1, returnedSize);
+
+        if (nextOffset <= offset) {
+          break;
+        }
```
GeoffreyKarnbach: If no results are available in returnedSize, we have a unnecessary 50 iterations. Maybe add a 

```
if (returnedSize == 0) {
    break;
}
```

src/main/java/org/damap/base/integration/pure/FileBasedPureAPI.java
```diff
@@ -57,4 +63,38 @@ public PureAPIPaginatedPersonsResponse listAllPersons(Long size, Long offset) {
   public PureAPIPerson getPerson(String uuid) {
     return listAllPersons().stream().filter(p -> p.getUuid().equals(uuid)).findFirst().orElse(null);
   }
+
+  private List<PureAPIProject> readAllProjectsFromFile() {
+    ObjectMapper mapper = new ObjectMapper();
```
GeoffreyKarnbach: Can we not just @Inject the mapper to not recreate a new one every time?

src/main/java/org/damap/base/integration/pure/PureAPI.java
lpandath: the reason to include it is not we don’t trust the Pure API at all, but justprotect against a single class of API misbehavior

src/main/java/org/damap/base/integration/pure/PureAPIContentRef.java
lpandath: this one indeed from the privous changes i had, can we revert.

src/main/java/org/damap/base/integration/pure/PureProjectService.java
lpandath: yes true by default, i can just use getRecommended() to use that and just filter the results, should be the same.

src/main/java/org/damap/base/integration/pure/FileBasedPureAPI.java
lpandath: i did included because the PureAPI interface requires listAllProjects(Long size, Long offset) to return proper pagination metadata - without it, the default listAllProjects() would loop somehow. I can resturn everything in a single plage but the idea was not not return everythng in a sinple page without ignoring size and page

src/main/java/org/damap/base/integration/pure/PureProjectService.java
lpandath: yes nice, ill add the eraly break

src/main/java/org/damap/base/integration/pure/FileBasedPureAPI.java
lpandath: Good point. I made it a private final field so it's created once instead of every call. Can't use @Inject since tests instantiate this class directly with new

src/main/java/org/damap/base/integration/pure/PureAPIProject.java
```diff
@@ -51,9 +51,6 @@ boolean titleContains(String text) {
 
   ProjectDO toProjectDO(String descriptionClassificationURI) {
     ProjectDO result = new ProjectDO();
-    if (pureId != null) {
-      result.setId(pureId);
-    }
```
ValentinFutterer, line 56: Good catch!

src/main/java/org/damap/base/integration/pure/HTTPBasedPureAPI.java
```diff
@@ -27,6 +27,20 @@ interface HTTPBasedPureAPI extends PureAPI {
...
+   * @param size the size of the page.
+   * @param offset the offset to start at.
+   * @return the response containing the projects on that page.
+   */
+  @GET
+  @Path("/projects")
+  @Override
+  PureAPIPaginatedProjectsResponse searchProjects(
+      @QueryParam("q") String q, @QueryParam("size") Long size, @QueryParam("offset") Long offset);
+
```
ValentinFutterer, line 72: I dont get this search function here. The [API docs](https://api.elsevierpure.com/ws/api/rapidoc.html#get-/projects) have no mention of a search endpoint, so this is just a second GetAll now, if I am not mistaken.

src/main/java/org/damap/base/integration/pure/FileBasedPureAPI.java
```diff
@@ -57,4 +64,30 @@ public PureAPIPaginatedPersonsResponse listAllPersons(Long size, Long offset) {
...
+    page.setCount(all.size());
+
+    PureAPIPageInformation info = new PureAPIPageInformation();
+    info.setOffset(0);
+    info.setSize(all.size());
+    page.setPageInformation(info);
+
+    page.setItems(new ArrayList<>(all));
+    return page;
+  }
```
ValentinFutterer, line 92: Is there a reason why this isnt used for `listAllPersons`

lpandath: the main improvement here is the pagination handling - previously we did load all projects/persons at once, but now we iterate through paginated responses 100 items per page, which avoids loading large datasets into memory. For listAllPersons, it follows the same pagination pattern but with the persons endpoint - we could combine the singlePage helper to both entities if u think is not appropriate.

src/main/java/org/damap/base/integration/pure/HTTPBasedPureAPI.java
```diff
@@ -40,18 +40,33 @@ PureAPIPaginatedProjectsResponse listAllProjects(
...
    * @param size the size of the page.
    * @param offset the offset to start at.
    * @return the response containing the projects on that page.
    */
-  @GET
-  @Path("/projects")
   @Override
-  PureAPIPaginatedProjectsResponse searchProjects(
-      @QueryParam("q") String q, @QueryParam("size") Long size, @QueryParam("offset") Long offset);
+  default PureAPIPaginatedProjectsResponse searchProjects(String q, Long size, Long offset) {
```
ValentinFutterer, line 65: Minor: Maybe rename this method to `searchProjectsPaginated` or something of a similar nature. If i want to call `searchProjects` from `PureProjectService` i get two methods with the same name but different signatures.

src/main/java/org/damap/base/integration/pure/PureProjectCache.java
```diff
@@ -0,0 +1,21 @@
...
+@ApplicationScoped
+class PureProjectCache {
+
+  @Inject PureAPI pureAPI;
+
+  @CacheResult(cacheName = "pure-all-projects")
+  public List<PureAPIProject> getAllProjects() {
+    return pureAPI.listAllProjects();
+  }
+}
```
ValentinFutterer: This extra class is annoying. I dont quite get the justification for it - i dont see the self invocation you mention. You could just cache `default List<PureAPIProject> listAllProjects()`, which doesnt have the self invocation problem, as its directly called by the service. I think the real issue is that PureAPI is a HTTPBasedRestClient. Can you please find the real cause and update your comment.

ValentinFutterer: Ok I read the docs more in depth and found this: https://quarkus.io/guides/cache#generating-a-cache-key-with-cachekeygenerator

With this we dont need this class anymore - we just need one key generator class for all pure caches. 
This also allows direct caching of `getRecommended`, since we can include the userId into the key generation directly. And more importantly, this also allows for making the chaching multitenancy safe by including the tenant-id as a key. Otherwise tenants can read each others caches. Other chaches are not affected, so for now only pure needs this extra layer of protection.

ValentinFutterer (commented):
Almost done. What i dont get is why implement caching only for `listAllProjects()` 😅. Project search and the Person methods should also be chached - especially search, as it will be called more often.

During rebasing, you missed the new error handling annotations and fallback method. they should be added to search, check HTTPBasedPureAPI fore reference.

src/main/java/org/damap/base/integration/pure/PureProjectCache.java
lpandath: Replaced PureProjectCache with a PureCacheKeyGenerator, the reason a separate class was needed is that Quarkus REST client proxies dont apply CDI interceptors to  default interface methods. Caching is now directly on PureProjectService.getRecommended() and search() using the generator, which also includes tenant+userId in the key for multitenancy safety.

getRecommended and search are now both cached via @CacheResult with PureCacheKeyGenerator and renamed to searchAllProjects(String q) to match the listAllProjects() naming pattern.

lpandath: for the rebase the searchProjectsPostFallback fallback method defined and also @Fallback(fallbackMethod = "searchProjectsPostFallback", skipOn = DamapApiException.class)

src/main/java/org/damap/base/integration/pure/PureCacheKeyGenerator.java
```diff
@@ -0,0 +1,51 @@
...
+      return affiliation != null ? affiliation : "default";
+    } catch (UnauthorizedException e) {
+      return "default";
+    }
+  }
+
+  private String getUserIdSafe() {
+    String userId = securityService.getUserId();
+    return userId != null ? userId : "anonymous";
+  }
```
ValentinFutterer, line 41: I dont get why you added those methods, why not use the defaults in security service?

ValentinFutterer (commented):
Can you please add caching to all methods like I asked - persons do have the same problem as projects.

src/main/java/org/damap/base/integration/pure/PureCacheKeyGenerator.java
lpandath: Your right, I was focused on resolving the null-safety issue and did not consider that getUserId() already handles that il inline it. Ill also add caching to PurePersonService as you suggested.

ValentinFutterer: `getTenantSafe` is still here. Is the purpose to always have a string returned, because the key doesnt take null. If yes, the remove the notloggedin check, since its done in getAffiliation and return the string "single-tenant" to make it clearer what happens here.

Or you can wait for this PR https://github.com/damap-org/damap-backend/pull/461
It will add a "isMultitenancyDisabled" method to the security service, where you can only add the affiliation in case of multitenncy and otherwise leave it out.

src/main/java/org/damap/base/integration/pure/PureCacheKeyGenerator.java
```diff
@@ -0,0 +1,42 @@
...
+class PureCacheKeyGenerator implements CacheKeyGenerator {
+
+  @Inject SecurityService securityService;
+
+  @Override
+  public Object generate(Method method, Object... methodParams) {
+    List<Object> key = new ArrayList<>();
+    key.add(getTenantSafe());
+    String userId = securityService.getUserId();
+    key.add(userId != null ? userId : "anonymous");
```
ValentinFutterer: This doesnt work and will remove inter user caching, since the userid will be unique for every user. The userID is only relevant for the getRecommended method caching, since there we do filtering dependant on the id. You need to creeate a separate generate method for this case. Would also be easier to just not pass the userid in case its null (although something else went wrong in this case entirely.

src/main/java/org/damap/base/integration/pure/PurePersonService.java
```diff
@@ -24,6 +25,7 @@ public class PurePersonService implements PersonService {
 
   /** {@inheritDoc} */
   @Override
+  @CacheResult(cacheName = "pure-read-person", keyGenerator = PureCacheKeyGenerator.class)
```
ValentinFutterer, line 28: read persons has a cahche but read projects doesnt for some reason.

src/main/java/org/damap/base/integration/pure/PurePersonService.java
```diff
@@ -34,6 +36,7 @@ public ContributorDO read(String id) {
 
   /** {@inheritDoc} */
   @Override
+  @CacheResult(cacheName = "pure-search-persons", keyGenerator = PureCacheKeyGenerator.class)
```
ValentinFutterer, line 39: You forgot to add the cachenames to the application.yaml, does this code even run?

ValentinFutterer: Apparently it autocreates chaches that never expire - pls fix.

ValentinFutterer (commented):
Time is really running out and i dont want to go through everything with a fine-toothed comb every review cycle (which we had 6 of). If you want to we can seperate the chaching from the PR and just merge your other changes and someone else takes over the caching, otherwise this will probably not make it in the release.

lpandath:
> Time is really running out and i dont want to go through everything with a fine-toothed comb every review cycle (which we had 6 of). If you want to we can seperate the chaching from the PR and just merge your other changes and someone else takes over the caching, otherwise this will probably not make it in the release.

Thanks for the feedback. To clarify on the cache key changes: the getTenantSafe method and the userId-in-all-keys issue both came directly from solutions suggested during this review cycle I  wasn't aware of the single-tenant behavior until it was pointed out todays meeting. I've now applied those suggestions and all tests pass.                                                                     
                                                                                                                                                                                                       
I understand the time pressure and I'd prefer to keep the caching in this PR rather than hand it off mid-way the fixes are minor and already done. Happy to address any remaining points quickly if you can point them out.

ValentinFutterer:
To clarify what i meant to say with my feedback about the id and affiliation:
> This also allows direct caching of getRecommended, since we can include the userId into the key generation directly. And more importantly, this also allows for making the chaching multitenancy safe by including the tenant-id as a key.

These two sentences are meant to be read separately - adding the userid into the key allows us to cache getRecommended specifically, which wouldnt be possible otherwise since it includes the filtering code.

`chaching multitenancy safe by including the tenant-id as a key` was meant generically for all keys.

ValentinFutterer (approved):
Looks good!

PR #463, opened 2026-02-16
Helm: point DAMAP default tags to release candidate

SotosTsepe:
Since the release candidate is now available to pull, we can safely reference it in the default values and avoid using `next`, which is unpredictable and can cause failures.

Also, apply requests and limits for the tenant Job, OpenShift requires it.


ValentinFutterer (approved):
👍

PR #467, opened 2026-02-16
Helm(fix): add missing environment variables for multitenancy

SotosTsepe:
When enabling multitenancy, the backend crashes on startup with:
Datasource 'bootstrap' is not configured

This happens because the bootstrap datasource was defined but did not read the values from the default datasource environment variables, so Quarkus could not connect to the main DB to initialize tenant databases.

### DIFF
--- a/helm/templates/03_damap.yaml
+++ b/helm/templates/03_damap.yaml
@@ -136,7 +136,7 @@ spec:
             {{- if .Values.damap.multitenancy.enabled }}
             - name: QUARKUS_DATASOURCE_BOOTSTRAP_DB_KIND
               value: postgresql
-            - name: QUARKUS_DATASOURCE_BOOTSTRAP_DRIVER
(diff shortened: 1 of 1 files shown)


helm/templates/03_damap.yaml
```diff
@@ -131,12 +131,29 @@ spec:
                   key: dbPassword
 
             {{- /*
-            Quarkus Profile for multitenancy
+            Quarkus config for multitenancy
             */}}
             {{- if .Values.damap.multitenancy.enabled }}
+            - name: QUARKUS_DATASOURCE_BOOTSTRAP_DB_KIND
+              value: postgresql
+            - name: QUARKUS_DATASOURCE_BOOTSTRAP_DRIVER
```
ValentinFutterer:
```suggestion
            - name: QUARKUS_DATASOURCE_BOOTSTRAP_JDBC_DRIVER
```

ValentinFutterer (approved):
👍

PR #471, opened 2026-02-25
Admin: Add customize templates feature

BlueTechie:
## Description
Feature

#### What does this PR do?
This PR introduces a new feature: DAMAP admins can now customize templates for their users. Admins can add any number of custom templates that are a variation of the Science Europe template. These custom templates will be treated by the backend like a Science Europe template would. Admins can also disable any template they wish, as long as one template is active. As a last point, admins can also delete custom templates.

- Database: 
  - Introduced new tables `export_template` and `export_template_seq` in a new ChangeLog. 
  - The 3 default templates (Science Europe with `templateId 1`, FWF with `templateId 2` and Horizon Europe with `templateId 3`) are seeded in the database but the Data column is Null. This is because in the case of these 3 default templates, the template is fetched from the backend's classpath instead of the database. 
  - The custom templates have `templateId > 3` and are stored as BLOBs in the database.
- Domain & DTOs:
  - Created new entity `ExportTemplate` that lazy-loads BLOBs of `.docx` files.
  - New `ExportTemplateDO` and `ExportTemplateDOMapper`, implemented so that they exclude file data for better performance. Otherwise the frontend would unnecessarily load all file contents when in the admin page.
- Service Layer:
  - New `ExportTemplateService` with CRUD operations and possibility to toggle template activation.
  - Service layer also enforces "At least one template active" rule (like the frontend).
  - Uses `FitsService` to verify that the binary file signature during upload is that of a `.docx` file.
- Export Logic Refactoring:
  - Updated `ExportTemplateBroker` to route exports based on database IDs. instead of hardcoded Enums (since any number of custom need to be supported).
- Config API:
   - Updated `ConfigResource` so that it includes template metadata.

#### Dependencies

Frontend PR: https://github.com/damap-org/damap-frontend/pull/519

closes GH-


src/main/java/org/damap/base/domain/ExportTemplate.java
```diff
@@ -0,0 +1,32 @@
...
+@EqualsAndHashCode(
+    callSuper = true,
+    exclude = {"data"})
+public class ExportTemplate extends PanacheEntity {
+
+  @Column(name = "name", nullable = false)
+  private String name; // e.g., "Standard FWF" or "Science Europe - Special Version"
+
+  @Column(name = "template_key", nullable = false)
+  private String templateKey; // "FWF", "HORIZON", or "SCIENCE_EUROPE"
```
ValentinFutterer: This is a String value, but the ETemplateType enum still exists - which had the Job of the string before your changes.

I believe you could just stick to the Enum instead of string or was there a good reason to choose string? I think with string you might be able to autochoose custom templates depending on project types if we refactor the logic. SO it might be good to stick to string for future advancements. 

But the backend logic still relies on ETemplateType in some places, e.g. in TemplateSelectorServiceImpl. I believe this could also be the cause for the crashes when you deactivate a template and try to export a dmp of the templates project type. You need the refactor the whole backend to use strings for template types (and also rewrite the comments that denote the possible types, since we might expand them in the future. We could make the funders customizable e.g. and then link custom funders to custom dmp types. Write this in the comments so its clear why we use strings and not enums.

src/main/java/org/damap/base/conversion/ExportTemplateBroker.java
ValentinFutterer: Unused now

ValentinFutterer: Same

src/main/java/org/damap/base/conversion/ExportTemplateBroker.java
```diff
@@ -51,27 +51,21 @@ public ExportTemplateBroker(
    * @param dmpId a long
    * @return a {@link org.apache.poi.xwpf.usermodel.XWPFDocument} object
    */
```
ValentinFutterer, line 52: Update javadocs

src/main/java/org/damap/base/rest/document/service/DocumentService.java
```diff
@@ -29,16 +30,17 @@ public class DocumentService {
   @Inject @RestClient GotenbergRestService gotenbergRestService;
 
   public StreamingOutput getExportDocument(
-      long dmpId, ETemplateType template, boolean download, String filetype) {
+      long dmpId, Long templateId, boolean download, String filetype) {
```
ValentinFutterer: `download is unused`

src/main/java/org/damap/base/rest/DmpDocumentResource.java
```diff
@@ -38,25 +38,25 @@ public class DmpDocumentResource {
...
   @Path("/{dmpId}")
   @Deprecated(since = "4.3.0", forRemoval = true)
   public Response exportTemplate(
-      @PathParam("dmpId") long dmpId, @QueryParam("template") ETemplateType template) {
+      @PathParam("dmpId") long dmpId, @QueryParam("template") Long templateId) {
     log.info("Return DMP document file for DMP with id=" + dmpId);
 
-    return this.export(dmpId, template, true, "docx");
+    return this.export(dmpId, templateId, true, "docx");
   }
```
ValentinFutterer: Lets just remove this, its deprecated and no one should be using it right now anyways

src/main/java/org/damap/base/conversion/ExportScienceEuropeTemplate.java
```diff
@@ -1,21 +1,33 @@
...
   /**
    * exportTemplate.
    *
    * @param dmpId a long
+   * @param templateId the unique identifier of the template record in the database. This ID
+   *     determines which .docx file is used as the basis:
+   *     <ul>
+   *       <li>ID 1: Default Science Europe template (falls back to classpath resource).
+   *       <li>IDs 2, 3: Default funder templates (FWF, Horizon Europe).
+   *       <li>ID &gt; 3: Custom variations of the Science Europe template uploaded by admins.
```
ValentinFutterer: What if we add a 4th supported template

src/main/java/org/damap/base/conversion/ExportTemplateBroker.java
```diff
@@ -51,27 +51,21 @@ public ExportTemplateBroker(
...
-  public XWPFDocument exportTemplate(long dmpId) {
-    return exportTemplateByType(
-        dmpId, templateSelectorService.selectTemplate(dmpService.getDmpById(dmpId)));
-  }
+  public XWPFDocument exportTemplate(long dmpId, Long templateId) {
+    if (templateId == null || templateId == 0) {
+      return exportScienceEuropeTemplate.exportTemplate(dmpId, 1L);
```
ValentinFutterer: Passing this `1L` is very close to a magic number code smell

src/main/java/org/damap/base/rest/ConfigResource.java
```diff
@@ -108,6 +113,13 @@ public ConfigDO config() {
     List<Image> images = imageThemeService.getImages();
     configDO.setImages(images);
 
+    List<ExportTemplate> entities = ExportTemplate.listAll();
+    List<ExportTemplateDO> dtos =
+        entities.stream()
+            .map(entity -> ExportTemplateDOMapper.mapEntityToDO(entity, new ExportTemplateDO()))
+            .toList();
+    configDO.setTemplates(dtos);
+
```
ValentinFutterer: Minor: The naming for entities and dtos is a little dangerous. What happens when you add aother entities and dtos to the config object? I would at least name them `templateEntities` or somethign similar.

Or you just put the list based mapping into the mapper. This would keep the `ConfigResource` cleaner.

src/main/java/org/damap/base/rest/AdminResource.java
```diff
@@ -136,4 +138,36 @@ public void deleteRecommendedRepository(@RestPath Long id) {
     log.info("Deleting recommended repository with id: " + id);
     this.recommendedRepositoryService.deleteRecommendedRepository(id);
   }
+
+  @GET
+  @Path("/export-templates")
+  @RolesAllowed("${damap.auth.admin-role-name}")
+  public List<ExportTemplate> getExportTemplates() {
+    return ExportTemplate.listAll();
+  }
```
ValentinFutterer: Unused by the frontend, i believe

src/main/java/org/damap/base/domain/ExportTemplate.java
ValentinFutterer: Also, templateKey is a bad naming and doesnt really tell you anything about the functionality. Maybe exportType, templateCategory or something similar would be better + comment on what it actually does

src/main/java/org/damap/base/rest/AdminResource.java
```diff
@@ -136,4 +138,36 @@ public void deleteRecommendedRepository(@RestPath Long id) {
...
+
+  @POST
+  @Path("/export-templates")
+  @RolesAllowed("${damap.auth.admin-role-name}")
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
+  public ExportTemplate uploadTemplate(
+      @RestForm("name") String name,
+      @RestForm("category") String category,
```
ValentinFutterer: Isnt category always `SCIENCE_EUROPE` ?
I think you can omit the paramter for now

src/main/java/org/damap/base/rest/AdminResource.java
```diff
@@ -136,4 +138,36 @@ public void deleteRecommendedRepository(@RestPath Long id) {
...
+      @RestForm("file") FileUpload file) {
+    return this.exportTemplateService.uploadTemplate(name, category, file);
+  }
+
+  @PUT
+  @Path("/export-templates/{id}/toggle")
+  @RolesAllowed("${damap.auth.admin-role-name}")
+  public void toggleTemplate(@RestPath Long id) {
```
ValentinFutterer: Weird naming choice

src/main/java/org/damap/base/rest/ExportTemplateService.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  @Transactional
+  public void toggleStatus(Long id) {
+    ExportTemplate t = ExportTemplate.findById(id);
+    if (t != null) {
+      if (t.isActive()) {
+        long activeCount = ExportTemplate.count("active = true");
+        if (activeCount <= 1) {
+          throw new jakarta.ws.rs.BadRequestException("At least one template must remain active.");
```
ValentinFutterer: You can remove the `jakarta.ws.rs.`

src/main/java/org/damap/base/rest/AdminResource.java
ValentinFutterer: Should be `PATCH` instead of `PUT`, since `PUT` implies complete replacement of the resource

ValentinFutterer (commented):
In general pretty good, but if you refactor logic, you need to do it consequentially througout the whole application. Few fixes, then we are good to go.

ValentinFutterer:
Ah and please also add the size check to the bakcend

ValentinFutterer (approved):
Looks good now

PR #473, opened 2026-02-26
Config: Fix small copy-paste errors in example.env

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?
Fixes copy-paste errors in the example.env file and adds a small amount of documentation to it.

#### Code review focus
The helm chart already uses the correct env variables, but I couldnt find the `DAMAP_PERSON_SERVICES_DISPLAY_TEXT`, `DAMAP_PERSON_SERVICES_QUERY_VALUE` and `DAMAP_PERSON_SERVICES_CLASS_NAME` variables in the helm chart. Can you check if the helm chart uses the correct variables? (I renamed them from SERVICE to SERVICES)

### DIFF
--- a/docker/example.env
+++ b/docker/example.env
@@ -18,17 +18,18 @@ QUARKUS_HTTP_CORS_ORIGINS=http://localhost:8085
 #   - default
 #   - elsevier-pure
 DAMAP_PROJECTS_SERVICE=default
-# Text shown in the frontend to allow users to switch between person services
-DAMAP_PERSON_SERVICE_DISPLAY_TEXT=University
+# Text shown  in the frontend to allow users to switch between person services
+DAMAP_PERSON_SERVICES_DISPLAY_TEXT=University
 # Sets which person service is used
 # Allowed values:
 #   - UNIVERSITY
 #   - PURE
-DAMAP_PERSON_SERVICE_QUERY_VALUE=UNIVERSITY
+DAMAP_PERSON_SERVICES_QUERY_VALUE=UNIVERSITY
 # Which java class is to be used as person service
 # Set to org.damap.base.integration.pure.PurePersonService for PURE
 # Or to your custom class, if you do use a custom integration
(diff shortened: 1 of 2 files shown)


docker/example.env
```diff
@@ -13,22 +13,23 @@ DAMAP_TITLE="DAMAP Tool"
 DAMAP_ENV=DEV
 # Comma-separated list of allowed browser origins for CORS
 QUARKUS_HTTP_CORS_ORIGINS=http://localhost:8085
-# Sets the project service used, should be set to elsevier-pure only when pure integration is used
+# Sets the project service used, should be set to els evier-pure only when pure integration is used
```
SotosTsepe: fix this

SotosTsepe (approved):

SotosTsepe:
Regarding the question for the helm chart, it uses the yaml value from the values.yaml, then converts it to JSON, then passes it as an environment variable, rather than passing each subitem individually, cf. https://github.com/damap-org/damap-backend/blob/next/helm/templates/03_damap.yaml#L220

I still have doubts about passing sub-items of a config object. Does this scale? For example, if one needs to pass 3 different list objects, can the subitems be overridden with an index for each item in the list?

docker/example.env
ValentinFutterer: Fixed

ValentinFutterer:
> Regarding the question for the helm chart, it uses the yaml value from the values.yaml, then converts it to JSON, then passes it as an environment variable, rather than passing each subitem individually, cf. https://github.com/damap-org/damap-backend/blob/next/helm/templates/03_damap.yaml#L220
> 
> I still have doubts about passing sub-items of a config object. Does this scale? For example, if one needs to pass 3 different list objects, can the subitems be overridden with an index for each item in the list?

I had to remove the json environment variables sadly, due to quarkus not playing ball. There will be a PR explaining this. So  this is not relevant anymore.

PR #474, opened 2026-03-06
Helm(fix): update outdated custom domain value in damap ingress & fix custom build refs

SotosTsepe:
This fix renames the outdated `custom_domain` value to `customDomain`.

This affects OpenShift deployments using a custom domain; the chart rendered the default OpenShift `Route` resources instead of the expected `Ingress` resources using the OpenShift `ingressClass` (fetched from the `values.yaml`).

The custom build was also fixed, referencing the correct image registry from OpenShift.
Note: If running the chart with kustomize (as in our case), remember to specify the namespace explicitly in the charts section ( {{ .Release.namespace }} may not be correctly populated in the custom build section)

### DIFF
--- a/helm/templates/03_damap.yaml
+++ b/helm/templates/03_damap.yaml
@@ -77,7 +77,7 @@ spec:
         - name: damap-backend
 
           {{- if and .Values.openshift (ne .Values.customBuild.backend.name "") }}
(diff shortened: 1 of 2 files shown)


ValentinFutterer (approved):
Looks good

PR #476, opened 2026-03-10
Maintenance: External API usage in export protected with catch-try

### LINKED ISSUE #470: External API dependecies crash DAMAP during export
**Changes**
Right now, if e.g. re3data is down and we try to export, DAMAP crashes with a 500 error. This is of course bad, since we tie our tool to the whims of the re3data API. To remedy this, first find all instances of external APIs being called during the export. Then wrap them all with try catch. If you catch an error, due to an external service being unavailable, replace the text with a reminder that we couldnt fetch the info there. E.g. re3data is down, so in the place of the repository description, we write something like re3data was not available. This shopuld be the description fetched from re3data. To get this information, please try tzo export at a a later date. (For the exact verbiage, consult Christiane)

**Additional notes**
This solution was discussed with Tomasz.

**Acceptance Criteria**
Export does not crash when external dependecies are down. Instead of the text that would have been fetched, a remindertext is exported. Also please write comments onw hy this try catch is there.

GeoffreyKarnbach:
## Description
Maintenance

#### What does this PR do?
This PR adds error handling and fallback values to the document export process. Previously, if external APIs (like the Project Service or re3data) were unreachable, the export would fail entirely. Now, it catches the exceptions, logs the error, and inserts placeholder text so the user still gets their document.

Specific changes:
  * Wrapped the `projectService.read()` call in a try-catch so the title page generation survives a Project Service failure.
  * Wrapped `repositoriesService` calls (fetching repo descriptions and URLs) in a try-catch. If the re3data API fails, it outputs a fallback message into the document explaining that re3data was unavailable.
  * Added a fallback `ContributorDO` for the project coordinator if the Project API throws an exception. The exported document will now show "Project Leader (Information currently unavailable from Project Service)" instead of leaving it blank or crashing.

closes GH-470

### DIFF
--- a/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
+++ b/src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
@@ -105,13 +105,15 @@ public abstract class AbstractTemplateExportScienceEuropeComponents
     // add funding program to funding item variables
     ProjectDO projectCRIS = null;
     if (project.getUniversityId() != null) {
-      try { // This try catch is there to prevent export crashing when an external dependency is down
+      try { // This try catch is there to prevent export crashing when an external dependency is
+        // down
         projectCRIS = projectService.read(project.getUniversityId());
       } catch (Exception e) {
         ProjectDO fallback = new ProjectDO();
         FundingDO dummyFunding = new FundingDO();
-        dummyFunding.setFundingProgram("Project service was not available. This should be the funding program fetched" +
-                " from the project service. To get this information, please try to export at a later date.");
+        dummyFunding.setFundingProgram(
+            "Project service was not available. This should be the funding program fetched"
+                + " from the project service. To get this information, please try to export at a later date.");
         fallback.setFunding(dummyFunding);
         projectCRIS = fallback;
         log.error("Could not reach Project Service for funding details", e);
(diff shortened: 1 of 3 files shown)


src/main/java/org/damap/base/conversion/AbstractTemplateExportScienceEuropeComponents.java
```diff
@@ -103,8 +103,13 @@ public void titlePage() {
...
     ProjectDO projectCRIS = null;
-    if (project.getUniversityId() != null)
-      projectCRIS = projectService.read(project.getUniversityId());
+    if (project.getUniversityId() != null) {
+      try {
+        projectCRIS = projectService.read(project.getUniversityId());
+      } catch (Exception e) {
+        log.error("Could not reach Project Service for funding details", e);
+      }
+    }
```
ValentinFutterer, line 121: Like you did in AbstractTemplateExportSetup, please create a dummy ProjectDO holding the replacement texts.

ValentinFutterer (commented):
Like asked in the Issue description, please write comments on why we need the try-catch statements. 

Also, I found another usage of the repositoriesService in `AbstractTemplateExportScienceEuropeComponents` in the method `composeTableDatasetPublication`. Be more mindful next time and recheck all the export files, you need to catch all occurences otherwise this PR wont fix anything and is justa  waste of code.

And please rename the commit to `try-catch` instead of `catch-try` 😅

ValentinFutterer (approved):
Mergeable

Issue #395, opened 2025-06-15
Remove and replace FITS service


bitmotivator:
The FITS service in its current configuration is unsuitable for multiple reasons:

1. The container is huge.
2. The container also requires us to run it as root, which is not suitable for production.

Since we are only using MIME type detection, this could be achieved with a much more lightweight process, for example using an implementation for Java's built-in `FileTypeDetector` class.

While doing this, the API should also be changed to be a non-leaky abstraction.

bitmotivator:
Note: this has moved up in urgency because the FITS service requires root permissions and is not deployable on ARI&Snet, which blocks DAMAP Cloud.

ValentinFutterer:
ImageThemeResource also has custom file type detection code thats not using the fitsservice. Parts of it can maybe be reused for the new file type detection service thats not using the fits service container. 
The file type detection in ImageThemeResource should then obviously be replaced with the new service.

ValentinFutterer:
https://github.com/damap-org/damap-backend/pull/426#discussion_r2732110136

https://github.com/damap-org/damap-backend/pull/426#discussion_r2732638424

PR #477, opened 2026-03-11
Replace FITS service

BlueTechie:
## Description

Refactoring
#### What does this PR do?

 This PR replaces the external FITS service with a Java library (Apache Tika). This is because the `islandora/fits` image takes up more than 1.6 GB and is only used for Mime type detection, which can be done by a Java library instead.

Changes in this PR:
- Replaced `FitsService`, `FitsMapper` and `FitsRestService` by a reusable `FileAnalysisService` (using Apache Tika) and a dedicated `FileAnalysisMapper` to handle file metadata.
- Removed the `FitsServiceAvailable` boolean in the config.
- Renamed the api endpoint from `/api/fits/examine` to `/api/file-analysis/examine`.
-  Updated `ImageThemeService` to use the new `FileAnalysisService` for image uploads. Previously, the backend relied on reading the file extension for Mime type detection, which is insecure. Apache Tika relies on magic bytes instead.
- Deleted some auto-generated files related to FITS.

#### Dependencies

Frontend PR: https://github.com/damap-org/damap-frontend/pull/522

closes GH-https://github.com/damap-org/damap-backend/issues/395


src/main/java/org/damap/base/rest/theme/service/ImageThemeService.java
```diff
@@ -1,32 +1,36 @@
...
+
   public enum AllowedImageType {
     PNG("image/png"),
     JPEG("image/jpeg"),
     JPG("image/jpg"),
     GIF("image/gif"),
     SVG("image/svg+xml"),
-    WEBP("image/webp");
+    WEBP("image/webp"),
+    ICO("image/x-icon");
```
ValentinFutterer: Why was ICO added to the list?

src/main/java/org/damap/base/rest/file_analysis/mapper/FileAnalysisMapper.java
```diff
@@ -0,0 +1,82 @@
...
+    if (size <= 100000000000L) return 100000000000L;
+    if (size <= 500000000000L) return 500000000000L;
+    if (size <= 1000000000000L) return 1000000000000L;
+    if (size <= 5000000000000L) return 5000000000000L;
+    if (size <= 10000000000000L) return 10000000000000L;
+    if (size <= 100000000000000L) return 100000000000000L;
+    if (size <= 500000000000000L) return 500000000000000L;
+    if (size <= 1000000000000000L) return 1000000000000000L;
+    return 1000000000000001L;
+  }
```
ValentinFutterer, line 84: The mapping is wrong and produces one sizerange too high. Look at `data-spec.ts` to look at how to stay in the range. Or we justremove the mapping entirely and just stick with the real size thats read.

src/main/java/org/damap/base/rest/file_analysis/mapper/FileAnalysisMapper.java
```diff
@@ -0,0 +1,82 @@
...
+      return type;
+    }
+
+    switch (mimetype) {
+      case "application/msword":
+      case "application/vnd.openxmlformats-officedocument.wordprocessingml.document":
+      case "application/vnd.ms-excel":
+      case "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet":
+      case "application/pdf":
+        type.add(EDataType.STANDARD_OFFICE_DOCUMENTS);
```
ValentinFutterer, line 39: Tika read docx files as application/x-tika-ooxml, which is an office docuemnt of unknown type if i understand it correctly. You need to fix somehting with Tika, e.g. it also reads xlsx as application/zip.

ValentinFutterer (commented):
Looks pretty good. The Tika library needs to be configured better, as it doesnt detect Office documents perfectly.

SotosTsepe:
@ValentinFutterer looks good now, tested on the staging cluster, ready to merge from my side :+1:

ValentinFutterer (approved):
We are good to go 👍

PR #478, opened 2026-03-11
Multitenancy: Replace SecurityAugmentor by ContainerRequestFilter to cache users in DB, while being compatible with multitenancy

GeoffreyKarnbach:
## Description
Refactoring / Bugfix

#### What does this PR do?
This PR resolves a compatibility issue between the OIDC user synchronization and multitenancy. 

* **Before:** User syncing was handled by `DamapSecurityAugmentor`. Because security augmentors run extremely early in the Quarkus reactive pipeline, it attempted to execute database operations before the multitenancy context was fully established, causing tenant routing errors.
* **Now:** Replaced the augmentor with `UserSyncFilter` (implementing `ContainerRequestFilter`). This ensures the user sync executes *after* authentication and tenant resolution are complete, allowing it to safely write to the correct tenant's database.

*(Note: This solution was tested locally with the multitenancy setup active and successfully passed, routing and saving users to the correct tenant databases.)*


SotosTsepe:
Tested on OpenShift, unfortunately doesn't work:
[message.txt](https://github.com/user-attachments/files/26190625/message.txt)

I still had to guard the null tenantId (before login) to make it work.

SotosTsepe:
This should be fine :+1:

ValentinFutterer (approved):
👍

PR #481, opened 2026-03-30
Multitenancy: Make no affiliation claim throw a 401

ValentinFutterer:
## Description

Refactoring

#### What does this PR do?
Instead of just failing with a nullpointer exception when ffiliation is missing, we now throw a 401 error to make it more verbose.


SotosTsepe (approved):
Much better now 👍

PR #493, opened 2026-04-27
feat: add instance configuration for public availability and consent form

GeoffreyKarnbach:
## Description
Feature

#### What does this PR do?

Adds backend support for singleton instance-level configuration.

This PR introduces:
- a new `instance_config` table with `public_available` and `consent_form_enabled`
- a Panache entity and repository for instance configuration
- admin GET/PUT endpoints for reading and updating instance config
- inclusion of instance config flags in the public application config response
- Liquibase changelog entries for the new table and default singleton row

The new config allows admins to control whether an instance is publicly available and whether the consent form is enabled.

#### Dependencies

Requires corresponding frontend changes to:
- consume `publicAvailable` and `consentFormEnabled`
- guard non-admin routes when the instance is not public
- expose admin toggles for both config values
- hide/skip the consent form when disabled


ValentinFutterer (approved):
👍

PR #494, opened 2026-04-28
Translation: Update consent data retention duration text

ValentinFutterer:
## Description

Translation update

#### What does this PR do?
Consent form satted that user data was only stored during the tools usage. Chnaged the wording to reflect that we store user accounts now.

#### Breaking changes
Liquibase


ValentinFutterer:
Solved by https://github.com/damap-org/damap-backend/pull/498

PR #497, opened 2026-05-04
dev: add multitenancy users to keycloak instance and local multitenancy markdown guide

GeoffreyKarnbach:
### Type
Config

### What does this PR do?

This PR improves the local developer experience for multitenancy by providing a complete, ready-to-use setup.

It introduces:
- A new **`DEV_MULTITENANCY.md`** guide explaining how to run DAMAP locally in multitenant mode
- Configuration instructions in markdown guide
- Updates to the Keycloak realm export:
  - adds an **`affiliation` protocol mapper** so the claim is available in JWTs
  - adds **example users and admins for two tenants (`tenant_1`, `tenant_2`)**
    - including proper `affiliation` attributes for tenant resolution

The goal is to allow developers to spin up a working multitenant environment locally with minimal manual setup.

---

### Dependencies

- Relies on frontend/backend logic that uses the **`affiliation` claim** for tenant resolution
- Requires local Postgres databases (`tenant_1`, `tenant_2`) to exist (see MD guide for command)

---

### Notes

- Default users (`user`, `admin`) are intentionally not affiliated with any tenant and cannot be used in multitenant mode
- Additional test users can be added via `/docker/sample-damap-realm-export.json`

### DIFF
--- a/DEV_MULTITENANCY.md
+++ b/DEV_MULTITENANCY.md
@@ -1,0 +1,99 @@
+# Local Multitenancy Setup
+
+## 1. Copy Tenant Configuration
+
+Copy the tenant configuration file from `/docker` into your application resources:
+
+```bash
+cp docker/tenants.yaml src/main/resources/tenants.yaml
+```
+
+---
+
+## 2. Configure Application
+
+Update `application.yaml` (multitenant profile):
+
+```yaml
+"%multitenant":
+  config:
+    locations: /tenants.yaml
+```
+
+---
+
+## 3. Configure Tenant Datasources
+
+Edit `tenants.yaml` and replace the datasource section with:
+
+```yaml
+datasource:
+  tenant_1:
+    db-kind: postgresql
+    username: damap
+    password: pw4damap
+    jdbc:
+      url: jdbc:postgresql://localhost:8088/tenant_1
+
+  tenant_2:
+    db-kind: postgresql
+    username: damap
+    password: pw4damap
+    jdbc:
+      url: jdbc:postgresql://localhost:8088/tenant_2
+```
+
+---
+
+## 4. Create Tenant Databases
+
+Run inside the Postgres container:
+
+```bash
+psql -U damap -d postgres -c "CREATE DATABASE tenant_1 OWNER damap;"
+psql -U damap -d postgres -c "CREATE DATABASE tenant_2 OWNER damap;"
+```
+
+---
+
+## 5. Start the Application
+
+Start Docker (DB + Keycloak), then run:
+
+```bash
+mvn quarkus:dev -Dquarkus.profile=dev,multitenant
+```
+
+---
+
+## 6. Test Users
+
+```
+tenant_1 user:
+  username: tenant1-user
+  password: password
+
+tenant_2 user:
+  username: tenant2-user
+  password: password
+
+tenant_1 admin:
+  username: tenant_1_admin
+  password: password
+
+tenant_2 admin:
+  username: tenant_2_admin
+  password: password
+```
+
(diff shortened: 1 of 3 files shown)


DEV_MULTITENANCY.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+---
+
+## 2. Configure Application
+
+Update `application.yaml` (multitenant profile):
+
+```yaml
+"%multitenant":
```
lpandath, line 18: since application.yaml is a tracked file if you accidentally commit this change it could cause issues. Maybe add a note not to commit it, or use the CLI flag instead: 

` Dquarkus.config.locations=classpath:tenants.yaml`

DEV_MULTITENANCY.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Notes
+
+- Default users (`user`, `admin`) are **not associated with any tenant** and will not work.
+- Tenant resolution depends on the **affiliation claim** in the JWT.
+- If tenants are not detected, verify:
+    - `tenants.yaml` exists in `src/main/resources`
+    - `config.locations` is set to `/tenants.yaml`
+- You can create or modify test users by editing `/docker/sample-damap-realm-export.json`.  
+  After changes, restart Keycloak (or re-import the realm) to apply them.
```
lpandath: Missing newline at end of file.

DEV_MULTITENANCY.md
```diff
@@ -0,0 +1,103 @@
...
+tenant_1 user:
+  username: tenant1-user
+  password: tenant1-user
+
+tenant_2 user:
+  username: tenant2-user
+  password: tenant2-user
+
+tenant_1 admin:
+  username: tenant1_admin
```
lpandath, line 87: minor: the table says tenant1_admin (underscore) but the realm export has tenant1-admin , same for tenant_2_admin vs tenant2-admin.

docker/sample-damap-realm-export.json
```diff
@@ -2325,6 +2340,106 @@
...
+      "firstName": "Tenant",
+      "lastName": "One User",
+      "email": "tenant1-user@example.com",
+      "realmRoles": [
+        "offline_access"
+      ],
+      "attributes": {
+        "personID": ["tenant1-user"],
+        "affiliation": [
+          "member@tenant_1",
```
lpandath, line 2357: Maybe worth adding a small note somewhere in the docsexplaining the member@tenant_1 format it's not obvious why it needs the @ part. The frontend splits by @ to get the tenant ID, so it would break if someone just puts tenant_1.

lpandath (approved):
Tested this PR locally. Overall the setup works well once everything is in place tenant resolution via the affiliation claim works, each tenant gets its own database with Liquibase migrations running  automatically on first login, and the instance config is separate per tenant.
 
A few things that tripped me up during setup:
- the Keycloak realm import gets skipped if you already have a damap realm running had to figure out the fresh DB requirement manually
- the psql commands in Step 4 need docker exec to run inside the container
- the affiliation format (member@tenant_1) isn't obvious 

once those were sorted, was able to log in as tenant1-admin, toggle public availability, and access the app as tenant1-user with data properly isolated to the tenant_1 database. 

Good addition for local dev!

DEV_MULTITENANCY.md
ValentinFutterer: Good catch, I messed that up 😅

DEV_MULTITENANCY.md
ValentinFutterer: Good point

PR #502, opened 2026-05-11
Multitenancy: Fix bug where affiliation dashes did not get replaced

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?
I assumed that affiliations would only contain dots, but apparently they also can contain dashes. I now used a regex which replaces every non alphanumerical character with an underscore and removes double underscores, if they should occur.

#### Breaking changes
Yes, potentially changes how affiliations are resolved, but it should not "break" anything except our test instance on staging.


SotosTsepe (approved):
looks good 👍

PR #512, opened 2026-05-27
translation: only allow for ISO 639-1 language codes for application translations and internal storage translations. remove old e_language reference table for internal storage translations, migrate "eng" to "en", "deu" to "de"

GeoffreyKarnbach:
## Description

Feature / Bugfix / Database migration

#### What does this PR do?

This PR aligns app translation and internal storage translation language-code handling with ISO 639-1 two-letter language codes.

Previously, app translation languages could be created with arbitrary input, while internal storage translations were restricted to the legacy values `eng` and `deu` through backend validation and a database foreign key to `e_language`. This caused mismatches between app translations and internal storage translations once additional languages were added.

This PR changes that by:

- adding backend validation for ISO 639-1 two-letter language codes
- normalizing app translation language codes before creation
- normalizing and validating internal storage translation language codes before creation/update
- replacing the old internal storage validation that only allowed `eng` and `deu`
- adding translation keys for frontend validation messages
- migrating existing internal storage translation values:
  - `eng` → `en`
  - `deu` → `de`
- removing the old `e_language` database table restriction by dropping the related foreign keys and table

This makes app translations and internal storage translations use the same language-code format and prevents invalid language codes from being created through direct backend requests.

#### Breaking changes

Potentially breaking for consumers or existing data that still rely on the legacy internal storage language codes `eng` and `deu`.

A database migration is included to migrate existing values to the new ISO 639-1 format:

- `eng` becomes `en`
- `deu` becomes `de`

The old `e_language` lookup table is removed, so validation is now handled in the backend service layer instead of through the database foreign key.

#### Dependencies

Requires corresponding frontend changes to:

- use the same ISO 639-1 language-code list
- replace free-text/dropdown language fields with autocomplete
- submit normalized two-letter language codes such as `en`, `de`, `fr`
- handle migrated internal storage translations using `en` / `de` instead of `eng` / `deu`

### DIFF
--- a/src/main/java/org/damap/base/rest/storage/InternalStorageTranslationValidator.java
+++ b/src/main/java/org/damap/base/rest/storage/InternalStorageTranslationValidator.java
@@ -9,6 +9,7 @@ import org.damap.base.domain.InternalStorage;
 import org.damap.base.domain.InternalStorageTranslation;
 import org.damap.base.repo.InternalStorageRepo;
 import org.damap.base.repo.InternalStorageTranslationRepo;
+import org.damap.base.rest.translation.service.LanguageCodeValidator;
 
 @UtilityClass
 @JBossLog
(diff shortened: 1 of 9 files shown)


ValentinFutterer (approved):
The changelogs where pretty messy - otherwise good.

Issue #408, opened 2025-08-04
Generic OpenAPI-based integration


bitmotivator:
In order to get DAMAP Cloud working and still offer custom integrations, we must create a generic HTTP-based integration as described in [this blog post](https://damap.org/blog/2025/07/15/an-early-look-at-the-common-dmp-api/). This integration should use the RDA DMP common API objects, see #416.

PR #401, opened 2025-07-07
Common DMP Standard / OpenAPI-based integration points

bitmotivator:
This PR adds OpenAPI-based integration points to DAMAP rather than Java-based ones. DAMAP Cloud users will be able to use this to build custom integrations without needing to inject code into DAMAP.

## TODO

- [ ] Documentation
- [ ] Tests
- [ ] application.yaml updates
- [ ] Integrate error handling from #419

## Endpoints

For more details see [the OpenAPI spec](https://github.com/damap-org/damap-backend/blob/bitmotivator/openapi-integration/src/main/resources/org/damap/base/integration/rest/openapi.yaml).

### `GET /projects?q={search}&offset={offset}&count={count}`

This endpoint will let you return a `Project` object in the [common standard format](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard) to auto-populate project data when creating DMPs.

### `GET /projects/{id}`

Returns a single `Project` object  in the [common standard format](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard) to auto-populate project data when creating DMPs.

### `GET /projects/{id}/contributors?q={search}&offset={offset}&count={count}`

This endpoint will let you return a `Contributor` object in the [common standard format](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard) to auto-populate contributor data.

### `GET /persons?q={search}&offset={offset}&count={count}`

Returns a list of `Contact` objects in the [common standard format](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard) to auto-populate project data when creating DMPs.

### `GET /persons/{id}`

Returns a single `Contact` object in the [common standard format](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard) to auto-populate project data when creating DMPs.

### Why not auto-generate the client code?

I could not get the microprofile client generation to work properly in conjunction with Quarkus.

### Why chain two OpenAPI generators?

The first one generates the common standard data models, the second one generates the DAMAP-specific integrations with extra fields. The first one will be used in the future to implement the common API.

PR #513, opened 2026-05-27
Make DAMAP RDA API compliant (+ OSTrails Application Profile)

### LINKED ISSUE #416: RDA Common API implementation
In order to increase interoperability with other systems, we would like to implement the [RDA DMP common API](https://github.com/RDA-DMP-Common/common-madmp-api) in DAMAP.

BlueTechie:
## Description

Feature
#### What does this PR do?

This PR adds OpenAPI-based integration endpoints to DAMAP based on the RDA DMP Common Standard (v1.2) and the [OSTrails application profile extension](https://docs.ostrails.eu/en/next/commons/dmp/application-profile.html#ostrails-ap-extension). This enables external systems to discover, retrieve, and create DMPs without requiring custom Java code modifications.

**Some implementation details**:
- Auto-generated classes: RDA Java classes are generated with the [OpenAPI Generator Maven plugin](https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator-maven-plugin), which pulls from the local `openapi.yaml` file. The generator is configured to enforce Bean Validation (`useBeanValidation`) and automatically strip empty/null optional fields using `@JsonInclude(NON_EMPTY)`
- OSTrails AP Extensions: Added support for the extended schema properties directly in `openapi.yaml` (DMP version, license, rights, dataset methodology, conforms_to, cost units/types, etc.).
- Bidirectional Mappers: The mappers (`ProjectMapper.java`, `DMPMapper.java` etc.) are bidirectional between the DAMAP data model and the RDA data model. For mapping, a `strict` mode is introduced: since the DAMAP data model only supports a single funder, grant etc., importing an RDA-format DMP which has multiple funders etc. with `strict` mode on throws an error.
- Defensive Fallbacks: All mappers are hardened with defensive fallback logic to guarantee that mandatory RDA fields (such as `title`, `created`/`modified` timestamps, `contact`, and `dataset_id`) are never null in the output json.
- DMP Search & Querying: Add support for DMP search queries with filtering (languages, funder IDs, grant IDs, and embargo dates) using the service layer.

**New DMP endpoints** (with an `/api/rda/` prefix):
- `GET /dmps?offset={offset}&count={count}&query={query}...`
Lists and filters DMPs in the RDA format (including support for filters such as languages, date of modification etc.).

- `POST /dmps`
Creates or imports a new DMP from a standard-compliant JSON payload.

- `GET /dmps/{id}`
Retrieves a single DMP in the RDA format.

- `PUT /dmps/{id}`
Overwrites an existing DMP.

- `DELETE /dmps/{id}`
Deletes a DMP.

See https://github.com/damap-org/damap-backend/pull/401 for more info about the first 4 commits.
#### Breaking changes

closes GH-416 and GH-408


src/main/java/org/damap/base/rda/dmpcommonstandard/IdentifierMapper.java
```diff
@@ -0,0 +1,37 @@
+package org.damap.base.rda.dmpcommonstandard;
+
+import org.damap.base.enums.EIdentifierType;
+import org.damap.base.rest.dmp.domain.IdentifierDO;
+
+class IdentifierMapper {
```
ValentinFutterer, line 6: This class needs documentation. I assume its meant to generate an Identifier from just an ID string - but at https://github.com/damap-org/damap-backend/pull/513/changes#diff-1ffaa3324fecf9be6b8417a5ed798024ef8d031c755af5cb25bf3f8ac7d91510R55 a type is passed?

src/main/java/org/damap/base/rda/dmpcommonstandard/ContributorMapper.java
```diff
@@ -0,0 +1,238 @@
...
+
+    if (contributor.getAffiliation() != null && !contributor.getAffiliation().isEmpty()) {
+      var rdaAffiliation = contributor.getAffiliation().get(0);
+      result.setAffiliation(rdaAffiliation.getName());
+
+      if (rdaAffiliation.getAffiliationId() != null) {
+        var damapAffiliationId = new IdentifierDO();
+        damapAffiliationId.setIdentifier(rdaAffiliation.getAffiliationId().getIdentifier());
+        if (rdaAffiliation.getAffiliationId().getType() != null) {
+          damapAffiliationId.setType(IdentifierMapper.getIdentifierDO(rdaAffiliation.getAffiliationId().getType()).getType());
```
ValentinFutterer: A type is passed to identifiermapper, but it expects an identifer string i think, this needs clarification.

src/main/java/org/damap/base/rest/rda/service/RdaDmpService.java
```diff
@@ -0,0 +1,757 @@
...
+    private boolean isOnOrBefore(Date date, OffsetDateTime boundary) {
+        if (boundary == null) {
+            return true;
+        }
+
+        OffsetDateTime value = toOffsetDateTime(date);
+        return value != null && (value.isBefore(boundary) || value.isEqual(boundary));
+    }
+
+    private boolean matchesCreatedAfter(DmpDO dmp, OffsetDateTime createdAfter) {
```
ValentinFutterer: These could be two functions instead of four, just call it with different arguments

src/main/java/org/damap/base/rest/rda/service/RdaDmpService.java
```diff
@@ -0,0 +1,757 @@
...
+
+    private boolean matchesDmpIds(DmpDO dmp, List<String> dmpIds) {
+        if (dmpIds == null || dmpIds.isEmpty()) {
+            return true;
+        }
+
+        String internalId = String.valueOf(dmp.getId());
+
+        return dmpIds.stream().anyMatch(id -> equalsIgnoreCase(id, internalId));
+    }
```
ValentinFutterer: This is not about internal id, its about an identifier, like a doi https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard?tab=readme-ov-file#properties-in-dmp_id

src/main/java/org/damap/base/rest/rda/service/RdaDmpService.java
```diff
@@ -0,0 +1,757 @@
...
+        ensureDmpExists(dmpId);
+
+        if (!accessValidator.canDeleteDmp(dmpId, personId)) {
+            throw new ForbiddenException(unauthorizedMessage(dmpId));
+        }
+
+        dmpService.delete(dmpId);
+    }
+
+    private void validateDocument(DMPDocument document) {
```
ValentinFutterer: This method is unused

src/main/resources/application.yaml
```diff
@@ -157,6 +160,8 @@ rest:
 "%dev":
   damap:
     title: "DAMAP (Development)"
+    generic-cris-endpoint-url: ${GENERIC_CRIS_URL:http://localhost:8888/}
+    generic-cris-api-key: ${GENERIC_CRIS_API_KEY:api-key-here}
```
ValentinFutterer: There is no need for this notaton, quarkus automatically generates env variables.

ValentinFutterer (approved):
It will do for our purposes for now, but we need to put in serious work to make it more maintainable.

PR #514, opened 2026-05-27
Add evaluation service connection

DanceaVlad:
#### What does this PR do?
Implement a Rest Client for connecting to the evaluation service. Read the URL for it in `application.yaml`. If the URL is missing, the `ConfigDTO` reports that the evaluation service is disabled.

#### Breaking changes
Modification of the `ConfigDTO`

#### Code review focus
Rest Client architecture

#### Dependencies
[Frontend PR 552](https://github.com/damap-org/damap-frontend/pull/552)


src/main/java/org/damap/base/rest/evaluation/EvaluationRemoteResource.java
```diff
@@ -0,0 +1,93 @@
...
+import org.damap.base.rest.evaluation.dto.EvaluationMultipartBodyDTO;
+import org.damap.base.rest.evaluation.dto.EvaluationResultDTO;
+import org.eclipse.microprofile.faulttolerance.Fallback;
+import org.eclipse.microprofile.faulttolerance.Timeout;
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+import org.jboss.logging.Logger;
+
+/** EvaluationRemoteResource interface. */
+@RegisterRestClient(configKey = "rest.evaluation")
+@Timeout(30000)
```
ValentinFutterer: Why `@Timeout(30000)` and not `@Timeout(10000)`?
Half a minute is really long and user will have to wait a long time for feedback.

src/main/resources/application.yaml
```diff
@@ -66,12 +66,12 @@ damap:
...
-        class-name: 'org.damap.base.integration.mock.MockUniversityPersonServiceImpl'
-      - display-text: 'ORCID'
-        query-value: 'ORCID'
-        class-name: 'org.damap.base.integration.orcid.ORCIDPersonServiceImpl'
+      - display-text: "University"
+        query-value: "UNIVERSITY"
+        class-name: "org.damap.base.integration.mock.MockUniversityPersonServiceImpl"
+      - display-text: "ORCID"
+        query-value: "ORCID"
+        class-name: "org.damap.base.integration.orcid.ORCIDPersonServiceImpl"
```
ValentinFutterer, line 76: Out of curiosity, was there a reason for this change?

src/main/java/org/damap/base/rest/EvaluationResource.java
```diff
@@ -0,0 +1,74 @@
...
+   * @param dmpId a long
+   * @param benchmark a {@link java.lang.String} object
+   * @return a {@link java.util.List} of {@link
+   *     org.damap.base.rest.evaluation.dto.EvaluationResultDTO} objects
+   */
+  @POST
+  @Path("/assess/{dmpId}")
+  public List<EvaluationResultDTO> assessBenchmark(
+      @PathParam("dmpId") long dmpId, @QueryParam("benchmark") String benchmark) {
+    log.info("Run evaluation for DMP with id: " + dmpId + " and benchmark: " + benchmark);
```
ValentinFutterer, line 62: Shouldnt this be a GET?

ValentinFutterer: Ah ops, you are  sending data

ValentinFutterer (approved):
Looks pretty solid, good work.
I reverted the vibecoded quickfix, we should fix that in a later release. I also deactivated the evaluation feature by default, as its a little experimental right now.

PR #515, opened 2026-06-02
Add repository step dataset/retention translations

lpandath:
Partner changeset for damap-frontend#538. Inserts the new repository-step keys

## Description
  Config / DB (translations)

#### What does this PR do?
Partner changeset for damap-frontend#538. The new repository-step UI keys introduced in that PR (dataset/deposit/retention table headers, retention assigned/unassigned tooltips, search, recommended headline) had no rows in the translation table, so they would render as raw keys at runtime.

This adds Liquibase changeset changeLog-4.7.0_15.yaml (id 39) which inserts those keys across all existing languages mirroring the cross-language pattern in changeLog-4.7.0_13.yaml and updates the default_value of dmp.steps.repositories.find.other from "Find other repository" to "Search other Repositories".

#### Breaking changes
  None. Data-only migration; no schema, table, or column changes.

#### Dependencies
  - damap-frontend#538 this PR supplies the translations consumed there.

### Checks

- [ ] Export updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-

### DIFF
--- a/src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_15.yaml
+++ b/src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_15.yaml
@@ -1,0 +1,107 @@
+databaseChangeLog:
+    - changeSet:
+          id: 39
+          author: Laura
+          comment: Insert translations for the repository step dataset/retention table (frontend PR #538)
+          changes:
+              - sql: |
+                    INSERT INTO translation (translation_key, language, default_value, active, version)
+                    SELECT
+                      'dmp.steps.repositories.table.header.dataset',
+                      l.language,
+                      'Dataset',
+                      l.active,
+                      0
+                    FROM (SELECT DISTINCT language, active FROM translation) l;
+
+              - sql: |
+                    INSERT INTO translation (translation_key, language, default_value, active, version)
+                    SELECT
+                      'dmp.steps.repositories.table.header.deposit',
+                      l.language,
+                      'Deposit in repository',
+                      l.active,
+                      0
+                    FROM (SELECT DISTINCT language, active FROM translation) l;
+
+              - sql: |
+                    INSERT INTO translation (translation_key, language, default_value, active, version)
+                    SELECT
+                      'dmp.steps.repositories.table.header.retention',
+                      l.language,
+                      'Minimum retention period (all repositories)',
+                      l.active,
+                      0
+                    FROM (SELECT DISTINCT language, active FROM translation) l;
+
+              - sql: |
+                    INSERT INTO translation (translation_key, language, default_value, active, version)
+                    SELECT
+                      'dmp.steps.repositories.retention.assigned',
+                      l.language,
+                      'The retention period applies to this dataset across all repositories.',
+                      l.active,
+                      0
+                    FROM (SELECT DISTINCT language, active FROM translation) l;
+
+              - sql: |
+                    INSERT INTO translation (translation_key, language, default_value, active, version)
+                    SELECT
+                      'dmp.steps.repositories.retention.unassigned',
+                      l.language,
+                      'Assign this dataset to at least one repository to set a retention period.',
+                      l.active,
+                      0
+                    FROM (SELECT DISTINCT language, active FROM translation) l;
+
+              - sql: |
+                    INSERT INTO translation (translation_key, language, default_value, active, version)
+                    SELECT
+                      'dmp.steps.repositories.search',
+                      l.language,
+                      'Search',
+                      l.active,
+                      0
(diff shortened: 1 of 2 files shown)


src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_15.yaml
```diff
@@ -0,0 +1,80 @@
...
+                      l.language,
+                      'Recommended Repositories',
+                      l.active,
+                      0
+                    FROM (SELECT DISTINCT language, active FROM translation) l;
+
+              - sql: |
+                    UPDATE translation
+                    SET default_value = 'Search other Repositories'
+                    WHERE translation_key = 'dmp.steps.repositories.find.other';
```
ValentinFutterer: You should never use SQL in a liquibase script, unless tere is a very good reason - consult the documentation for more information.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_15.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+              - sql: |
+                    INSERT INTO translation (translation_key, language, default_value, active, version)
+                    SELECT
+                      'dmp.steps.repositories.search',
+                      l.language,
+                      'Search',
+                      l.active,
+                      0
+                    FROM (SELECT DISTINCT language, active FROM translation) l;
```
ValentinFutterer, line 65: This is not a new key, but a replacement for `dmp.steps.repositories.filters`. So either you do an update, or you delete the old key. If you leave it, old unused keys will clutter up the translations over time.

ValentinFutterer (commented):
Please do not use SQL if it cannto be avoided like with adding new keys. Also make sure to always clean up unused keys, so they dont build up in he db and bog down the translation feature.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_15.yaml
lpandath: @ValentinFutterer 
Kept the SQL only for the new keys. A native insert can only add one row, but a key needs to exist for every language, and admins make their own languages per instance so I can't know them. The select-distinct-language way is the only safe one across DBs - it's what the README translation part says to do and what changeLog-4.7.0_13 (id 37) already does.

ValentinFutterer: Its not a completely new key though - its a replacement. There are already entries in the database, therefore you can just use a liquibase update statement without SQL.

The SQL statement is only needed if you want to add a completely new key. 
If you leave it like you have now, the old key will stay in the database and pollute the feature.

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_15.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+              - sql: |
+                    INSERT INTO translation (translation_key, language, default_value, active, version)
+                    SELECT
+                      'dmp.steps.repositories.recommended.headline',
```
ValentinFutterer, line 70: We use the `view.button` to structure keys for the primary and secondary view.

ValentinFutterer (commented):
See the comments

src/main/resources/org/damap/base/db/changeLog-4.x/changeLog-4.7.0_15.yaml
lpandath: @ValentinFutterer  Went through all the keys again. Every key I added in the frontend has a matching insert in the backend changeset so 9 of them, and all are actually used in the components. find.other is a value change so it's a native update. question.datasets is removed in the frontend and now has a matching delete in the backend - that was the one missing before, it's added now. So no leftovers in either direction. 

  This is my list of the keys and what I'm doing - please correct me if I'm missing something:

  | Frontend change | Key | Backend op | Used in code |
  |---|---|---|---|
  | added | `table.header.dataset` | INSERT | repo.component |
  | added | `table.header.deposit` | INSERT | repo.component |
  | added | `table.header.retention` | INSERT | repo.component |
  | added | `retention.assigned` | INSERT | repo.component |
  | added | `retention.unassigned` | INSERT | repo.component |
  | added | `search` | INSERT | repo-table |
  | added | `recommended.headline` | INSERT | repo.component |
  | added | `view.button.recommended` | INSERT | repo-instruction |
  | added | `view.button.other` | INSERT | repo-instruction |
  | value changed | `find.other` | UPDATE | repo.component |
  
  
note: updated frontend

PR #516, opened 2026-06-03
Include guidance in assessment results

DanceaVlad:
Feature: Include Guidance in evaluation results

### What does this PR do?

Results from the Evaluation Service can now include a new field called "guidence". This object is now sent to the frontend.

[Frontend PR](https://github.com/damap-org/damap-frontend/pull/554)


ValentinFutterer (approved):
Looks good!

PR #518, opened 2026-06-04
Remove Helm related files

SotosTsepe:
The helm chart is now being actively developed at a dedicated repository:
https://github.com/damap-org/damap-chart


ValentinFutterer (approved):
👍

PR #521, opened 2026-06-17
Rds/add rda host mapping

BlueTechie:
## Description

Feature

#### What does this PR do?

With these changes, DAMAP can now import and export hosts in the final RDA compliant JSON DMP.

**Bidirectional Mapping logic**
*   **Export:** A new distribution is created for each internal/external storage and repository, which itself contains a host, in line with the RDA common standard.
*   **Import:** Incoming hosts are dynamically reconstructed into either `ExternalStorageDO` (if a URL is present) or a standard `RepositoryDO` and correctly linked back to their datasets. Note that a best-effort approach is used.

Unfortunately, DAMAP's architecture regarding hosts can be confusing: For context, `Distribution` is the parent object of `Host`. But in DAMAP, `InternalStorage`, `ExternalStorage` and `Repository` objects all inherit from `Host`, and each of the three has some variations and inconsistencies between what is in the UI and the underlying data structure, for example: 
- For internal storage, there is a UI field `backup_frequency`, but not in the corresponding Domain Object. 
- `ExternalStorage` also has no `url` field in the UI, which is a mandatory field in the RDA standard for a host (but there is a `url` field in the corresponding DO). As a temporary workaround, this field gets assigned the value `https://google.com`.

In addition to the mandatory fields, `available_until`,  `backup_frequency` and `backup_type` (the name is misleading, as this will contain the specified backup location value) are mapped wherever possible.

**IMPORTANT NOTE**: The changes introduced by this PR are only meant to be temporary, since this code is not maintainable in the long run. A future PR will try to automate the mappers as much as possible.

closes GH-


src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -72,8 +76,11 @@ public DMPMapper(
     this.datasetMapper = datasetMapper;
   }
 
-  public DMPWithID convert(DmpDO dmp) {
-    return new DMPWithID().id(String.valueOf(dmp.getId())).dmp(convertData(dmp));
+  public DMPWithID convert(
+      DmpDO dmp, org.damap.base.r3data.RepositoriesService repositoriesService) {
```
ValentinFutterer:
```suggestion
      DmpDO dmp, RepositoriesService repositoriesService) {
```

This is just unnecessary clutter.
But is there even a reason to pass this as an argument, nothing here is static. Why not just inject the service?

src/main/java/org/damap/base/rest/rda/service/RdaDmpService.java
```diff
@@ -503,7 +506,7 @@ private void ensureDmpExists(long dmpId) {
 
   private DMPWithID toRdaDmp(DmpDO dmp) {
     try {
-      return dmpMapper.convert(dmp);
+      return dmpMapper.convert(dmp, repositoriesService);
```
ValentinFutterer: Same as above, why not just inject in the mapper itself? This leaves the method head cleaner.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
ValentinFutterer, line 29: Is there a reason why those are not injected?

src/main/java/org/damap/base/rest/rda/service/RdaDmpService.java
ValentinFutterer, line 39: Is there a reason why injection is not used?

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -135,6 +142,43 @@ private DMPData convertData(DmpDO dmp) {
...
+
+              distributions.add(clone);
+            }
+          }
+          rdaDataset.setDistribution(distributions);
+        }
+      }
     } else {
       result.setDataset(new ArrayList<>());
     }
```
ValentinFutterer: Like discussed in person, this code should be its own mapper and needs comments explainign whats happening here and why, and waht assumptions we make due to the DAMAP data model.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -192,9 +236,26 @@ private void convertData(DMPData data, DmpDO target) {
...
+          var rdaHost = distribution.getHost();
+          if (rdaHost != null) {
+            importHost(target, rdaHost, datasetDO.getReferenceHash());
+          }
+        }
+      }
 
-    if (!datasets.isEmpty()) {
-      target.setDatasets(datasets.stream().map(datasetMapper::convert).toList());
+      target.setDatasets(damapDatasets);
```
ValentinFutterer, line 321: Same

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+      DmpDO dmp,
+      DatasetDO datasetDO,
+      org.damap.base.r3data.RepositoriesService repositoriesService) {
+
+    List<Host> hosts = new ArrayList<>();
+    if (datasetDO == null) {
+      return hosts;
+    }
+
+    String refHash = datasetDO.getReferenceHash();
```
ValentinFutterer: We are trying to remove refhash in the future - why not just use the dataset id?

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+    String idStr = datasetDO.getId() != null ? String.valueOf(datasetDO.getId()) : null;
```
ValentinFutterer: id cannot be null to my knowledge, as it comes directly from the db.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+
+    // 1. Search Repositories
+    if (dmp.getRepositories() != null) {
```
ValentinFutterer: This is not necessary, its always an empty list.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+      for (var repo : dmp.getRepositories()) {
+        if (repo.getDatasets() != null) {
+          if ((refHash != null && repo.getDatasets().contains(refHash))
+              || (idStr != null && repo.getDatasets().contains(idStr))) {
+
+            String repoUrl = "https://google.com";
```
ValentinFutterer: Missing comment on why this fallback exists

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+                if (re3Data != null && !re3Data.getRepository().isEmpty()) {
+
+                  // Get the repositoryURL from the schema
+                  String url = re3Data.getRepository().get(0).getRepositoryURL();
+                  if (url != null && !url.isBlank()) {
+                    repoUrl = url;
+                  }
+                }
+              } catch (Exception e) {
+                // Silent fallback
```
ValentinFutterer: We should ask Tomasz if we want the silent google url fallback, throw an error, or use a fallbacl url that menioned an error.
How this is currently coded means that we do not have a consistent export.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+              }
+            }
+
+            hosts.add(
+                new Host()
+                    .title(repo.getTitle() != null ? repo.getTitle() : "Repository")
+                    .url(repoUrl));
+          }
```
ValentinFutterer: You could / should use the title from re3data

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+    // 2. Search External Storage
+    if (dmp.getExternalStorage() != null) {
+      for (var ext : dmp.getExternalStorage()) {
+        if (ext.getDatasets() != null) {
+          if ((refHash != null && ext.getDatasets().contains(refHash))
+              || (idStr != null && ext.getDatasets().contains(idStr))) {
+            hosts.add(
+                new Host()
+                    .title(ext.getTitle() != null ? ext.getTitle() : "External Storage")
+                    // TODO: Map this to the real URL in the future (currently no url supported)
```
ValentinFutterer: What does this comment mean

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+                    .url(
+                        ext.getUrl() != null && !ext.getUrl().isBlank()
+                            ? ext.getUrl()
+                            : "https://google.com")
+
+                    // TODO: This is misleading but technically allowed by the RDA common standard
```
ValentinFutterer: Good comment 👍

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+        if (store.getDatasets() != null) {
+          if ((refHash != null && store.getDatasets().contains(refHash))
+              || (idStr != null && store.getDatasets().contains(idStr))) {
+
+            String url = "https://google.com";
+            String backupLocation = null;
+
+            if (store.getInternalStorageId() != null) {
+              try {
+                org.damap.base.domain.InternalStorage internalStore =
```
ValentinFutterer:
```suggestion
               InternalStorage internalStore =
```

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+                    org.damap.base.domain.InternalStorage.findById(store.getInternalStorageId());
```
ValentinFutterer: Same

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+                  if (internalStore.getUrl() != null && !internalStore.getUrl().isBlank()) {
+                    url = internalStore.getUrl();
+                  }
+                  if (internalStore.getBackupLocation() != null
+                      && !internalStore.getBackupLocation().isBlank()) {
+                    backupLocation = internalStore.getBackupLocation();
+                  }
+                }
+              } catch (Exception e) {
+              }
```
ValentinFutterer: There should be a comment here.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+            new Host()
+                .title(store.getTitle())
+                .url(url)
+                // TODO: This is misleading but technically allowed by the RDA common standard
+                .backupType(backupLocation));
+      }
+    }
+
+    return hosts;
+  }
```
ValentinFutterer: I dont get why this is here and what it does exactly. It also seems like there is a lot of code duplication in this section. Can you clarify and refactor of necessary.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+  private Distribution cloneDistribution(Distribution source) {
+    Distribution target = new Distribution();
+    target.setTitle(source.getTitle());
+    target.setDataAccess(source.getDataAccess());
+    target.setAvailableUntil(source.getAvailableUntil());
+    target.setByteSize(source.getByteSize());
+    target.setFormat(source.getFormat());
+    target.setLicense(source.getLicense());
+    return target;
+  }
```
ValentinFutterer: This breaks when the distribution classchanges. I would move it into the class itself as a clone constructor.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+
+  private void importHost(DmpDO target, Host rdaHost, String refHash) {
```
ValentinFutterer: Again I would stay away from refhash if its not 100% necessary-

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -216,4 +277,255 @@ private void convertData(DMPData data, DmpDO target) {
...
+    if (host == null || dmp.getRepositories() == null) {
+      return false;
+    }
+    for (var repo : dmp.getRepositories()) {
+      if (host.getTitle() != null && host.getTitle().equalsIgnoreCase(repo.getTitle())) {
+        return true;
+      }
+    }
+    return false;
+  }
```
ValentinFutterer: You can just check the class

ValentinFutterer (commented):
This code here is very similar to your `findAllHostsForDatasets` https://github.com/damap-org/damap-backend/blob/60eedde313ea0c99269e8227eb5335647f7fe39a/src/main/java/org/damap/base/rest/dmp/mapper/DmpDOMapper.java#L112-L152

It uses classes to check the host type, maybe it helps/ you can refactor the classes and reuse code?

Also please start writing javadocs for your methods :)

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
BlueTechie: Yes the mappers are not injected for 2 reasons:
1) If they were injected, we would not be able to pass the strict flag as a single parameter when instantiating the mappers: we would need to add it as a parameter to every single method signature. This would be too cumbersome in my eyes.
2) An injected mapper would mean only one instance at runtime, but what if we want to run a mapper in a strict mode and one in non-strict mode? This is especially relevant considering multi-tenancy.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
BlueTechie: After discussion with Tomasz, any repository host that does not have a URL is skipped and the code does not throw an error but a warning is written in the logs.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
BlueTechie: I believe this is not possible unfortunately, since `Distribution` is the automatically generated class from the RDA Common Standard API spec.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
BlueTechie: From what I could see in debugging, when we trigger an export, the JPA database entities are converted to DOs, but instead of taking the dataset ids that are stored in the database, the `hostDO.getDatasets()` get populated with the refHash values of the datasets. This means that I only get the refHash of the dataset and to get a matching with the array I need to be comparing refHash to refHash. I don't see another way of solving this within the scope of this PR?

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
BlueTechie: I believe it is necessary in this case: when we import a new RDA DMP, the incoming datasets are transient and have not been saved to the database yet, meaning their database id is null. To preserve the relationships between these new datasets and their hosts in memory before they are saved, we have to generate and use the refHash as a temporary link. Without it, the hosts and datasets would be saved disconnected, I think.

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
BlueTechie: I deleted this method 👍

src/main/java/org/damap/base/rest/rda/service/RdaDmpService.java
BlueTechie: (see comment above)

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
BlueTechie: To clean up the method signatures as suggested, I have removed `RepositoriesService` from all the mapping method parameters entirely.

Because our mappers are instantiated manually with `new` (to configure the `strict` flag), the standard `@Inject` inside `HostsMapper` would resolve to `null`. So I used Quarkus's programmatic lookup (`Arc.container().instance(RepositoriesService.class).get()`) directly inside the `mapRepository` method where the service is used.

src/main/java/org/damap/base/rest/rda/service/RdaDmpService.java
BlueTechie: (see comment above)

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
ValentinFutterer: Sadge :/

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
ValentinFutterer: Ah good point, i forgot 👍

ValentinFutterer: That still makes it brittle, but I dont see an easy way around it.
maybe the library can generate a clone constructor?

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
ValentinFutterer: I see, I didnt consider the strict flag

src/main/java/org/damap/base/rda/dmpcommonstandard/DMPMapper.java
```diff
@@ -153,14 +162,27 @@ private DMPData convertData(DmpDO dmp) {
     return result;
   }
 
+  /** RDA to DAMAP (Import). * */
+  public DmpDO convert(DMPWithID dmp) {
+    var dmpDO = new DmpDO();
+    dmpDO.setId(Long.valueOf(dmp.getId()));
```
ValentinFutterer: Why set the ID on import, shouldnt the database set it?

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
```diff
@@ -0,0 +1,404 @@
...
+  public void importHost(DmpDO target, Host rdaHost, String refHash) {
+    String title = !rdaHost.getTitle().isBlank() ? rdaHost.getTitle() : "Imported Host";
+    String url = rdaHost.getUrl();
+
+    if (!url.isBlank()) {
+      importAsExternalStorage(target, title, url, refHash, rdaHost);
+    } else {
+      importAsRepository(target, title, refHash);
+    }
+  }
```
ValentinFutterer, line 477: Is it correct that InternalStorage cannot be imported?

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
```diff
@@ -0,0 +1,404 @@
...
+   * method assumes an implicit 1-to-1 relationship and maps that single host to the dataset. If the
+   * repository host has no URL, then it is skipped since that is a mandatory field.
+   *
+   * <ul>
+   *   <li>Repositories retain the 'availableUntil' retention date from the base distribution.
+   *   <li>Internal and external storages have their 'availableUntil' date cleared to null, as there
+   *       is no support for it in the DAMAP model.
+   * </ul>
+   */
+  private void applySingleHostFallback(
```
ValentinFutterer, line 162: When will this realistically happen? Distribution is not madnatory in the common standard right? So why is it necessary? Sorry if you told me already, then I have forgotten 😅

ValentinFutterer (commented):
Only smaller things + the fixes to the import we discussed in person.

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
BlueTechie: In the newest commit, the mapper now checks first if the host is a `repository` by checking the ID,  URL and title against re3data, then checks the URL and title against `internal storage`, and if there is still no match, it gets mapped to an `external storage` object.

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
BlueTechie: This is the case whenever you create a DMP with exactly one dataset and exactly one internal/external storage or repository in the UI, but you don't explicitly link them. I believe this scenario can realistically happen. While distribution is not a mandatory subfield of Dataset in the RDA Common Standard, a lot of relevant information such as licensing and data access would get lost without the fallback option.

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
```diff
@@ -0,0 +1,662 @@
...
+        log.errorv("Unexpected exception retrieving repository name for ID {0}", r3dataId, e);
+      }
+    }
+
+    if (repoUrl == null || repoUrl.isBlank()) {
+      log.warnv(
+          "Repository '{0}' (ID: {1}) has no valid URL and will be skipped in the export.",
+          repoTitle, r3dataId);
+      return null;
+    }
```
ValentinFutterer, line 252: `repoUrl` is always at least the google URL. I would basically write the warning if there is no other option and say soemthing like "no repo url available, google is used as backup"

ValentinFutterer: And I would remove the return null, or does it serve a purpose here?

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
```diff
@@ -0,0 +1,662 @@
...
+          log.warnv("Repository registry ID {0} not found in re3data", r3dataId);
+        } else if (errorCode == EErrorCode.RE3DATA_NOT_AVAILABLE) {
+          log.errorv(
+              "re3data service not available when fetching metadata for ID {0}", r3dataId, e);
+        } else {
+          log.errorv("Unexpected API error retrieving repository metadata for ID {0}", r3dataId, e);
+        }
+      } catch (Exception e) {
+        log.errorv("Unexpected exception retrieving repository name for ID {0}", r3dataId, e);
+      }
```
ValentinFutterer, line 244: Should this return null on error?

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
```diff
@@ -0,0 +1,662 @@
...
+   * @param target the target DMP to import the host into
+   * @param rdaHost the source RDA standard host
+   * @param refHash the reference hash of the dataset linked to this host
+   */
+  public void importHost(DmpDO target, Host rdaHost, String refHash) {
+    if (rdaHost == null) {
+      return;
+    }
+    String title =
+        rdaHost.getTitle() != null && !rdaHost.getTitle().isBlank()
```
ValentinFutterer: Title cannot be null, sinc eits marked as nonnull

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
```diff
@@ -0,0 +1,662 @@
...
+            ? rdaHost.getTitle()
+            : "Imported Host";
+    String url = rdaHost.getUrl() != null ? rdaHost.getUrl() : "";
```
ValentinFutterer: Same for url

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
```diff
@@ -0,0 +1,662 @@
...
+        }
+      } catch (Exception e) {
+        log.errorv("Unexpected exception checking re3data by explicit hostId {0}", hostIdVal, e);
+      }
+    }
+
+    // B. Check match by URL patterns if pointing to re3data registry
+    String url = rdaHost.getUrl();
+    if (url != null && !url.isBlank()) {
+      if (url.contains("doi.org/10.17616/") || url.contains("re3data.org/repository/")) {
```
ValentinFutterer: why contains doi when the mthod extracts re3data urls?

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
```diff
@@ -0,0 +1,662 @@
...
+            } else {
+              log.errorv("Unexpected API error checking URL-extracted ID {0}", extractedId, e);
+            }
+          } catch (Exception e) {
+            log.errorv(
+                "Unexpected exception checking re3data by URL-extracted ID {0}", extractedId, e);
+          }
+        }
+      }
+    }
```
ValentinFutterer: Will this ever happen? Usually for URL you give the URl of the actual system like homepage right? Is it not possible to search the re3data API by URL?

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
```diff
@@ -0,0 +1,662 @@
...
+   * @return the matching InternalStorage entity, or null if no match is found
+   */
+  private InternalStorage findInternalStorageMatch(String title, String url) {
+    try {
+      List<InternalStorage> allStorages = InternalStorage.listAll();
+      for (InternalStorage storage : allStorages) {
+        if (title != null) {
+          if (title.equalsIgnoreCase(storage.getStorageLocation())
+              || title.equalsIgnoreCase(storage.getBackupLocation())) {
+            return storage;
```
ValentinFutterer: The title of a storage is held by its storage translation, comparing title against location will very likely never produce a match.

src/main/java/org/damap/base/rda/dmpcommonstandard/HostsMapper.java
```diff
@@ -0,0 +1,662 @@
...
+      storeDO.setTitle(
+          internal.getStorageLocation() != null
+              ? internal.getStorageLocation()
+              : (internal.getBackupLocation() != null
+                  ? internal.getBackupLocation()
+                  : "Internal Storage"));
+      storeDO.setDatasets(new ArrayList<>(List.of(refHash)));
+      target.getStorage().add(storeDO);
+    }
+  }
```
ValentinFutterer: Doesnt this miss the `InternalStorageTranslation`?

ValentinFutterer (commented):
I know its not part of this PR, but contributors get duplicated when reimporting a DMP after exporting, I think the fault is with the export:

Image

Image

Datasets are only imported as new data, even though they were reused before - that really bad. The CS has an `is_reused` field, please use it.

Estimated type of new data is not re-imported, is that a limitation of the CS or a fault with the code?

Storage option name changes when exporting and reimporting.

At least the report is not exported / imported. Exists i think is ex/imported. Description I am not sure if we have this in DAMAP, please check.

Image

Costs are completely missing from the export / import.

You can assign two datasets to a single storgage/repo. But when exporting and then reimporting, there is only one datset assigned to each host.

ValentinFutterer:
As an addendum to the above comment, the estimated publication date of a license is also not exported / imported. Its equivalent to the `start_date` in the CS license. Its called `startDate` in the `DatasetDO`.

image

src/main/java/org/damap/base/rda/dmpcommonstandard/DatasetMapper.java
ValentinFutterer: This should be removed, as its a mistake with the CS.

ValentinFutterer (approved):
Its still lacking and has visible bugs, see https://github.com/damap-org/damap-backend/issues/532. The code is fineish, but it will be almost impossible to maintain. We need to automize the translations between the data formats as much as possible. We also need to define an DAMAP application profile, to make exporting / importing less lossy.

PR #522, opened 2026-07-01
Byte Size: Specify long (int64) instead of default int32

BlueTechie:
## Description

Bugfix
#### What does this PR do?

Because the `openapi.yaml` spec does not specify a format for the `byte_size` field of the `Distribution` object, `int32` is taken as default. However, this is inadequate for byte sizes, since exporting the RDA JSON DMP with a dataset of more than roughly 2.2 GB would already result in an Integer overflow. Currently, an internal server error (500) is generated without proper error handling. This will be addressed in another slightly larger PR shortly, along with some other issues with the mappers.


ValentinFutterer (approved):
This should fix the issue 👍

PR #527, opened 2026-08-03
Pure: Fix wrong config key causing http api to not be loaded

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?
Added a missing `tenant-aware` to the key in `@LookupIfProperty` of `HTTPBasedPureAPI`. Without this key, the bean is not loaded.


lpandath:
@ValentinFutterer  Tested against real Pure API today with a project that has staff + PI.

- Pure project search works
- Pure person search works
- Project staff fetch works
- Contributors get hydrated correctly (see Eduard Paschke in recommended, Armin Gerger as PI in contributor list)
- No warnings or errors related to the HTTPBasedPureAPI bean in the boot log
- No regressions

Screenshot 2026-08-07 at 10 54 45


Fix looks good. LGTM.

